PR #4, opened 2019-11-19
Set up api-extractor

vidartf:
This does not yet build any documentation off the extracted APIs, but is used for tracking any changes to the API.

A possible future TODO: Add `@public` / `@internal` etc tags to docstrings (https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags) and make it a warning/error to not have such a tag.


vidartf:
Note: I'm assuming the check-in docs previously fulfilled this role, ref https://github.com/jupyterlab/lumino/commit/cf7b68fc3c1da6761ef4e57d8dfa31bca463bdfb

blink1073:
LGTM, thanks!

PR #14, opened 2019-12-06
DataGrid Cell Editing

mbektasbbg:
Adds editing capabilities to DataGrid. A new DataModel type MutableDataModel added and ICellEditorController interface introduced with a concrete implementation and ICellEditor interface with several cell editor implementations and ICellInputValidator interface with concrete implementations for some types. I also added cursor movement using Tab & Enter keys within body or confined to selections when there are any.

Editing is possible if data model is of type MutableDataModel and there is a selection model. It is triggered using double click or typing. Editing can also be triggered programmatically, optionally with a custom editor and validator. Editor type to use is inferred from metadata if the metadata follows the schema defined in https://frictionlessdata.io/specs/table-schema/, otherwise if cell data is a primitive type TextCellEditor is used.

Cell editors to used can be set/overridden based on cell data type or metadata key-value pair match.

DataGrid example is modified to present most of the editing features including a custom cell editor.

CellEditing

### Cursor movement

- Tab -> right, Shift + Tab -> left
- Enter -> down, Shift + Enter -> up
- Editing starts at cursor position and cursor moves to next column / next row when next move step requires so
- Cursor moves within selection rectangles and jumps to next rectangle when next move step requires so

CursorMovement

Deleting Cell values (setting to null) using Delete key

MutableDataModel which defines new method setData

Cell Editor
- If cells have metadata information (based on schema at https://frictionlessdata.io/specs/table-schema/), type information is used to choose best cell editor for a cell

TextCellEditor
- Default editor unless there is a better match and cell data type is primitive
Editor for cells of type string, displayed as a text input
- Field Metadata {type: 'string'}

NumberCellEditor
- Editor for real number typed cells, displayed as a text input
- Field Metadata {type: 'number'}

IntegerCellEditor
- Editor for integer typed cells, displayed as a spinner input
- Field Metadata {type: 'integer'}

BooleanCellEditor
- Editor for boolean typed cells, displayed as a checkbox
- Field Metadata {type: 'boolean'}

DateCellEditor
- Editor for date typed cells, displayed as a date picker
- Field Metadata {type: 'date'}

OptionCellEditor
- Editor for cell with data types that have enum options, displayed as a drop down list
- Field Metadata {type: any, constraint: {enum: [list of options]}}

DynamicOptionCellEditor
- Editor for cell with data types that have options from previously entered values in the same column, displayed as an autocomplete drop down list
- Field Metadata {type: any, constraint: {enum: 'dynamic'}}. This is not a standard schema format, I extended it.

Editor Controller
- Manages editor overrides and instances
- Editors can be overridden based on cell data type and metadata key-value pair match
- Controller makes sure there is only one editor active at a time

Validators
- As user enters a value, input is validated and a warning message is shown to user. Committing is only possible if there are no validation errors. Validation error messages are shown using standard HTML form validation system.

TextInputValidator
- Checks for valid text entry
- Can check for min/max text length and regex pattern match requirements

NumberInputValidator
- Checks for valid real number entry
- Can check for min/max bounds

IntegerInputValidator
- Checks for valid integer entry
- Can check for min/max bounds


packages/datagrid/src/basickeyhandler.ts
```diff
@@ -57,6 +66,40 @@ class BasicKeyHandler implements DataGrid.IKeyHandler {
...
+      if (/[a-zA-Z0-9-_ ]/.test(input)) {
+        const row = grid.selectionModel!.cursorRow;
+        const column = grid.selectionModel!.cursorColumn;
+        const cell: CellEditor.CellConfig = {
+          grid: grid,
+          row: row,
+          column: column
+        };
+        grid.editorController!.edit(cell, {
+          onCommit: (response: ICellEditResponse) => {
```
blink1073: This logic is the same as the one used in the mousehandler (but slightly different).  Perhaps the commit/cancel behavior should have a default implementation in `editorController` and allow it to be overridden?

packages/datagrid/src/datagrid.ts
```diff
@@ -129,11 +134,13 @@ class DataGrid extends Widget {
...
     this._scrollCorner = new Widget();
 
+    this._editorController = new CellEditorController();
+
     // Add the extra class names to the child widgets.
-    this._viewport.addClass('p-DataGrid-viewport');
-    this._vScrollBar.addClass('p-DataGrid-scrollBar');
-    this._hScrollBar.addClass('p-DataGrid-scrollBar');
-    this._scrollCorner.addClass('p-DataGrid-scrollCorner');
+    this._viewport.addClass('lm-viewport');
```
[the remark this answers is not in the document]
mbektasbbg: yes @blink1073, I will update

packages/datagrid/src/basickeyhandler.ts
mbektasbbg: @blink1073 agreed. I will update.

mbektasbbg:
@blink1073 @jasongrout @afshin, except from code comments I think this is ready to go. I fixed issues on non-chrome browsers and it is stable on Chrome, Firefox and Safari now. I will start adding comments.

blink1073:
I looked through the latest commits and gave it a test spin in Firefox.  It is looking good!

SylvainCorlay:
> I looked through the latest commits and gave it a test spin in Firefox. It is looking good!

Yeah! Do you think this could go into master soon?

blink1073:
@mbektasbbg are you able to make a pass at adding the comments?

mbektasbbg:
> @mbektasbbg are you able to make a pass at adding the comments?

sure @blink1073, I will do that at earliest chance.

mbektasbbg:
@blink1073 I am done with comments. could you take another look?

blink1073:
Released as `@lumino/datagrid@0.5.0`

telamonian:
Bravo @mbektasbbg! This is pretty cool, and another big step forward for `datagrid`, alongside the recent copy/paste/select work

Issue #9, opened 2019-12-04
CSS selector discussion


afshin:
Should all of the legacy `p-*` CSS selectors be renamed to `lm-*`?
- In the interest of clarity (and purity), my intuition is "yes"
- In the interest of lower churn and a smaller amount of of pain imposed on current consumers of these APIs, "no" seems better.

jasongrout:
I think this should be a change for Lumino 2.0.

afshin:
So you think it should be a "yes" ... eventually? Do you think there's room for just leaving them as is or do you consider it too much of an inconsistency in nomenclature to keep the `p-*` classes?

jasongrout:
I think the community around Lumino is just starting, and it's more valuable in the long term to have `lm-` prefixes.

So my intuition is "yes, at the earliest opportunity, which is 2.0".

jasongrout:
In fact, I think if that is our decision, we should add `lm-` classes right now, and deprecate `p-` classes even before 2.0.

afshin:
> In fact, I think if that is our decision, we should add lm- classes right now, and deprecate p- classes even before 2.0.

This is a good idea.

vidartf:
> deprecate p- classes even before 2.0.

I'm not sure what you mean here. The way I read semver, you cannot remove anything (whether deprecated or not) in a minor release.

afshin:
We won't remove it. We'll add a note that the selector is deprecated, but it will continue to work until 2.0 and the `lm-*` will work as well.

afshin:
Hm.

I am not sure we can support the legacy deprecation because when we are checking for `p-mod-foo` in the TS code, do we now need to check for either `p-mod-foo` and also `lm-mod-foo`?

afshin:
Okay, so the crux of the question is encapsulated in this example CSS:

```css
/* <DEPRECATED> */
.p-TabBar-tab.p-mod-hidden,
/* </DEPRECATED> */
.lm-TabBar-tab.lm-mod-hidden {
  display: none !important;
}
```

The `.p-mod-hidden` inside the deprecated class should never naturally arise once we switch to using `lm-mod-hidden`, however a consumer may have written custom CSS that targets something with a `p-mod-hidden` class on it. This is a problem that cannot be solved in pure CSS. So the question that arises is: **should we add `p-mod-hidden` in addition to `lm-mod-hidden` in the TS logic?** This doesn't feel correct to me.

jasongrout:
[...]

PR #20, opened 2019-12-07
Update selector, data attribute, and event namespaces.

afshin:
Fixes https://github.com/jupyterlab/lumino/issues/9

All of the legacy `p-*` CSS selectors should be renamed to `lm-*` and the `p-*` classes should continue to work but be deprecated until the next major release.


afshin:
This is ready for review.

I think it needs careful consideration.

PR #23, opened 2019-12-10
Enable / disable runtime tab dragging in DockPanel

GordonSmith:
Allow user to enable / disable tab dragging at runtime.

### DIFF
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -68,13 +68,16 @@ class DockPanel extends Widget {
     this._mode = options.mode || 'multiple-document';
     this._renderer = options.renderer || DockPanel.defaultRenderer;
     this._edges = options.edges || Private.DEFAULT_EDGES;
+    if (options.tabsMovable !== undefined) {
+      this._tabsMovable = options.tabsMovable;
+    }
 
     // Toggle the CSS mode attribute.
     this.dataset['mode'] = this._mode;
 
     // Create the delegate renderer for the layout.
     let renderer: DockPanel.IRenderer = {
-      createTabBar: () => this._createTabBar(),
+      createTabBar: () => this._createTabBar(this._tabsMovable),
       createHandle: () => this._createHandle()
     };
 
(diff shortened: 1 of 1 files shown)


packages/widgets/src/dockpanel.ts
```diff
@@ -192,6 +195,21 @@ class DockPanel extends Widget {
...
 
+  /**
+   * Whether the tabs can be dragged / moved at runtime.
+   */
+  get tabsMovable(): boolean {
+    return this._tabsMovable;
+  }
+
+  /**
+   * Enable / Disable draggable / moveable tabs.
```
afshin: sp. `moveable` => `movable`

packages/widgets/src/dockpanel.ts
```diff
@@ -1010,6 +1029,14 @@ namespace DockPanel {
      * If not given, default values will be used.
      */
     edges?: IEdges;
+
+    /**
+     * Allow tabs to be movable by user.
+     * When disabled the user will be unable to customize the layout.
```
afshin: I would remove the `When disabled` line ... because presumably resizing the `TabPanel` instances still counts as customizing the layout.

packages/widgets/src/dockpanel.ts
```diff
@@ -1010,6 +1029,14 @@ namespace DockPanel {
...
+     *
+     * The deafult is `'true'`.
```
afshin: sp. `deafult` => `default`

GordonSmith:
@afshin changes made as requested.

packages/widgets/src/dockpanel.ts
```diff
@@ -68,13 +68,16 @@ class DockPanel extends Widget {
     this._mode = options.mode || 'multiple-document';
     this._renderer = options.renderer || DockPanel.defaultRenderer;
     this._edges = options.edges || Private.DEFAULT_EDGES;
+    if (options.tabsMovable !== undefined) {
+      this._tabsMovable = options.tabsMovable;
```
afshin: I think I would just do:
```typescript
this._tabsMovable = !!options.tabsMovable; 
```

GordonSmith: Wouldn't that mean that an `undefined` would override the default `true`?

afshin: Correct. I am dumb. I am used to thinking of defaults as `false`.

GordonSmith: In TSC 3.7 it would be:  `this._tabsMovable = ??options.tabsMovable;` (from memory)

PR #24, opened 2019-12-10
chore(build): Add missing package.json dependencies

GordonSmith:


afshin:
Hi @GordonSmith! Thanks for the PR.

Two things:
1. Could you please add newlines at the end of each file? We use the Unix standard EOF, so each file ends with a newline.
1. I'm not sure we want to switch to NPM instead of Yarn, but missing dependencies are missing dependencies, so thanks for fixing this!

GordonSmith:
@afshin newlines added (auto formatting editor removed them)

PR #25, opened 2019-12-10
fix(messaging): Remove non standard functions (setImmediate, clearImm…

GordonSmith:
…ediate)

Remove non standard set/clear Immediate and replace them with an animationFrame polyfill

### DIFF
--- a/packages/messaging/src/index.ts
+++ b/packages/messaging/src/index.ts
@@ -494,7 +494,14 @@ namespace MessageLoop {
    */
   const schedule = (() => {
     let ok = typeof requestAnimationFrame === 'function';
-    return ok ? requestAnimationFrame : setImmediate;
+    let prev = 0;
+    return ok ? requestAnimationFrame : (callback: FrameRequestCallback) => {
+      const curr = new Date().getTime();
+      const next = Math.max(0, 16 - (curr - prev));
+      const handle = window.setTimeout(() => callback(curr + next), next);
+      prev = curr + next;
+      return handle;
+    };
   })();
 
   /**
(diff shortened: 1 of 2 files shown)


GordonSmith:
Depending on what browsers you support, there may be no need for the polyfill at all (IOW you could assume that requiestAnimationFrame always exists)?

afshin:
The `@lumino/messaging` library is meant to support `node` as well. It is not browser-only, so the logic ought to remain as is.

GordonSmith:
@afshin just to clarify, your closing because:
* All browsers have adequate `requestAnimationFrame` support
* If `requestAnimationFrame` is absent, your assuming its a NodeJS environment and setImmediate is known to exist
?

afshin:
Yes, the browsers Lumino supports have `requestAnimationFrame` and the only other environment we currently expect this library to run in is `node`, which has `setImmediate` defined.

GordonSmith:
Ok - works for me (FYI you will start to get TS syntax errors in the future - but a ts-ignore and suitable comment should be fine...)

afshin:
Can we add `node` to the `lib` array in `tsconfig.json` to resolve that when it arises?

GordonSmith:
Probably - but not with the current TS compiler version. (I will test later today - slowly working through my forks diffs).
FYI you won't be able to upgrade the TS compiler to 3.7.x due to:  https://github.com/microsoft/TypeScript/issues/33939

afshin:
Cool, thank you!

GordonSmith:
Looks like there isn't a specific ts lib option for this.  We can continue this discussion when I issue a PR for TSC 3.6.4

afshin:
Cheers!

PR #27, opened 2019-12-10
chore(build): Bump typescript to version 3.6.4

GordonSmith:

### DIFF
--- a/packages/datagrid/src/graphicscontext.ts
+++ b/packages/datagrid/src/graphicscontext.ts
@@ -94,33 +94,33 @@ class GraphicsContext implements IDisposable {
     }
   }
 
-  get textAlign(): string {
+  get textAlign(): CanvasTextAlign {
     return this._context.textAlign;
   }
 
-  set textAlign(value: string) {
+  set textAlign(value: CanvasTextAlign) {
     if (this._state.textAlign !== value) {
       this._state.textAlign = value;
       this._context.textAlign = value;
     }
   }
 
-  get textBaseline(): string {
+  get textBaseline(): CanvasTextBaseline {
     return this._context.textBaseline;
   }
 
-  set textBaseline(value: string) {
+  set textBaseline(value: CanvasTextBaseline) {
     if (this._state.textBaseline !== value) {
       this._state.textBaseline = value;
       this._context.textBaseline = value;
     }
   }
 
-  get lineCap(): string {
+  get lineCap(): CanvasLineCap {
     return this._context.lineCap;
   }
 
-  set lineCap(value: string) {
+  set lineCap(value: CanvasLineCap) {
     if (this._state.lineCap !== value) {
       this._state.lineCap = value;
       this._context.lineCap = value;
@@ -138,11 +138,11 @@ class GraphicsContext implements IDisposable {
     }
   }
 
-  get lineJoin(): string {
+  get lineJoin(): CanvasLineJoin {
     return this._context.lineJoin;
   }
 
-  set lineJoin(value: string) {
+  set lineJoin(value: CanvasLineJoin) {
     if (this._state.lineJoin !== value) {
       this._state.lineJoin = value;
       this._context.lineJoin = value;
@@ -415,7 +415,7 @@ class GraphicsContext implements IDisposable {
     return this._context.createRadialGradient(x0, y0, r0, x1, y1, r1);
   }
 
-  createPattern(image: HTMLImageElement | HTMLCanvasElement | HTMLVideoElement, repetition: string): CanvasPattern {
+  createPattern(image: HTMLImageElement | HTMLCanvasElement | HTMLVideoElement, repetition: string): CanvasPattern | null {
     return this._context.createPattern(image, repetition);
   }
 
--- a/packages/messaging/src/index.ts
+++ b/packages/messaging/src/index.ts
@@ -479,10 +479,12 @@ namespace MessageLoop {
     console.error(err);
   };
 
+  type ScheduleHandle = number | any; //  requestAnimationFrame (number) and setImmediate (any)
+
   /**
    * The id of the pending loop task animation frame.
    */
-  let loopTaskID = 0;
+  let loopTaskID: ScheduleHandle = 0;
 
   /**
    * A guard flag to prevent flush recursion.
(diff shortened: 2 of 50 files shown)


GordonSmith:
Note:  You should avoid upgrading to TS compiler 3.7.x due to: microsoft/TypeScript#33939

afshin:
cc: @jasongrout, who is usually interested in the compiler version

vidartf:
Will this change cause builds by consumers to fail if they use an older version of TS?

GordonSmith:
@vidartf I don't believe so - unfortunately the same is not true of 3.7.x

packages/messaging/src/index.ts
```diff
@@ -494,6 +494,7 @@ namespace MessageLoop {
    */
   const schedule = (() => {
     let ok = typeof requestAnimationFrame === 'function';
+    // @ts-ignore setImmediate is valid for NodeJS
```
vidartf: Could this be resolved by adding a dev dependency on `@types/node`?

packages/datagrid/src/graphicscontext.ts
```diff
@@ -415,7 +415,7 @@ class GraphicsContext implements IDisposable {
     return this._context.createRadialGradient(x0, y0, r0, x1, y1, r1);
   }
 
-  createPattern(image: HTMLImageElement | HTMLCanvasElement | HTMLVideoElement, repetition: string): CanvasPattern {
+  createPattern(image: HTMLImageElement | HTMLCanvasElement | HTMLVideoElement, repetition: string): CanvasPattern | null {
```
vidartf: This is the only line I can think of that might break something downstream, but this should really be considered a "bug fix" (fixing an incorrect typing).

GordonSmith: I Agree.

GordonSmith:
I just did a diff of the generated files - the only potential issue I see is the use of the "correct" types in  `graphicscontext.d.ts` (do older TS versions have access to the latest "lib.dom.d.ts")?

FWIW I have not used these types in my other projects (which have had no issues with older TSC versions).

GordonSmith:
I believe `polling` was already included?

vidartf:
> I believe `polling` was already included?

I don't see the package.json in the diff? And a `yarn why typescript` off the branch still reported polling as being on `~3.0.3`..

vidartf:
(I can push these changes if you are not at a computer)

packages/messaging/src/index.ts
GordonSmith: I don't think so as its not a NodeJS API "thing" (its a global "window"/"this" thing).

GordonSmith:
I just rebased and see that now (I suspect I just visually missed it before as polling wasn't in my original branch).

I see that "notebooks" is on 3.7.2 should this also be 3.6.4?

GordonSmith:
What is notebooks and should it be in lumino?

packages/messaging/src/index.ts
vidartf: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/028e63335139a3a0639b538a6aab9fe3fb7cfa38/types/node/timers.d.ts#L10-L15 defines it at least.

GordonSmith: I believe that is a different one to the "window.setImmediate" that lumino is using (the v8 JS runtime one)

vidartf:
> What is notebooks and should it be in lumino?

Take it up with @ellisonbg https://github.com/jupyterlab/lumino/pull/8 (it is just some maintainer code to discuss code in the datastore package, but I would argue it should be a gist instead).

packages/messaging/src/index.ts
GordonSmith: (IOW timers.setImmediate vs GLOBAL.setImmediate)

GordonSmith:
If it (notebooks) doesn't make it into the npm tar file, then its really up to you folks (I can't say I even noticed it was there).

packages/messaging/src/index.ts
vidartf: https://nodejs.org/api/timers.html#timers_timers

> The timer module exposes a global API for scheduling functions to be called at some future period of time. Because the timer functions are globals, there is no need to call require('timers') to use the API.

vidartf:
> If it (notebooks) doesn't make it into the npm tar file

it doesn't.

packages/messaging/src/index.ts
GordonSmith: And that shows my lack of NodeJS experience!  Let me test - default to @types/node @ v12 (LTS)?

vidartf: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5f59f3dc7a0965fb1767cf13c670c633bbbbf0a8/types/node/globals.d.ts#L177-L181

vidartf: As the typing of `schedule`/`unschedule` is not exported, it should suffice to ensure that it compiles, so v12 seems reasonable.

GordonSmith: no luck (after testing) and visually inspecting the NodeJS typings its declared in the "timers" module and not globally.

There is a @types/setimmediate which should work - but do you want to go down that road?

At some point you might want a "platform" test:
```
declare const process: any;

//  @ts-ignore
export const root: any = new Function("return this;")(); //  Prevent bundlers from messing with "this"

export const isBrowser: boolean = typeof window !== "undefined" && root === window;
export const isNode: boolean = typeof process !== "undefined" && process.versions != null && process.versions.node != null;
export const isTravis: boolean = isNode && process.env != null && process.env.TRAVIS != null;
```
See:  https://github.com/hpcc-systems/Visualization/blob/master/packages/util/src/platform.ts 
  
And will be able to add TypeScript guards for things like setImmediate.

Its still pretty tricky getting a single TypeScript package to "just work" for Node and Browser especially when you get into fetch polyfills and the like...

vidartf: @GordonSmith Did you see my link to the `global.d.ts` file? Also, did you add the types dependency to `tsconfig.json`'s `types` field?

GordonSmith:
Pushed changes for polling which will need a new review.

packages/messaging/src/index.ts
GordonSmith: I see that now - pushed new version with suggested change.

GordonSmith:
Pushed with @types/node included, the following will need re-review:
* messaging
* polling

PR #29, opened 2019-12-14
Adds a "pass thru" virtual element

telamonian:
Fixes phosphorjs/phosphor#395, phosphorjs/phosphor#436, and probably a bunch of other issues.

This PR adds the hpass function and VirtualElementPass class described in phosphorjs/phosphor#436.

This PR was originally phosphorjs/phoshor#437. @blink1073 Since you reviewed the original PR, could you please take a look at this one too?

### DIFF
--- a/packages/widgets/src/tabbar.ts
+++ b/packages/widgets/src/tabbar.ts
@@ -32,7 +32,7 @@ import {
 } from '@lumino/signaling';
 
 import {
-  ElementDataset, ElementInlineStyle, VirtualDOM, VirtualElement, h
+  ElementDataset, ElementInlineStyle, VirtualDOM, VirtualElement, h, hpass, VirtualNode
 } from '@lumino/virtualdom';
 
 import {
@@ -1339,9 +1339,15 @@ namespace TabBar {
      *
      * @returns A virtual element representing the tab icon.
      */
-    renderIcon(data: IRenderData<any>): VirtualElement {
+    renderIcon(data: IRenderData<any>): VirtualNode {
+      const { title } = data;
       let className = this.createIconClass(data);
-      return h.div({ className }, data.title.iconLabel);
+
+      if (title.iconPass) {
+        return hpass(title.iconPass);
+      } else {
+        return h.div({className}, data.title.iconLabel);
+      }
     }
 
     /**
--- a/packages/widgets/src/tabbar.ts
+++ b/packages/widgets/src/tabbar.ts
@@ -1343,8 +1343,8 @@ namespace TabBar {
       const { title } = data;
       let className = this.createIconClass(data);
 
-      if (title.iconPass) {
-        return hpass(title.iconPass);
+      if (title.iconRender) {
+        return hpass(title.iconRender, 'div');
       } else {
         return h.div({className}, data.title.iconLabel);
       }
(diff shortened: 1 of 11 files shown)


telamonian:
Here's the original motivation from https://github.com/phosphorjs/phosphor/issues/436#issue-504405971:

> Following up on a suggestion made by @jasongrout during the latest Phosphor session/call, I think we should add a "pass thru" vdom element that renders a real DOM element using a callback. This could act like an escape hatch from Phosphor's vdom and allow for composition with vdom elements from other libraries (like React, hint, hint).
>
> Given the structure of the virtualdom package and the updateContent function, I think it would make sense to implement this via an hpass function with the following signature:
>
>```typescript
>export function hpass(callback: (host: HTMLElement) => void): VirtualElementPass {
>  ...
>}
>```
>
> This could exist alongside the extant h function. During updateContent, when a node with:
>
> ```typescript
>node.type === 'passthru'
>```
>
> is encountered, node.callback(host) will be invoked, and from that point on any diffing/updating will be the responsibility of the callback.
>
> I think the most important/urgent use case for this is allowing for the blending of React vdom into Phosphor vdom. The above pattern can mesh very nicely with ReactDOM.render(host, Component), since that function does its own diffing. This would allow Phosphor to better accommodate React while still avoiding any dependency on it.

telamonian:
And here's a summary of the contents of the PR from https://github.com/phosphorjs/phosphor/pull/437#issuecomment-541738804:

> This PR is now functional, and at a point where I could really use some feedback. Here's what `hpass` currently looks like:

> https://github.com/jupyterlab/lumino/blob/71f427ecb9743357d79524127fe43edd08317df0/packages/virtualdom/src/index.ts#L986-L1000

> `hpass` will create a vdom node just like `h` does, but it doesn't allow children. Instead, the node's children will be created/managed by the `render` callback. This ensures that nodes added by `render` won't have any immediate siblings or children. It's set up this way since React (and presumably other vdom libs) doesn't play well with DOM nodes that it didn't create.

blink1073:
Released `@lumino/virtualdom@1.3.0`

telamonian:
> Released `@lumino/virtualdom@1.3.0`

Awesome. That's going to save me a lot of fiddling with `yarn link`. Thanks Steve!

packages/widgets/src/tabbar.ts
```diff
@@ -1339,9 +1339,15 @@ namespace TabBar {
...
-    renderIcon(data: IRenderData<any>): VirtualElement {
+    renderIcon(data: IRenderData<any>): VirtualElement | VirtualElementPass {
+      const { title } = data;
       let className = this.createIconClass(data);
-      return h.div({ className }, data.title.iconLabel);
+
+      if (title.iconRenderer) {
+        return hpass('div', title.iconRenderer);
+      } else {
+        return h.div({className}, data.title.iconLabel);
```
vidartf: If the `className` is completely disregarded when `title.iconRenderer` is set, then maybe `this.createIconClass(data)` should not be called at all?

telamonian: You're right, this is inconsistent, and kind of a holdover from an earlier point in the PR. I thought about it for a bit, and what I think should be done is that the `className` and `label` should be passed to the hpass element as well, to be set on its parent element. There's a PR for it at #36

PR #30, opened 2019-12-17
Use the standby value generated instead of ignoring it.

jasongrout:
The standby value generated in the line above is ignored in the current code if it is a boolean. This makes it actually be used where intended.


jasongrout:
CC @afshin

PR #32, opened 2019-12-19
Allow commands to accept partial json objects

blink1073:


blink1073:
I don't see any downside.

blink1073:
Published `commands@1.9.0`

vidartf:
This should probably be update as well then:

https://github.com/jupyterlab/lumino/blob/4030e959710d64892a6da59bb2fa7df8f3fa48d9/packages/widgets/src/menu.ts#L990

vidartf:
Note: I think this might break compilation of command implementor code.

vidartf:
e.g.
```ts
// ...
execute: args => {
  return args['foo'].toString();
}
// ...

PR #33, opened 2019-12-20
Update menu and commandpalette to accept partial json

blink1073:

PR #34, opened 2019-12-20
Allow partial JSON for command callers but not receivers

blink1073:
Alternative to #33, addresses https://github.com/jupyterlab/lumino/pull/32#issuecomment-568076133.

Allows invokers of commands to pass in `PartialReadonlyJSONObject` values, but ensures that the command consumers receive `ReadonlyJSONObject` values (through API and by stripping `undefined` values).

### DIFF
--- a/packages/coreutils/src/json.ts
+++ b/packages/coreutils/src/json.ts
@@ -64,7 +64,7 @@ type ReadonlyJSONValue = JSONPrimitive | ReadonlyJSONObject | ReadonlyJSONArray;
  * Note: Partial here means that JSON object attributes can be `undefined`.
  */
 export
-type PartialJSONValue = JSONPrimitive | PartialJSONObject | PartialJSONArray;
+type PartialJSONValue = ReadonlyJSONValue | PartialJSONObject | PartialJSONArray;
 
 
 /**
--- a/packages/coreutils/src/json.ts
+++ b/packages/coreutils/src/json.ts
@@ -249,6 +249,56 @@ namespace JSONExt {
   }
 
   /**
+   * Strip out any `undefined` values from a JSON object.
+   *
+   * @param value - The JSON value to copy.
+   *
+   * @returns The given JSON value with `undefined` values stripped, recursively.
+   */
+  export
+  function stripUndefined(value: ReadonlyPartialJSONObject): ReadonlyJSONObject {
+    for (let key in value) {
+      // Remove undefined values.
+      let subvalue = value[key];
+      if (subvalue === undefined) {
+        delete (value as any)[key];
+        continue;
+      }
+
+      // Do nothing for primitive values.
+      if (isPrimitive(subvalue)) {
+        continue;
+      }
+
+      // Strip an array.
+      if (isArray(subvalue)) {
+        stripUndefinedArray(subvalue);
+        continue;
+      }
+
+      // Strip an object.
+      stripUndefined(subvalue);
+    }
+    return value as ReadonlyJSONObject;
+  }
+
+
+  /**
+   * Strip a ReadonlyPartialArray of any underlying `undefined` keys.
+   */
+  function stripUndefinedArray(value: ReadonlyPartialJSONArray) {
+    for (let i = 0, n = value.length; i < n; ++i) {
+      const subValue = value[i];
+      if (isPrimitive(subValue)) {
+      } else if (isArray(subValue)) {
+        stripUndefinedArray(subValue);
+      } else {
+        stripUndefined(subValue);
+      }
+    }
+  }
+
+  /**
    * Compare two JSON arrays for deep equality.
    */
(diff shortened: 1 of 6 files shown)


vidartf:
While this compiles, does this make sense? E.g. won't this allow errors to creep into the JS that were previously caught by the compiler?

```ts
execute(args) {
  for (let key in args) {
    if (args[key] !== null) {
      console.log(args[key].toString());
    }
  }
}

commands.execute('myid', { foo: undefined });
```

blink1073:
We could strip out all args that are `undefined`

vidartf:
> We could strip out all args that are undefined

I think that would defeat the purpose of accepting partial types 😄

blink1073:
I don't think so - we aren't claiming to use any `undefined` params for consumers.

blink1073:
Notes:

- add methods to `JSONExt` to convert from `Partial` to regular JSON types, stripping out anything that is `unknown`
- use those methods whenever we are converting from one type to the other
- add a test to make sure we can pass a `Partial` object to a command, and that we can cast the args of the execute method to a `Partial` as well, and that any `unknown` values are stripped
- add tests for the utility methods

blink1073:
It turned out that `JSONExt.deepyCopy()` already removed `undefined` values, so I used that instead.

blink1073:
After this is merged, I'll cut a patch of `commands` to fix what was a breaking change in the last minor, and a minor release of `widgets`.

vidartf:
Ci failure seems relevant.

blink1073:
I addressed it in the last commit, does that look good to you?

blink1073:
What the heck, it is working for me locally...

vidartf:
It fails for me locally. Maybe a `git clean -fdx`?

blink1073:
I couldn't find a way to allow `PartialReadonlyJSONObject` as the `args` type for the `execute` callback, even with a type cast, since we can't make `PartialReadonlyJSONObject` extend `ReadonlyJSONObject`.   I think this is good to go now.

packages/coreutils/src/json.ts
```diff
@@ -64,7 +64,7 @@ type ReadonlyJSONValue = JSONPrimitive | ReadonlyJSONObject | ReadonlyJSONArray;
  * Note: Partial here means that JSON object attributes can be `undefined`.
  */
 export
-type PartialJSONValue = JSONPrimitive | PartialJSONObject | PartialJSONArray;
+type PartialJSONValue = ReadonlyJSONValue | PartialJSONObject | PartialJSONArray;
```
vidartf: Is this still meant to be included?

blink1073: Reverted

blink1073:
@vidartf are we good to merge this?

vidartf:
I've going back and forth with myself whether I think this is a good idea. Adding a bunch of recursive copy operations just to satisfy backwards compatibility / typings seems like a bad trade to make. On the other hand, I don't want this to be left unresolved much longer.

blink1073:
What if instead of `deepCopy` we use a different method that strips out any values that are `undefined`?  It would still have to be recursive, though.

telamonian:
> Adding a bunch of recursive copy operations just to satisfy backwards compatibility / typings seems like a bad trade to make

That sounds scary. There's no limit on the depth of a json obj, so recursive anything could end up being arbitrarily expensive. Since this is part of Lumino, I think that in general we should favor performance over getting the typings perfect.

As a pragmatic matter, how frequently are calls actually made to eg `execute`? From what I can remember, commands are usually `execute`-ed only in direct response to user input (though some commands do `execute` other commands). Are there any existing situations in which `execute` gets called many times in quick succession?

Also, would the `removeUndefined` routine get removed in the next major release, or would this become a permanent feature of how commands are handled?

blink1073:
It isn't just `execute, menu calls things link `isEnabled` before every render.   In theory you wouldn't be passing around deeply nested JSON as command args, but it is possible.  I think it would be fair for `removeUndefined` to stick around.    

Another alternative is to expose just a `removeUndefined` helper that folks can use on the object they're passing to `execute()`.

vidartf:
Just for clarity and summary: The main point of this work is that previous to #32, command implementors *could* rely on that if a key was present in the args, then the corresponding value was *not* `undefined`. #32 changed that, and this PR is being proposed as a fix to put us back into backwards compatibility.

The other alternative is to revert #32, but add a utility method `removeUndefined`/`stripUndefined` on `JSONExt` (`unPartial`?).

The last alternative is a more pragmatic "what is done is done", and just leave things as-is. Not sure about that one.

vidartf:
> What if instead of deepCopy we use a different method that strips out any values that are undefined? It would still have to be recursive, though.

It is likely to be a lot cheaper since it modifies in-place. I vote for this, given the alternatives. Then we can potentially relax this constraint for the next major version bump.

blink1073:
Updated to add the `stripUndefined` function.  We still have to traverse into arrays, since they may contain objects as value.

Issue #21, opened 2019-12-08
Fix failing FocusTracker tests in Firefox.


afshin:
The `FocusTracker` tests (in `focustracker.spec.ts`) pass in CI and in Chrome, but not in Firefox.

pplonski:
Hi! I was trying to reproduce this with following steps:

```
git clone https://github.com/jupyterlab/lumino.git
cd lumino
yarn
cd packages/widgets
yarn build
cd -
yarn clean
yarn build
yarn build:test
yarn test
```

But it looks like test cant find the `bundle.test.js` file for all packages. The beginning of the test output:

```
yarn run v1.21.1
$ lerna run test
lerna notice cli v3.19.0
lerna info versioning independent
lerna info Executing command in 16 packages: "yarn run test"
lerna info run Ran npm script 'test' in '@lumino/algorithm' in 3.9s:
$ npm run test:firefox

> @lumino/algorithm@1.2.2 test:firefox /home/piotr/sandbox/jl/lumino/packages/algorithm
> cd tests && karma start --browsers=Firefox


START:
19 12 2019 22:01:45.112:WARN [filelist]: Pattern "/home/piotr/sandbox/jl/lumino/packages/algorithm/tests/build/bundle.test.js" does not match any file.
19 12 2019 22:01:45.127:INFO [karma-server]: Karma v4.4.1 server started at http://0.0.0.0:9876/
19 12 2019 22:01:45.127:INFO [launcher]: Launching browsers Firefox with concurrency unlimited
19 12 2019 22:01:45.131:INFO [launcher]: Starting browser Firefox
19 12 2019 22:01:47.825:INFO [Firefox 71.0.0 (Ubuntu 0.0.0)]: Connected on socket 1oRhntqybNuKnk1WAAAA with id 60825187

Finished in 0.001 secs / 0 secs @ 22:01:48 GMT+0100 (Central European Standard Time)

SUMMARY:
✔ 0 tests completed
```

vidartf:
@pplonski Your steps seem to differ from those in the contributing guide: https://github.com/jupyterlab/lumino/blob/master/CONTRIBUTING.md Does it work if you follow that?

pplonski:
[...]

PR #35, opened 2019-12-21
Fix `output.path` for webpack 4

telamonian:
Ref: https://github.com/jupyterlab/lumino/issues/21#issuecomment-567859371

Recently in #31, we upgraded the Webpack used to build the tests in all of the `@lumnio` packages from `v2` => `v4`.  

Webpack `v4` [adds a default value to `output.path`](https://webpack.js.org/configuration/output/#outputpath): `path.join(process.cwd(), 'dist')`. So currently, while our tests sucessfully build, they end up in the `tests/dist/build` dir, and Karma can't find them. This makes all of our test cases empty but successful. This PR fixes that

### DIFF
--- a/packages/algorithm/tests/webpack.config.js
+++ b/packages/algorithm/tests/webpack.config.js
@@ -4,6 +4,7 @@ module.exports = {
   entry: './build/index.spec.js',
   mode: 'development',
   output: {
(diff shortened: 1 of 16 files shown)


vidartf:
Good spot.

PR #36, opened 2019-12-21
Improve handling of attributes for hpass virtualdom elements

telamonian:
cf #29, https://github.com/jupyterlab/lumino/pull/29#discussion_r359318649

This improves the handling of the set of attributes of an hpass virtualdom element. With this PR, the handling of attrs for h and hpass elements now use mostly the same code path.

This is branched off of #35, so that PR should go in before this one


blink1073:
I just published `@lumino/virtualdom@1.4.0` with these changes.

PR #40, opened 2020-01-14
feat(build): Add UMD support

GordonSmith:
Use rollup.js to bundle each package and add UMD support.
Use terser to provide minimized packages.


GordonSmith:
**Notes:**
* The bundled packages include any CSS which was directly imported by the sources
* The "build" step is now actually two steps:  
    * Compile TS files to JS with ES6 compatible modules (used to be commonjs modules)
    * Bundle ES6 modules into UMD + ES6 packages

(Now that I type this I suspect the "watch" scripts will need a tweak)

blink1073:
Hi @GordonSmith, thank you for pushing on this, but I think it over-complicates our build chain.   Most other libraries produce either CommonJS or ES6.  I think we will switch to ES6 when we switch to 2.0 libraries across the board.

GordonSmith:
From a front end (Web Browser) developers perspective I would have to disagree (phosphor didn't "just work" like most of the UI libs I have used).  I would go as far as saying that I am struggling to think of any Browser UI library that only ship as as a commonjs or es6 library!

While its not a big deal for myself as I have already gone through the exercise of creating a UMD shim for phosphor (https://github.com/hpcc-systems/Visualization/tree/master/packages/phosphor-shim), it was certainly bit of a pain at the time and is the type of thing that will put off some potential users...  (IMO)

I could re-work the PR so that the added build steps are only needed during the "publish" pipeline if you like - but if your not interested that's ok as well...

saulshanabrook:
Hey @GordonSmith, thanks for putting in this pr!

We chatted about this at our dev meeting this week and we had some questions about this change. None of us are all that familiar with UMD, but we do want to make it easy for folks to consume lumino! Would you be able/interested to come to our meeting next week to walk us through some of these changes? https://github.com/jupyterlab/team-compass/#weekly-developer-meeting

One question I had is if we could ship the UMD support separately, so we could keep shipping the original `lib` directory for downstream users (maybe keeping it as `commonjs` instead of `es6`?) and then also shipping the UMD stuff you generate with rollup?

GordonSmith:
I am GMT based so if the meeting is at a reasonable hour here I should be able to make it...

While you could create a @jupyterlab/lumino-umd project / npm package, no one really does that anymore (as all the tooling pretty much just works these days) 

In this PR there are basically two packages created
* UMD (a convention which is commonjs / AMD / IIFE friendly)
* ES6 (the "modern" format)

Which are exposed in package.json via:
* main -> dist/index.js (UMD)
* module -> dist/index.es6.js (ES6)

Now I could change that so "main" points at lib/index.js and ensure its the same as it is today. While also including the ES6 / UMD support, your package.json will look like this:
* main -> lib/index.js (CommonJS)
* module -> dist/index.es6.js (ES6)
* browser -> dist/index.js (UMD)

Who will that affect?  
* Straight up NodeJS folks - nothing changes
* Older WebPack users will default to the "main" entry point (no change).
* Newer WebPack users will default to the "module" entry point (its the preferred format for WebPack these days)
* IIFE folks will use the UMD version (will need some manual intervention) 
* AMD folks will use the UMD version (may need some manual intervention)
* Other bundlers (RollupJS etc.) will use the module entry point

Which is still an improvement - but I suspect you fall into the "newer WebPack users" category so you will get hit with the change?  

Finally it is worth noting that this PR didn't touch any of the WebPack / Karma test folders (which was a very conscious decision) as I hoped it would help "prove" that the bundled UMD (dist/index.js) was syntactically the same as the loose files in lib/* (with a slightly opinionated inclusion of the default CSS files...)

saulshanabrook:
Sorry I forgot to reply with the meeting time. It's in half an hour if you can join https://github.com/jupyterlab/team-compass/#weekly-developer-meeting

GordonSmith:
@saulshanabrook I am travelling this week (so only seeing this message now) - next week should work, do you have a Date / GMT time so I can get into a calendar?

afshin:
@GordonSmith Our next call will be on Wednesday, 5 February at 17:00 GMT: https://calpoly.zoom.us/my/jupyter

Thanks!

packages/application/rollup.config.js
```diff
@@ -0,0 +1,35 @@
+import nodeResolve from 'rollup-plugin-node-resolve';
+import sourcemaps from "rollup-plugin-sourcemaps";
+import postcss from "rollup-plugin-postcss";
```
afshin: Purely cosmetic and it probably didn't get picked up because this is `.js` instead of `.ts`, but the Lumino convention is for strings to be single quotes, please.

GordonSmith:
@afshin PR updated with:
* IIFE example
* SourceMap support (debug the IIFE example and step into any of the lumio code and it should resolve all the way back to the TS files now)
* Double quotes replaced with single quotes.

examples/example-dockpanel-amd/index.html
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE html>
```
blink1073: I'm wary of adding the other examples since they aren't TS and they can easily get out of date.

packages/domutils/tsconfig.json
```diff
@@ -1,12 +1,15 @@
 {
   "compilerOptions": {
     "composite": true,
+    "sourceMap": true,
```
telamonian: Won't we also need something like `"inlineSources": true,`? Otherwise, won't we just run into the same issue as in #41 again?

Back in December I experimented extensively with various combinations of the typescript and webpack source map config options. I found that when using both `"sourceMap"` and `"inlineSources"`, you also end up needing to set `"sourceRoot"`. Otherwise, you run the very real risk of ending up with 15 indistinguishable `index.ts` files next to each other at the root of the inferred filesystem of your bundle (ie what you see in the left panel on the "sources" tab of devtools), which makes debugging a real pain. Ideally you want the source files to all be appropriately sorted in their respective `@lumino/whatever` dirs.

However, I was never able to find the "correct" value to use for `"sourceRoot"`. For one thing, when building Lumino packages as part of Jupyterlab's build, the effect of `"sourceRoot"` seems to change depending on whether you do do a dev mode build or just the regular app build. Do you have any insight into the issue, @GordonSmith?

packages/domutils/tsconfig.json
```diff
@@ -1,12 +1,15 @@
...
     "declaration": true,
+    "declarationDir": "./types",
+    "declarationMap": true,
```
telamonian: What's the significance of moving the typings into a new separate dir? What exactly does `"declarationMap"` do here?

packages/domutils/tsconfig.json
GordonSmith: In this case it should be fine as the source maps get bundled as part of the rollup (and minification) processes (including the es6 module which webpack uses).

I will do sanity check on one of the WebPack tests today.

Re the sourceRoot setting - So far I have never had to use this setting in any of my projects...

examples/example-dockpanel-amd/index.html
GordonSmith: Having them non TS is actually a good thing IMO (and why I wen't out of the way to convert them - including IE support).
While they shouldn't break for any minor/point releases they can certainly be added to the suite of tests (to check they load without warning) - I will look into this today?

packages/domutils/tsconfig.json
GordonSmith: This is a slightly opinionated change to avoid generating code into the same folder as the sources.  
It works as the package.json has a "types" property specifically to point to the new location.
`declarationMap` is analogous to `sourceMap` but for definition files.

examples/example-dockpanel-amd/index.html
GordonSmith: Further Note:  Once lumino has UMD support, then these types of examples work well in things like gist ObservableHQ and any online JS editor (the include paths will need to point to jasdelivr or unpkg or some other CDN server).

GordonSmith:
One more observation WRT WebPack + Source Maps - the following configuration should "just work":
```
module.exports = {
    entry: './src/index.js',
    devtool: "source-map",
    output: {
        path: path.join(__dirname, "dist"),
        filename: 'index.min.js',
        library: "quickstart"
    },
    module: {
        rules: [
            { test: /\.js$/, use: ["source-map-loader"], enforce: "pre" }
        ]
    }
}
```
And does indeed work fine on my similar projects.  **But** isn't quite working as expected in the Lerna configuration + examples folder (the source maps don't resolve all the way to the TS files).

I am pretty sure this is because of the way Lerna is hoisting and symlinking the libraries, but once it is released and consumed via "normal" npm mechanism it will just work (and something I will be testing with a RC release).

examples/example-dockpanel-amd/index.html
blink1073: 👍 to including these with tests.

packages/domutils/tsconfig.json
vidartf: `declarationMap` is mostly useful for:
- when you are linking locally (e.g. yarn link). With it, if you try to follow a lumino type to its declaration from lab source code, and lumino is linked locally, you will go to the source file, not the typings file.
- if you distribute the TS source with the package, you will be able to do the above also for non-linked cases.

examples/example-dockpanel-amd/index.html
GordonSmith: @blink1073 you would think that would have been easy - ended up rolling my own with puppeteer

kinow:
Don't have a lot of experience with UMD, but I think the current code deployed to NPM cannot be used directly in unpkg, and I think with UMD that would be possible (I'm trying to write a codepen with a Vue.js example and struggling to find a way to import Lumino modules). If so, +1 to adding UMD support

GordonSmith:
@kinow yup this will solve that issue - FWIW I would recommend jsdelivr over unpkg, but both should work.

kinow:
Sounds good to me @blink1073. Happy to help testing it once merged and a UMD version has been uploaded somewhere :+1:

jasongrout:
Thanks again everyone (and especially @GordonSmith) for working on this. I think it's a really valuable contribution.

jasongrout:
We now have a stable JupyterLab 2.0 release, so I'd like to propose merging this again (looks like a conflict needs to be resolved at this point).

GordonSmith:
@jasongrout rebased and conflict resolved.

blink1073:
image

blink1073:
Thanks @GordonSmith!

PR #44, opened 2020-01-26
Simplified/improved custom rendering of virtual nodes: removed `hpass` and `VirtualElementPass`, added optional `renderer` param

telamonian:
This PR contains a massive simplification and a number of improvements to the implementation of virtual nodes with custom renderers (formerly known as "passthru" virtual nodes) in the virtualdom pkg.

Ideally this PR will get pulled in sometime in the next week. so that I can incorporate its improvements into jlab before the 2.0 release, since it may involve some breaking changes in jlab (see note bellow about breaking changes w.r.t. lumino).

Overview of changes:
- I removed `VirtualElementPass` and `hpass` and folded their implementations into `VirtualElement` and `h`. A custom renderer (what I had formerly been calling a "passthru" renderer) can now be set on any virtual node via an optional arg to either `VirtualElement` or `h`. This has a number of benefits:
    - reduced the number of runtime type checks needed in the virtualdom
implementation
    - no one liked the "pass" or "passthru" names. All uses of such have now been removed
    - we're now back to having only two virtual element types (and one is super simple and just for text). As well, the notion of setting a custom renderer as an optional argument is simpler than having a separate special "escape hatch" virtual element type. This should make the custom renderer stuff easier to pick up, learn, and use

- `renderer` is now an optional property for both `h` and `new VirtualElement`. Since `renderer` is optional, backwards compatibility is preserved for these functions.

- it is now possible for a node to have both a custom renderer and children.When node.renderer.render is called, the node's .attr and .children fields will be passed as options. The render function will then be able to create actual DOM children based on the passed .children field. If the particular render function being used does not support .children, it will just be ignored

- I fixed up and added to all of the related docstrings, with the goal in mind of making it easier to understand what a `VirtualElement.IRenderer` actually is and does. Hopefully that will make it easier for people to create and use their own custom element renderers

Technically, this PR includes breaking changes in the form of the removal of `hpass` and `VirtualElementPass`. However, a search of github reveals that there is currently 0% adoption of either `hpass` or `VirtualElementPass` (aside from my own work in jlab core), so I think this will be safe to pull in.

#### update

This PR now also includes a fix for a minor problem with tabbar icons. Previously, when `title.iconRenderer` was set, `title.iconLabel` would get set as the title attribute of the icon's DOM element. This was not correct.

Instead, `.iconLabel` will now be treated the same regardless of the presence of `.iconRenderer`: `.iconLabel` will be set as a child of the virtual element corresponding to the icon. If `.iconRenderer` is set, it's then the responsibility of `.iconRenderer.render` whether and how to create the actual DOM corresponding to `.iconLabel`.

### DIFF
--- a/packages/virtualdom/src/index.ts
+++ b/packages/virtualdom/src/index.ts
@@ -907,7 +907,7 @@ export function h(tag: string): VirtualElement {
     } else if (arg instanceof Array) {
       extend(children, arg);
     } else if ((i === 1 || i === 2) && arg && typeof arg === 'object') {
-      if ("render" in arg && "unrender" in arg) {
+      if ("render" in arg) {
         renderer = arg;
       }
       else {
@@ -1076,7 +1076,7 @@ export function hpass(tag: string): VirtualElementPass {
   if (arguments.length === 2) {
     const arg = arguments[1];
 
-    if ("render" in arg && "unrender" in arg) {
+    if ("render" in arg) {
       renderer = arg;
     } else {
       attrs = arg;
--- a/packages/virtualdom/src/index.ts
+++ b/packages/virtualdom/src/index.ts
@@ -901,6 +901,8 @@ namespace h {
   interface IFactory {
     (...children: Child[]): VirtualElement;
     (attrs: ElementAttrs, ...children: Child[]): VirtualElement;
+    (renderer: VirtualElement.IRenderer, ...children: h.Child[]): VirtualElement;
+    (attrs: ElementAttrs, renderer: VirtualElement.IRenderer, ...children: h.Child[]): VirtualElement;
   }
 
   export const a: IFactory = h.bind(undefined, 'a');
--- a/packages/virtualdom/src/index.ts
+++ b/packages/virtualdom/src/index.ts
@@ -760,7 +760,8 @@ class VirtualElement {
   }
 }
 
-export namespace VirtualElement {
+export
+namespace VirtualElement {
   /**
    * A type describing a custom element renderer
    */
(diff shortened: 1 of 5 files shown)


telamonian:
Here's what's left for WIP:
- [ ] add fixes to widgets (title and tabbar)
- [ ] fix unittests in virtualdom pkg

telamonian:
This PR now also includes a fix for a minor problem with tabbar icons. Previously, when `title.iconRenderer` was set, `title.iconLabel` would get set as the title attribute of the icon's DOM element. This was not correct.

Instead, `.iconLabel` will now be treated the same regardless of the presence of `.iconRenderer`: `.iconLabel` will be set as a child of the virtual element corresponding to the icon. If `.iconRenderer` is set, it's then the responsibility of `.iconRenderer.render` whether and how to create the actual DOM corresponding to `.iconLabel`.

telamonian:
I looked, and to my surprise there do appear to be [4 external projects that depend on the exact signature of `VIrtualElement.constructor`](https://github.com/search?q=%22new+VirtualElement%22+phosphor&type=Code). Interestingly, only one of them is javascript. Two are scala (by way of scala.js), and the other is F# of all things.

On the other hand, there's a ton of projects that import `VirtualElement` to use as a type (mostly as part of defining a new renderer for a Widget). Direct calls to `new VirtualElement` are rare on purpose: by convention you're supposed to use `h` to construct `VirtualElement` instances. It's beyond the scope of this PR, but would it make sense to split `VirtualElement` into both an interface and a class, and then make the implementation (ie the class) private?

blink1073:
I think we should save API changes for a major release of the virtualdom package, for jlab 3.0.

telamonian:
> I think we should save API changes for a major release of the virtualdom package, for jlab 3.0.

Which API changes do you mean?

1. Removal of `hpass` and `VirtualElementPass`

2. Addition of optional `renderer` param to `h` and `VirtualElement.constructor`
 
3. Addition of optional `options` param to `IRenderer.render`, and the change of `IRenderer.unrender` to optional

1) is technically breaking, which yeah, even if nobody's using them probably still isn't a great idea. I'll add a shim that provides them.

I apologize if I'm trying to push for too much with this PR. To be honest, I'm only really after the improvements to `IRenderer` in 3). Doing 1) and 2) helped me to cleanly implement 3), but I may have gotten a bit carried away.

I would like to get at least the `options` param for `IRenderer.render` pulled in soon, since that will be required for eg `labIcon` to implement `iconLabel` correctly. How should I proceed?

packages/virtualdom/src/index.ts
```diff
@@ -856,15 +904,18 @@ export function h(tag: string): VirtualElement {
...
     } else if (arg instanceof VirtualElement) {
       children.push(arg);
-    } else if (arg instanceof VirtualElementPass) {
-      children.push(arg);
     } else if (arg instanceof Array) {
       extend(children, arg);
-    } else if (i === 1 && arg && typeof arg === 'object') {
-      attrs = arg;
+    } else if ((i === 1 || i === 2) && arg && typeof arg === 'object') {
+      if ("render" in arg && "unrender" in arg) {
```
blink1073: Since `unrender` is optional, we shouldn't check for it here, right?

telamonian: Yes, good catch. Removing the check for `unrender` will weaken the run time type constraint, but I just doubled checked and it looks like `render` is never used as an html attribute, so it should be fine

telamonian: Done. Removing `unrender` from the runtime typecheck made me a little bit nervous since it doesn't leave much to distinguish an `IRenderer` from an attributes object. It should be fine, though, since `render` is never a valid html attribute (I checked the spec)

blink1073:
Published `@lumino/virtualdom@1.5.0`

telamonian:
Sweet. Thanks @blink1073!

PR #46, opened 2020-02-09
IRenderer cleanup; normalize icon fields across all interfaces

telamonian:
needed for jupyterlab/jupyterlab#7864

This PR has two goals:

- I recently got this bit of feedback from one of the `LabIcon` early adopters (@vidartf [via gitter](https://gitter.im/jupyterlab/jupyterlab?at=5e3d3ec7433e1d40399f5d07)):

    > by inspecting the lab source, I figured `iconRenderer` was the attribute to set

    He was talking about figuring out that `Title.iconRenderer` was the field you're supposed to pass a `LabIcon` object into. Which, fair point, it's pretty confusing that `LabIcon` gets passed into the `Foo.icon` field for all interfaces except those defined in Lumino. On top of that, `iconRenderer` as a name is kind of clunky and maybe a bit conceptually confusing. Given all that, I want to experiment with reclaiming the deprecated `Title.icon` field, and see if I can do so in a backwards compatible way

- Given that it's possible to reclaim `Title.icon`, I want to normalize the behavior of the `.icon` field in all of the relevant interfaces across Lumino, so that all icons in Lumino are able to use custom renderers. Once that's done, I'll be able to finally migrate the last few holdout icons in Jupyterlab to `LabIcon` and resolve the remaining issues with that (see jupyterlab/jupyterlab#7864)

There's also a minor fix to the custom renderer implementation in `virtualdom` that needs getting in (the `attrs` and `children` args currently aren't getting passed in to a custom renderer in a consistent fashion)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,10 +7,12 @@ build
 review/api/temp
 *.tsbuildinfo
 
-# jetbrains IDE stuff
+# jetbrains ide stuff
 *.iml
 .idea/
 
-# ms IDE stuff
+# ms ide stuff
+*.code-workspace
+.history
 .vscode
 
--- a/packages/commands/package.json
+++ b/packages/commands/package.json
@@ -34,10 +34,20 @@
     "clean": "rimraf lib && rimraf *.tsbuildinfo",
     "clean:test": "rimraf tests/build",
     "docs": "typedoc --options tdoptions.json src",
+
     "test": "npm run test:firefox",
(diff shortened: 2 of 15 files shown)


telamonian:
I've now implemented everything I described in the top post

blink1073:
Released:

 - @lumino/commands@1.10.0
 - @lumino/virtualdom@1.6.0
 - @lumino/widgets@1.11.0

PR #52, opened 2020-02-16
Constrain tabs to their source DockPanel

### LINKED ISSUE #43: DockPanel inside a DockPanel
Given the following layout (adapted from the [DockPanel example](https://github.com/jupyterlab/lumino/tree/master/examples/example-dockpanel)), where `dock2` is a second `DockPanel` added to the first one with:

```typescript
let dock2 = new DockPanel();
dock2.addWidget(g2);
dock2.addWidget(y2, { mode: 'split-right', ref: g2 });
dock2.id = 'dock2';
dock2.title.label = 'Dock 2';
dock.addWidget(dock2, { mode: 'split-bottom', ref: r1});
```

image

Is is possible to have control on where the widgets of an inner DockPanel can be docked? For example to disallow dropping them outside of "Dock 2":

inner-dockpanel

jtpio:
Fixes #43.

This change constrains tabs in a DockPanel to their source DockPanel. It means that tabs cannot be dragged from one DockPanel and dropped in another one.

This allows using a DockPanel inside another DockPanel, but also two DockPanel side by side like this this example:

dock-source

It is true that creating two (or more) DockPanel can lead to a convoluted UX in some cases, especially when they have the same importance, the same size, or are positioned side by side like in the example above.

However we can imagine having a much "smaller" DockPanel inside a larger web application, for example something similar to this gallery: https://raw.githack.com/hpcc-systems/Visualization/master/demos/gallery/gallery.html?./samples/layout

In that case a user would expect each DockPanel to be self-contained, and it should not be possible to drag a tab from one panel and drop it in another one.

This change adds this constraint by default, since it doesn't alter the case when only one DockPanel is used. But we can also consider making this an opt-in by adding a configurable option that would allow such behavior.

### DIFF
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -516,7 +516,7 @@ class DockPanel extends Widget {
 
     // Show the drop indicator overlay and update the drop
     // action based on the drop target zone under the mouse.
-    if (this._showOverlay(event.clientX, event.clientY) === 'invalid') {
+    if (event.source !== this || this._showOverlay(event.clientX, event.clientY) === 'invalid') {
       event.dropAction = 'none';
     } else {
       event.dropAction = event.proposedAction;
@@ -970,6 +970,7 @@ class DockPanel extends Widget {
       mimeData, dragImage,
       proposedAction: 'move',
       supportedActions: 'move',
+      source: this
     });
 
     // Hide the tab node in the original tab.


vidartf:
> This change adds this constraint by default, since it doesn't alter the case when only one DockPanel is used. But we can also consider making this an opt-in by adding a configurable option that would allow such behavior.

Having it be opt-in would be needed for backwards compatibility, I think.

jtpio:
Fixed by https://github.com/jupyterlab/lumino/pull/137

Issue #51, opened 2020-02-14
Chore: Write tests for RateLimiter-derived classes.


afshin:
This is as much to verify that `Throttler` and `Debouncer` work as intended as it is to have full test coverage.

PR #54, opened 2020-02-20
Update throttler API

afshin:
1. [x] Move `Poll` tests to `poll.spec.ts`.
1. [x] Update `Throttler` to support `leading` and `trailing` configuration. 
1. [x] Write tests for `Debouncer` and `Throttler` in `ratelimiter.spec.ts`.

#### Notes to reviewer(s):
- The tests in `poll.spec.ts` are unchanged, they used to exist in `index.spec.ts`. Only the `ratelimiter.spec.ts` tests are new.
- This API change requires a **minor** version bump.

cf. https://github.com/jupyterlab/lumino/issues/51, https://github.com/jupyterlab/jupyterlab/pull/7895


packages/polling/src/ratelimiter.ts
```diff
@@ -123,13 +123,72 @@ export class Debouncer<T = any, U = any> extends RateLimiter<T, U> {
...
+   * When both `leading` and `trailing` are `true`, `leading` has priority.`
+   */
+  constructor(fn: () => T | Promise<T>, options?: Throttler.IOptions | number) {
+    super(fn, typeof options === 'number' ? options : options && options.limit);
+    let leading = true;
+    let trailing = true;
+    if (typeof options !== 'number') {
+      options = options || {};
+      leading = 'leading' in options ? options.leading! : true;
+      trailing = 'trailing' in options ? options.trailing! : true;
```
ikiw: Is trailing flag required ? Can't it be considered as trailing when the leading is set as false ? With the current way of passing options, the throttler has to be instantiated as `new Throttler(fn, {leading:false, trailing:true}); `to ensure it'll be trailing. Just doing `new Throttler(fn, { trailing:true });  ` will still be considered as leading as by default leading is set to true which will override the options trailing flag.

afshin: I think this probably makes sense, with the current implementation. I was wondering if *both* `leading` and `trailing` could be true, where we privilege the *leading call* and we wait until the *trailing edge* of the cycle. But given the underlying `RateLimiter` I don't see a straightforward way to implement that logic.

afshin: Because if `leading` is `false` and `trailing` is `true`, it would mean that we invoke the *last call* before the *trailing edge* of the cycle, which requires storing invocations in a queue. And since invocations are basically meant to be replaceable with one another, I don't think this makes sense.

cc: @jasongrout

blink1073:
https://github.com/jupyterlab/lumino/blob/master/CHANGELOG.md#2020-03-22

Issue #56, opened 2020-02-25
Command Palette - No easy way to clear search with mouse


ggbhat:
In the current implementation of Command Palette search input is missing clear input option. It is making user not to use single mouse click option to clear input. Forcing user to use keyboard to clear text by selection or backspace.

solution:
We can add input type to search by default and make use of useragent (browser) default clear search feature.
By default in edge clear search is visible, for webkit based browser we need to specify the type to search so that clear input will appear.

<img width="211" alt="searchwithclear" src="https://user-images.githubusercontent.com/4777119/75236812-58ab4200-57e4-11ea-99a2-c299eca3b784.PNG">

I would love to make PR if this solution is fine..

PR #57, opened 2020-02-26
Added type search to command pallet search input

ggbhat:
Fixes - #56 
Jupyter Lab Issue - [6617](https://github.com/jupyterlab/jupyterlab/issues/6617)

Added type search to input to show clear search icon. It will solve usability issue by user should be able to use mouse click to clear search text.

searchwithclear1

### DIFF
--- a/packages/default-theme/style/commandpalette.css
+++ b/packages/default-theme/style/commandpalette.css
@@ -26,12 +26,13 @@
   padding: 4px 6px;
   background: white;
   border: 1px solid #E0E0E0;
+  position: relative;
 }
 
 
 /* <DEPRECATED> */ .p-CommandPalette-input, /* </DEPRECATED> */
 .lm-CommandPalette-input {
-  width: 100%;
+  width: 92%;
   border: none;
   outline: none;
   font-size: 16px;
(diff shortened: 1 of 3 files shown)


blink1073:
cc @tgeorgeux

vidartf:
What happened to our binder bot? :(

I cannot see such a cross when testing here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search  (Win 10, Firefox).

ggbhat:
@vidartf  Yeah, clear search is not available for firefox. Need to add custom clear button then.

ggbhat:
Added custom clear search button since input type search is not works on Firefox.

jasongrout:
> Added custom clear search button since input type search is not works on Firefox.

input type search definitely works in Firefox, since Firefox 4, IIRC. However, having a clear button is one of the differences between browsers: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#Differences_between_search_and_text_types

jasongrout:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=558594 (which has seen recent work) and https://stackoverflow.com/a/46687047

tgeorgeux:
I am overtly +1 on this, provided it's technically sounds (which I am far from qualified to assess).

PR #65, opened 2020-04-24
Customize minimum row and column section sizes for datagrid

nmichaud:
The datagrid currently has a hard-coded minimum size of 10 pixels for row heights and column widths. This PR exposes a setting similar to defaultSizes to allow the user to set the minimum size for each dimension.


blink1073:
See updated [changelog](https://github.com/jupyterlab/lumino/blob/master/CHANGELOG.md#2020-04-24)

PR #66, opened 2020-04-24
Fixes tabsMovable on DockPanel

nmichaud:
The current implementation binds the value of tabsMovable in the DockPanel constructor, so the behavior is inconsistent in the following case:

1. Create a new DockPanel with `tabsMovable=false`.
2. Later change `DockPanel.tabsMovable` to `true`.
3. Newly created TabBars will still set `tabsMovable` to false.

### DIFF
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -71,16 +71,14 @@ class DockPanel extends Widget {
     this._mode = options.mode || 'multiple-document';
     this._renderer = options.renderer || DockPanel.defaultRenderer;
     this._edges = options.edges || Private.DEFAULT_EDGES;
-    if (options.tabsMovable !== undefined) {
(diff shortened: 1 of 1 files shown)


packages/widgets/src/dockpanel.ts
```diff
@@ -71,16 +71,14 @@ class DockPanel extends Widget {
     this._mode = options.mode || 'multiple-document';
     this._renderer = options.renderer || DockPanel.defaultRenderer;
     this._edges = options.edges || Private.DEFAULT_EDGES;
-    if (options.tabsMovable !== undefined) {
-      this._tabsMovable = options.tabsMovable;
-    }
+    this._tabsMovable = options.tabsMovable || true;
```
vidartf, line 74: Won't this flip a `false` to a `true` ? `false || true === true`

vidartf: Fixed in #69 👍

PR #69, opened 2020-05-05
Fix boolean logic when false is specified

nmichaud:
Fixes issue pointed out at the end of this PR: 

https://github.com/jupyterlab/lumino/pull/66#discussion_r420248675


jasongrout:
Can't wait until we can use `this._tabsMovable = options.tabsMovable ?? true`!

PR #71, opened 2020-05-09
Fix `disconnectAll` implementation.

AlbertHilb:
Let `A` be an object which sends and receives signals, `B` an object which sends a signal received by `A`.

References to the object `A` are present as:

1. `signal.sender` property of the connections associated to `A` in the `receiversForSender` `WeakMap`.

1. `thisArg` property of the connections associated to `A` in the `sendersForReceiver` `WeakMap`.

1. `thisArg` property of a connection associated to `B` in the `receiversForSender` `WeakMap`.

Calling `Signal.ClearAll` on `A`, when `A` is disposed, removes the references mentioned in the points 1 and 2, but does'not remove the reference mentioned in 3. This leads to memory leaks ([see here][1]).

This PR should fix this issue.

Regards

[1]: https://github.com/jupyterlab/jupyterlab/issues/8371

### DIFF
--- a/packages/signaling/src/index.ts
+++ b/packages/signaling/src/index.ts
@@ -522,19 +522,10 @@ namespace Private {
    */
   export
   function disconnectAll(object: any): void {
-    // Clear and cleanup any receiver connections.
-    let receivers = receiversForSender.get(object);
-    if (receivers && receivers.length > 0) {
-      each(receivers, connection => { connection.signal = null; });
-      scheduleCleanup(receivers);
-    }
-
-    // Clear and cleanup any sender connections.
-    let senders = sendersForReceiver.get(object);
-    if (senders && senders.length > 0) {
-      each(senders, connection => { connection.signal = null; });
-      scheduleCleanup(senders);
-    }
+    // Remove all connections where the given object is the sender.
+    disconnectSender(object);
+    // Remove all connections where the given object is the receiver.
+    disconnectReceiver(object);
   }
 
   /**


blink1073:
Excellent detective work @AlbertHilb!   Might there be a test we can add to cover this case?

AlbertHilb:
@blink1073 

I fear it's impossible to add such a test without some change to the structure of the module, since the test should have access to `sendersForReceiver` and `receiversForSender` which are variables defined within a non-exported namespace.

blink1073:
Fair enough, these changes are straightforward.  Thanks!

blink1073:
I'll make a release with this tomorrow morning.

jasongrout:
In https://github.com/jupyterlab/lumino/blob/b2803f353ecff4c7b08a93183af05f763111a07d/packages/signaling/src/index.ts#L358-L361 we see that the references 2 and 3 noted above are actually the same thing - the same object is used for both the sendersForReceiver and the receiversForSender weakmaps, so how does taking care of 2 not take care of 3?

AlbertHilb:
@jasongrout 

Let 

- `connection` be the `IConnection` object associated to the signal sent by `B` and received by `A`. 
- `sendersA`, the `IConnection` arrays associated to `A` by `sendersForReceiver` and
- `receiversB`, the `IConnection` array associated to `B` by `receiversForSender`.

Naturally, a reference to `connection` is present in `sendersA` and another is present in `receiversB`.

On calling `Signal.clearData(A)`, `connection.signal` is set to `null`, then all `IConnection` object references pointing to objects with `signal` property set to `null` are removed  from `sendersA`, so in particular the `connection` reference present in `sendersA` is removed. But the `connection` reference present in `receiversB` survives and it keeps alive `connection`, which, in turn, keeps alive `A`.

jasongrout:
I modified a test and stepped through the code to understand this better, and I agree with the fix. Here's my attempt at an explanation:

Here is my test code, using the objects in https://github.com/jupyterlab/lumino/blob/b2803f353ecff4c7b08a93183af05f763111a07d/packages/signaling/tests/src/index.spec.ts#L19 and https://github.com/jupyterlab/lumino/blob/b2803f353ecff4c7b08a93183af05f763111a07d/packages/signaling/tests/src/index.spec.ts#L39

```typescript
        let testSender = new TestObject();
        let testReceiver = new TestHandler();
        testSender.two.connect(testReceiver.onTwo, testReceiver);
        Signal.clearData(testReceiver);
```

The original code in clearData was this:

```typescript
  function disconnectAll(object: any): void {
    // Clear and cleanup any receiver connections.
    let receivers = receiversForSender.get(object);
    if (receivers && receivers.length > 0) {
      each(receivers, connection => { connection.signal = null; });
      scheduleCleanup(receivers);
    }

    // Clear and cleanup any sender connections.
    let senders = sendersForReceiver.get(object);
    if (senders && senders.length > 0) {
      each(senders, connection => { connection.signal = null; });
      scheduleCleanup(senders);
    }
```
In the test case above, after the connection things are initialized so that `receiversForSender` has one entry with key `testSender` and value an array containing the single IConnection object, and the `sendersForReceiver` has one entry with key `testReceiver` and value an array containing the same single IConnection object.

When we execute `Signal.clearData(testReceiver);`, `receiversForSender.get(object)` returns nothing, since `testReceiver` is not a sender, so (crucially) the receivers cleanup is not scheduled. The `sendersForReceiver.get(object)` does correctly return the array of the connection object, and the connection object is correctly flagged as invalid by setting the signal to null, and the senders array is correctly scheduled for cleanup. However, the receiversForSender entry is *not* scheduled for cleanup to remove the invalid connection object.

The main difference between this and using `disconnectReceiver` is that `disconnectReceiver` also looks up the sender and schedules a cleanup for the other place the IConnection object exists in the `receiversForSender` array: https://github.com/jupyterlab/lumino/blob/27d417cf35dbeb68492b24684ee102cd411dd68e/packages/signaling/src/index.ts#L504-L511

In short, I agree with the PR!

jasongrout:
@AlbertHilb - I just saw your comment above! Thanks! I wish I would have seen it before I started diving in. But at least we got the same conclusion! Hopefully my example code and attempt at an explanation is also useful to others that may look at this in the future.

blink1073:
Thanks @jasongrout  and @AlbertHilb  for the diligence!

I released `@lumino/signaling@1.3.6`

PR #82, opened 2020-06-24
fix columnCount signature

mbektasbbg:
fixes the type of region in `columnCount` method


blink1073:
Thanks @mbektasbbg!

blink1073:
https://github.com/jupyterlab/lumino/blob/master/CHANGELOG.md#2020-06-24

PR #94, opened 2020-07-21
Have the DataGrid syncViewport when receiving a DataModel.ChangedArgs signal of type "rows-moved" or "columns-moved"

### LINKED ISSUE #93: DataGrid does not display changes after receving a changed signal from it's DataModel with a type attribute of "rows-moved" or "columns-moved"
Hi, my name is Logan McNichols. I am an intern for Jupyter Cal Poly, working with @kgoo124 and @ryuntalan to develop a tabular data editor powered by the `DataGrid`. On of the editing functions we are including is the ability to move rows and columns.

We have already enabled adding and removing rows and columns. The pattern to do this has been (1) create the desired change to the `DataModel` and (2) emit a `DataModel.ChangedArgs` signal to the `DataGrid` with the appropriate type (e.g. if we were adding a row we would submit a signal with a  `type` attribute `'rows-added'`).

We attempted this same pattern for moving columns and rows. But, submitting a `'rows-moved'` or `'columns-moved'` signal to the `DataGrid` does not seem to update it visually. Interestingly, the grid will update if we move to another panel in JupyterLab and then come back to it.

lmcnichols:
Hi, my name is Logan McNichols. I am an intern for Jupyter Cal Poly, working with @kgoo124 and @ryuntalan to develop a tabular data editor powered by the DataGrid.

We noticed that submitting a `'rows-moved'` or `'columns-moved'` signal to the `DataGrid` does not seem to update it visually. I believe this is because, unlike the methods which handle adding or removing rows or columns, the methods `_onRowsMoved` and `_onColumnsMoved` do not call `_syncViewport() `.

This pull request would replace `_onRowsMoved` and `_onColumnsMoved` call to `_repaintOverlay` with a call to `_syncViewport`. It would also update the comments above these lines of code to reflect the change.

Fixes #93

### DIFF
--- a/packages/datagrid/src/datagrid.ts
+++ b/packages/datagrid/src/datagrid.ts
@@ -2682,8 +2682,8 @@ class DataGrid extends Widget {
       this._repaintRegion('corner-header', r1, 0, r2, Infinity);
     }
 
-    // Schedule a repaint of the overlay.
-    this._repaintOverlay();
+    // Sync the viewport.
+    this._syncViewport();
   }
 
   /**
@@ -2740,8 +2740,8 @@ class DataGrid extends Widget {
       this._repaintRegion('corner-header', 0, c1, Infinity, c2);
     }
 
-    // Schedule a repaint of the overlay.
-    this._repaintOverlay();
+    // Sync the viewport.
+    this._syncViewport();
   }
 
   /**


lmcnichols:
Thank you!

blink1073:
Published as `@lumino/datagrid@0.11.1`

PR #96, opened 2020-07-23
Change the Drag class's private method _moveDragImage to a public method moveDragImage.

### LINKED ISSUE #95: Difficult to extend lumino's Drag class to create custom drag behavior.

We would like to use lumino's `Drag` class to create a column/row shadow which appears on a mouse click of a column/row header and is drag-able.

In order to do this, we would need to extend the `Drag` class to change to how it updates the position of the drag image. For example, when a column header is selected we want to fix the `style.top` attribute to be the top edge of the `DataGrid` and keep `style.left` within the bounds of the `DataGrid`. The method we want to override with our child class is  `_moveDragImage`, but this is currently a **private** method of the `Drag` class.

lmcnichols:
Hi, my name is Logan McNichols. I am an intern for Jupyter Cal Poly. I am working on a tabular data editor powered by the datagrid along with @kgoo124 and @ryuntalan.

We would like to use lumino's `Drag` class to create a column/row shadow which appears on a mouse click of a column/row header and is drag-able. The movement of this shadow can't just follow the cursor wherever it goes, but must be constrained to the bounds of the `DataGrid`. This PR would allow us to do this, by letting us extend `Drag` and make modifications to `moveDragImage`.

This PR would rename `_moveDragImage` to `moveDragImage` and update it's implementations to reflect this change. It would remove the **private** access modifier and relocate `moveDragImage` to the area where the other public methods are.

fixes #95

### DIFF
--- a/packages/dragdrop/src/index.ts
+++ b/packages/dragdrop/src/index.ts
@@ -286,7 +286,7 @@ class Drag implements IDisposable {
    *
    * This is a no-op if there is no drag image element.
    */
-  moveDragImage(clientX: number, clientY: number): void {
+  protected moveDragImage(clientX: number, clientY: number): void {
     if (!this.dragImage) {
       return;
     }
--- a/packages/dragdrop/src/index.ts
+++ b/packages/dragdrop/src/index.ts
@@ -282,6 +282,20 @@ class Drag implements IDisposable {
   }
 
   /**
+   * Move the drag image element to the specified location.
+   *
+   * This is a no-op if there is no drag image element.
+   */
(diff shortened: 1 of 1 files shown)


vidartf:
If the goal is only to be able to *override* the method when subclassed, I would say `protected` is a better middle ground than making it public.

afshin:
@lmcnichols Does Vidar's suggestion of making this method `protected` still satisfy your use case?

lmcnichols:
Yes it still does. I added the `protected` access modifier to `moveDragImage`

blink1073:
Released `@lumino/datagrid@0.12.0`

PR #101, opened 2020-08-01
Modified function spliceArray in datastore/src/listfield.ts so that it behaves like Array.splice on large inputs.

### LINKED ISSUE #100: Function spliceArray in datastore/src/listfield.ts does not behave like Array.splice when items.length > 100000.

We are relying on the `Datastore` to store some data for us. In particular we have two `ListField<number>` objects tracking the position of row and columns respectively. I encountered some unexpected behavior when trying to work with a `ListField<number>` object who's length was greater than 100000. 

I think I traced the problem to the function in `listfield.ts` called `spliceArray`. I copied this function to my workspace and ran the following test.
```
function spliceArray(arr, start, deleteCount, items) {
    if (!items) {
        return arr.splice(start, deleteCount);
    }
    let size = 100000;
    if (items.length < size) {
        return arr.splice(start, deleteCount || 0, ...items);
    }
    let deleted = arr.splice(start, deleteCount);
    let n = Math.floor(items.length / size);
    let idx = 0;
    for (let i = 0; i < n; i++, idx += size) {
        arr.splice(idx, 0, ...items.slice(idx, size));
    }
    arr.splice(idx, 0, ...items.slice(idx));
    return deleted;
}
const arr = new Array(500).fill(0);
const items = new Array(10**6).fill(0);
const start = 500;
const deleteCount = 0;
spliceArray(arr, start, deleteCount, items);
console.log("expected output", 1000500, "actual output", arr.length);

// expected output 1000500 actual output 100500
```

lmcnichols:
Hi, my name is Logan McNichols. I am an intern for Jupyter Cal Poly. I am working on a tabular data editor powered by the DataGrid along with @kgoo124 and @ryuntalan.

We are relying on the Datastore to store some data for us. In particular, we have two `ListField<number>` objects tracking the position of rows and columns respectively. I encountered some unexpected behavior when trying to work with a `ListField<number>` object who's length was greater than 100000.

I think I traced the problem to the function in `listfield.ts` called `spliceArray`. It was producing odd behavior when its argument `items` had length greater than 100000. With a slight modification to line 682, I think it is now having the intended mutation on `arr`.

Fixes #100

### DIFF
--- a/packages/datastore/src/listfield.ts
+++ b/packages/datastore/src/listfield.ts
@@ -681,7 +681,7 @@ namespace Private {
     for (let i = 0; i < n; i++, idx += size) {
       arr.splice(start + idx, 0, ...items.slice(idx, idx + size));
     }
-    arr.splice(idx, 0, ...items.slice(idx));
+    arr.splice(start + idx, 0, ...items.slice(idx));
     return deleted;
   }
 }
--- a/packages/datastore/src/listfield.ts
+++ b/packages/datastore/src/listfield.ts
@@ -679,7 +679,7 @@ namespace Private {
     let n = Math.floor(items.length / size);
     let idx = 0;
     for (let i = 0; i < n; i++, idx += size) {
-      arr.splice(idx, 0, ...items.slice(idx, size));
+      arr.splice(start + idx, 0, ...items.slice(idx, idx + size));
     }
     arr.splice(idx, 0, ...items.slice(idx));
     return deleted;
--- a/packages/datastore/tests/src/listfield.spec.ts
+++ b/packages/datastore/tests/src/listfield.spec.ts
@@ -395,7 +395,7 @@ describe('@lumino/datastore', () => {
       it('should handle large patches', () => {
         let previous = field.createValue();
         let metadata = field.createMetadata();
-        let values = new Array(2**6).fill(0);
+        let values = new Array(2*10**5).fill(0);
         // Create a patch
         let { patch } = field.applyUpdate({
             previous,
@@ -412,7 +412,7 @@ describe('@lumino/datastore', () => {
             metadata,
             patch
         });
-        expect(patched.value).to.eql(values);
+        expect(patched.value.length).to.eql(values.length);
       })
 
     });
--- a/packages/datastore/tests/src/listfield.spec.ts
+++ b/packages/datastore/tests/src/listfield.spec.ts
@@ -392,6 +392,29 @@ describe('@lumino/datastore', () => {
         expect(patchC3.value).to.eql([4]);
       });
 
+      it('should handle large patches', () => {
+        let previous = field.createValue();
+        let metadata = field.createMetadata();
+        let values = new Array(10**6).fill(0);
+        // Create a patch
+        let { patch } = field.applyUpdate({
+            previous,
+            update: { index: 0, remove: 0, values },
+            metadata,
+            version: 1,
+            storeId: 1
+        });
+        // Reset the metadata
+        metadata = field.createMetadata();
+        // Apply the patch
+        let patched = field.applyPatch({
+            previous,
+            metadata,
+            patch
+        });
+        expect(patched).to.eql(values);
+      })
+
     });
 
     describe('unapplyPatch', () => {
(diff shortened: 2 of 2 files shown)


vidartf:
Thanks. Cross-referencing: https://github.com/phosphorjs/phosphor/pull/429

jasongrout:
I think the current version still isn't right. If you change `size` to 5, for example, the typescript compiles to this js:

```js
function spliceArray(arr, start, deleteCount, items) {
    if (!items) {
        return arr.splice(start, deleteCount);
    }
    let size = 5;
    if (items.length < size) {
        return arr.splice(start, deleteCount || 0, ...items);
    }
    let deleted = arr.splice(start, deleteCount);
    let n = Math.floor(items.length / size);
    let idx = 0;
    for (let i = 0; i < n; i++, idx += size) {
        arr.splice(start + idx, 0, ...items.slice(idx, idx + size));
    }
    arr.splice(idx, 0, ...items.slice(idx));
    return deleted;
}
```
Running it with this gives the wrong result:
```js
x = [0,1,2];
spliceArray(x, 1, 0, [100,101,102,103,104,105,106]);
console.log(x);
```
gives me `[ 0, 100, 101, 102, 103, 105, 106, 104, 1, 2 ]`, but should give `[ 0, 100, 101, 102, 103, 104, 105, 106, 1, 2 ]`

I think that perhaps there is a problem with using `Math.floor` and skipping the last step? I haven't checked too closely.

packages/datastore/src/listfield.ts
```diff
@@ -679,7 +679,7 @@ namespace Private {
     let n = Math.floor(items.length / size);
     let idx = 0;
     for (let i = 0; i < n; i++, idx += size) {
-      arr.splice(idx, 0, ...items.slice(idx, size));
+      arr.splice(start + idx, 0, ...items.slice(idx, idx + size));
```
jasongrout, line 682:
```suggestion
      arr.splice(start + idx, 0, ...items.slice(0, size));
```

packages/datastore/src/listfield.ts
```diff
@@ -679,7 +679,7 @@ namespace Private {
...
     }
     arr.splice(idx, 0, ...items.slice(idx));
```
jasongrout, line 684:
```suggestion
    arr.splice(start + idx, 0, ...items);
```

packages/datastore/src/listfield.ts
jasongrout: This should keep popping `size` elements off the front of the `items` list

packages/datastore/src/listfield.ts
jasongrout: This needs to also start at the appropriate place, and we know what is left of `items` is less than `size`.

jasongrout, line 684: Don't forget one more place where we need to shift the starting element.

```suggestion
    arr.splice(start+idx, 0, ...items.slice(idx));
```

lmcnichols: Thanks @jasongrout! I applied that fix to the PR.

jasongrout:
Can you change the test to not be an even multiple of the size, so we are testing that the line after the for loop works?

lmcnichols:
Yes, that makes sense. Size is now set to 250,000.

lmcnichols:
@jasongrout @blink1073 does this look good to you? Anything else I should do?

jasongrout:
Thanks for the ping. Looks good. Thanks!

ian-r-rose:
Thanks for the fix @lmcnichols! This bug was my fault in https://github.com/phosphorjs/phosphor/pull/429 . For context, this is a reimplementation of `Array.splice` that is intended to work for long arrays (except that I made a mistake!). Browsers have a limitation on the number of function args in Javascript which prevents the standard function from working for long arrays.

There is almost certainly a corresponding bug in the string version of this function: https://github.com/jupyterlab/lumino/blob/d1b1b68352dc8211aca7f46a2cbe09e56d253d26/packages/datastore/src/textfield.ts#L704-L721

Would you be willing to make the same fix there?

lmcnichols:
Sure @ian-r-rose! I'll look into that.

PR #107, opened 2020-08-17
Drop deprecated 'p-*' selectors

nmichaud:
Finishes #9. 

This will require careful review.

### DIFF
--- a/packages/widgets/tests/src/commandpalette.spec.ts
+++ b/packages/widgets/tests/src/commandpalette.spec.ts
@@ -687,7 +687,7 @@ describe('@lumino/widgets', () => {
           let name = renderer.createItemClass({ item, indices: null, active: false });
           let expected = 'lm-CommandPalette-item testClass';
           /* <DEPRECATED> */
-          expected = 'lm-CommandPalette-item p-CommandPalette-item testClass';
+          expected = 'lm-CommandPalette-item testClass';
           /* </DEPRECATED> */
           expect(name).to.equal(expected);
         });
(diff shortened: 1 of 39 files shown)


jasongrout:
Thanks.

I think the big decision that wasn't clear from #9 is: is this a change requiring a major version Lumino? If so, I think this shouldn't be the *only* change in Lumino 2, so what is the larger story about what Lumino 2 is going to be?

afshin:
This requires a major version bump, I think.

I don't think we *must* have other API breakages here. This could simply be a transitional version. I think releasing a version of Lumino that has no deprecated APIs is enough of a win, personally.

blink1073:
How would this affect ipywidgets 7?

packages/widgets/tests/src/commandpalette.spec.ts
```diff
@@ -686,9 +686,6 @@ describe('@lumino/widgets', () => {
         it('should create the full class name for the item node', () => {
           let name = renderer.createItemClass({ item, indices: null, active: false });
           let expected = 'lm-CommandPalette-item testClass';
-          /* <DEPRECATED> */
-          expected = 'lm-CommandPalette-item testClass';
```
afshin: Looks like this was a typo to begin with?

packages/widgets/tests/src/commandpalette.spec.ts
```diff
@@ -697,9 +694,6 @@ describe('@lumino/widgets', () => {
           toggledFlag = true;
           let name = renderer.createItemClass({ item, indices: null, active: true });
           let expected = 'lm-CommandPalette-item lm-mod-disabled lm-mod-toggled lm-mod-active testClass';
-          /* <DEPRECATED> */
-          expected = 'lm-CommandPalette-item lm-mod-disabled lm-mod-toggled lm-mod-active testClass';
```
afshin: Same here?

packages/widgets/tests/src/commandpalette.spec.ts
nmichaud: Yeah both of these look like typos in the initial PR.

afshin:
Closing this PR because it is implemented in [this commit](https://github.com/jupyterlab/lumino/pull/319/commits/0a8b1f61534b403eec2832ad44ccf93237f65e16) (authored by @afshin and @nmichaud) in #319

PR #108, opened 2020-08-18
Implement plus button on TabBar

nmichaud:
This PR implements a plus button on the TabBar (similar to google chrome) that will trigger an event when clicked. I've wired up this event for the TabPanel to emit a AddWidgetRequested event, which can be handled by the end user to create the appropriate widget and add it to the panel. A similar approach can also be implemented for the DockPanel if desired.

The CSS is a bit tricky - currently the plus button floats to the end of the TabBar instead of following the tabs.

### DIFF
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -928,7 +928,7 @@ class DockPanel extends Widget {
     tabBar.tabCloseRequested.connect(this._onTabCloseRequested, this);
     tabBar.tabDetachRequested.connect(this._onTabDetachRequested, this);
     tabBar.tabActivateRequested.connect(this._onTabActivateRequested, this);
-    tabBar.tabAddRequested.connect(this._onTabAddRequested, this);
+    tabBar.addRequested.connect(this._onTabAddRequested, this);
 
     // Return the initialized tab bar.
     return tabBar;
@@ -977,7 +977,7 @@ class DockPanel extends Widget {
   /**
    * Handle the `tabAddRequested` signal from the tab bar.
    */
-  private _onTabAddRequested(sender: TabBar<Widget>, args: TabBar.ITabAddRequestedArgs): void {
+  private _onTabAddRequested(sender: TabBar<Widget>, args: void): void {
     this._widgetAddRequested.emit({
       tabBar: sender
     });
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -136,8 +136,8 @@ class DockPanel extends Widget {
    * A signal emitted when the add button on a tab bar is clicked.
    *
    */
-  get widgetAddRequested(): ISignal<this, DockPanel.IWidgetAddRequestedArgs<Widget>> {
-    return this._widgetAddRequested;
+  get addRequested(): ISignal<this, TabBar<Widget>> {
+    return this._addRequested;
   }
 
   /**
@@ -978,9 +978,7 @@ class DockPanel extends Widget {
    * Handle the `tabAddRequested` signal from the tab bar.
    */
   private _onTabAddRequested(sender: TabBar<Widget>, args: void): void {
-    this._widgetAddRequested.emit({
-      tabBar: sender
-    });
+    this._addRequested.emit(sender);
   }
 
   /**
@@ -1057,7 +1055,7 @@ class DockPanel extends Widget {
   private _pressData: Private.IPressData | null = null;
   private _layoutModified = new Signal<this, void>(this);
 
-  private _widgetAddRequested = new Signal<this, DockPanel.IWidgetAddRequestedArgs<Widget>>(this);
+  private _addRequested = new Signal<this, TabBar<Widget>>(this);
 
 }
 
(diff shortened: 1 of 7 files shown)


ellisonbg:
Woah! I have been talking to folks about this recently. Would be spectacular to see this happen. I will dive into this and begin to review soon.

ellisonbg:
In the meantime, can you post a screenshot?

nmichaud:
Here's an example with the DockPanel (not yet part of this PR - I'll push the commits soon)

image

nmichaud:
@ellisonbg @jasongrout @afshin It should be ready to review now. I've also updated the dock panel example to expose an additional context menu item when you right-click on the dock panel to toggle the plus button.

afshin:
I'm happy to look as well.

nmichaud:
@ellisonbg @afshin Any thoughts on this?

nmichaud:
Rebased onto master and force pushed

nmichaud:
Pinging @ellisonbg @jasongrout @afshin again. Any interest in this?

afshin:
Sorry for leaving you hanging with this PR @nmichaud! I'll commit to reviewing this PR this week! Sorry again for such a long wait!

nmichaud:
@afshin  Sure no problem, just wanted to make sure this was still of interest.

packages/widgets/src/tabbar.ts
```diff
@@ -127,6 +128,14 @@ class TabBar<T> extends Widget {
     return this._tabActivateRequested;
   }
 
+  /**
+   * A signal emitted when the tabbar add button is clicked.
+   *
+   */
+  get tabAddRequested(): ISignal<this, TabBar.ITabAddRequestedArgs> {
```
afshin, line 135: I would have been inclined to use this signature `ISignal<this, void>`; did you have a purpose in mind for `TabBar.ITabAddRequestedArgs`?

packages/widgets/src/tabbar.ts
```diff
@@ -865,6 +920,12 @@ class TabBar<T> extends Widget {
       // Clear the drag data.
       this._dragData = null;
 
+      // Handle clicking on the add button
+      if (this.addButtonEnabled && this.addButtonNode.contains(event.target as HTMLElement)) {
+        this._tabAddRequested.emit({});
```
afshin, line 925: (continued from earlier comment) suggestion would be `emit(undefined)` here

packages/widgets/src/tabpanel.ts
```diff
@@ -76,6 +76,8 @@ class TabPanel extends Widget {
     this.tabBar.currentChanged.connect(this._onCurrentChanged, this);
     this.tabBar.tabCloseRequested.connect(this._onTabCloseRequested, this);
     this.tabBar.tabActivateRequested.connect(this._onTabActivateRequested, this);
+    this.tabBar.tabAddRequested.connect(this._onTabAddRequested, this);
+    this.tabBar.tabAddRequested.connect(this._onTabAddRequested, this);
```
afshin, line 80: Duplicate line

packages/widgets/src/tabbar.ts
afshin: I think both here and below, I'd advocate for calling all these signals `addRequested`.

afshin:
@nmichaud thank you so much for your patience!

I am 👍 on this broadly, I left some API surface area suggestions above.

Could you please push you example update as well?

packages/widgets/src/tabbar.ts
nmichaud: Nope, didn't realize void could be used there.

nmichaud:
Thanks for the review. I've updated with the API changes you proposed.

The dockpanel example has already been updated - right click to bring up the context menu, and then select the last menu option ('Toggle add button') to see it in action.

afshin:
> The dockpanel example has already been updated - right click on the DockPanel to bring up the context menu, and then select the last menu option ('Toggle add button') to see it in action.

Oops, I got confused by the screenshot 😂

nmichaud:
> > The dockpanel example has already been updated - right click on the DockPanel to bring up the context menu, and then select the last menu option ('Toggle add button') to see it in action.
> 
> Oops, I got confused by the screenshot 😂

Yeah, apologies I didn't make it obvious in the beginning.

packages/widgets/src/dockpanel.ts
```diff
@@ -129,6 +132,14 @@ class DockPanel extends Widget {
     return this._layoutModified;
   }
 
+  /**
+   * A signal emitted when the add button on a tab bar is clicked.
+   *
+   */
+  get widgetAddRequested(): ISignal<this, DockPanel.IWidgetAddRequestedArgs<Widget>> {
```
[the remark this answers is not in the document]
nmichaud: That works. Using TabBar also makes sense. Will push a fix.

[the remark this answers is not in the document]
nmichaud: OK pushed.

afshin (approved):
Thanks for this feature and for the patience!

PR #111, opened 2020-08-18
Give extending classes access to some of the data grid's paint utilities.

### LINKED ISSUE #110: Give extending classes access to some of the data grid's paint utilities.

We are working on a design that involves painting `LabIcons` onto the `DataGrid`'s canvas to correspond the the type of field in a given column. To achieve this, we would like to be able to extend the `DataGrid` class to add custom drawing functions. To do this, our class would need to be able to access some properties & methods of the `DataGrid` that are currently set to `private`.

The methods we would need access to are
`_paintContent`
`_repaintContent` 
& `_repaintOverlay`

The properties we would need getters for are
- `_canvasGC`
- `_rowSections`
- `_columnSections`
- `_columnHeaderSections`
- `_rowHeaderSections`.

lmcnichols:

We are working on a design that involves painting `LabIcons` onto the `DataGrid`'s canvas to correspond the the type of field in a given column. To achieve this, we would like to provide getters for some of the `DataGrid`'s properties and change some `private` methods to `protected`.

This PR would change the access modifier from `private` to `protected` on the following methods.
- `paintContent`
`repaintContent` 
& `repaintOverlay`

It would also add getters for the following properties.
- `_canvasGC`
- `_rowSections`
- `_columnSections`
- `_columnHeaderSections`
- `_rowHeaderSections`.

Fixes #110

### DIFF
--- a/packages/datagrid/src/datagrid.ts
+++ b/packages/datagrid/src/datagrid.ts
@@ -3772,7 +3772,7 @@ class DataGrid extends Widget {
     }
 
     // Create the paint region object.
-    let rgn: Private.PaintRegion = {
+    let rgn: DataGrid.PaintRegion = {
       region: 'body',
       xMin: x1, yMin: y1,
       xMax: x2, yMax: y2,
@@ -3900,7 +3900,7 @@ class DataGrid extends Widget {
     }
 
     // Create the paint region object.
-    let rgn: Private.PaintRegion = {
+    let rgn: DataGrid.PaintRegion = {
       region: 'row-header',
       xMin: x1, yMin: y1,
       xMax: x2, yMax: y2,
@@ -4022,7 +4022,7 @@ class DataGrid extends Widget {
     }
 
     // Create the paint region object.
-    let rgn: Private.PaintRegion = {
+    let rgn: DataGrid.PaintRegion = {
       region: 'column-header',
       xMin: x1, yMin: y1,
       xMax: x2, yMax: y2,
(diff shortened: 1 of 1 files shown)


packages/datagrid/src/datagrid.ts
```diff
@@ -728,6 +728,41 @@ class DataGrid extends Widget {
       this.dataModel instanceof MutableDataModel;
   }
 
+  /**
+   * The renduring context for painting the data grid.
+   */
+  get canvasGC(): CanvasRenderingContext2D {
```
blink1073, line 734: Do these need to be `public`?  Would `protected` getters work for your use case?

packages/datagrid/src/datagrid.ts
```diff
@@ -728,6 +728,41 @@ class DataGrid extends Widget {
       this.dataModel instanceof MutableDataModel;
   }
 
+  /**
+   * The renduring context for painting the data grid.
```
blink1073, line 732: (sp) rendereing

lmcnichols: Fixed, thanks!

packages/datagrid/src/datagrid.ts
lmcnichols: Yes, `protected` is fine.

nmichaud:
I'm not sure this is necessary - it's possible to leverage a combination of metadata and renderers to draw images/icons onto the grid. What are you trying to do in a subclass?

kgoo124:
@nmichaud We are creating a ghost row/column for the data grid to add new rows/columns. Here is a [prototype](https://github.com/jupytercalpoly/jupyterlab-tabular-data-editor/issues/113) of what we created.

nmichaud:
@kgoo124 ah ok, that looks interesting! Definitely not possible with the current grid code. It looks like you want to draw on the canvas outside of the grid region.

kgoo124:
@nmichaud yes that's right. Since the ghost rows/columns serve as UI to add a new row/column, we want to be able to paint an extra row/column on the grid without having the manage the data model underneath it

PR #114, opened 2020-08-19
mouseDown now uses cell, column, and row selection modes

kgoo124:
I am currently working on the Tabular Data Editor extension with @lmcnichols and @ryuntalan. We noticed the BasicMouseHandler doesn't utilize the different selection modes onMouseDown. This change would set the proper `selectionMode` during the `onMouseDown()` function. There are other places in the mouse handler where the grid is selecting certain areas, so those could also be changed as well.

### DIFF
--- a/packages/datagrid/src/basicmousehandler.ts
+++ b/packages/datagrid/src/basicmousehandler.ts
@@ -203,6 +203,9 @@ class BasicMouseHandler implements DataGrid.IMouseHandler {
         clear = 'all';
       }
 
+      // Use selection mode 'cell'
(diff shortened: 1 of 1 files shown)


blink1073:
cc @nmichaud I released the following changes:

```
 - @lumino/application: 1.10.4 => 1.11.0
 - @lumino/datagrid: 0.12.0 => 0.13.0
 - @lumino/datastore: 0.10.3 => 0.11.0
 - @lumino/default-theme: 0.4.4 => 0.5.0
 - @lumino/widgets: 1.13.4 => 1.14.0
```

PR #116, opened 2020-08-21
Make private _drawCornerHeaderRegion protected drawCornerHeaderRegion

### LINKED ISSUE #115: Difficult to keep corner header from getting painted on when painting on the column header.
Hi, my name is Logan McNichols. I am an intern for Jupyter Cal Poly. I am working on a tabular data editor powered by the DataGrid along with @kgoo124 and @ryuntalan.

Our editor paints `LabIcons` onto the `column-header` of the grid. During a horizontal scroll, these icons can get partially painted onto the corner header. Sometimes, these kind of issues can be solved by bounding the dirty rect very precisely. However, in this case, our canvas is painting on complex svg paths. It would be challenging to bound this object to a given x-coordinate. Alternatively, we could be conservative and say that if any of the svg stands outside of the dirty rect, then don't paint it at all. But this means that our icons will not appear and disappear gradually on a scroll, but will abruptly appear and disappear.

I think the preferable solution is to redraw the corner header as the final draw operation. This is in fact how the `DataGrid` already does it. Can we allow extending classes access to `_drawCornerHeader` so that they can do this as well?

lmcnichols:
Hi, my name is Logan McNichols. I am an intern for Jupyter Cal Poly. I am working on a tabular data editor powered by the DataGrid along with @kgoo124 and @ryuntalan.

This pull request would make the `DataGrid`'s private method `_drawCornerHeaderRegion` a protected `drawCornerHeaderRegion`. It is helpful for extending classes to be able to access this method to repaint the `'corner-header'` as the last paint operation.

Fixes #115

### DIFF
--- a/packages/datagrid/src/datagrid.ts
+++ b/packages/datagrid/src/datagrid.ts
@@ -3653,7 +3653,7 @@ class DataGrid extends Widget {
     this._drawColumnHeaderRegion(rx, ry, rw, rh);
 
     // Draw the corner header region.
-    this._drawCornerHeaderRegion(rx, ry, rw, rh);
+    this.drawCornerHeaderRegion(rx, ry, rw, rh);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


blink1073:
Released as part of ` @lumino/datagrid: 0.13.0 => 0.14.0`

PR #126, opened 2020-10-29
Drag image offset

cameron-toy:
This change adds optional `dragAdjustX` and `dragAdjustY` options for `Drag` objects. This allows the cursor to be somewhere other than the upper left corner of the drag image during a drag.

Without adjust

old_drag

With adjust

new_drag

Background (shameless plug): I originally used this hacked version of lumino dragdrop for the [JupyterLab Interactive Dashboard Editor](https://github.com/jupytercalpoly/jupyterlab-interactive-dashboard-editor) and figured the changes would be useful in core.


afshin:
Hi @cameron-toy! Thanks for the PR! This change seems useful to me.

Could you please:

1. Add docstrings to the implementation in the class?
2. Add these to the tests because they are public API?

PR #128, opened 2020-11-02
Fix sluggish tab dragging in the tab bar.

subhav:
This fixes a typo from commit 45c6f21.

If you need to reproduce the issue, you can roll back an application with a dock panel to widgets@1.9.7 and default-theme@0.1.8, and compare the responsiveness of the tab bar to an application built at the latest versions.

### DIFF
--- a/packages/widgets/style/tabbar.css
+++ b/packages/widgets/style/tabbar.css
@@ -121,7 +121,7 @@
 
 
 /* <DEPRECATED> */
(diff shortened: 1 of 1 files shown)


jasongrout:
Oh, wow, thanks for fixing this typo.

jasongrout (approved):

subhav:
Thanks for the quick review!

PR #129, opened 2020-11-04
Add isToggleable command state

marthacryan:
This allows interfaces (including accessible ones) to indicate to the user that the command may toggle state.


marthacryan:
@ajbozarth

blink1073 (approved):
Thanks!

PR #133, opened 2020-11-23
Batch Add Items to Command Pallette

jhamet93:
Currently, the `CommandPallette` has no item to batch add an Array of commands. Instead, users must add an item one at a time via the `addItem` method. This has a performance impact if the user wants to add a large list of items since `refresh` is called after adding each item. In a 3rd party plugin we are trying to leverage, the performance impact of refreshing after adding every item was making it unusable which has motivated this proposed enhancement to the API.

This pull request adds a new method: `addItems` which allows the user to supplies an Array of items to be added rather than needing to iterate and call `addItem` for each.  Upon all the items being added, `refresh` is called then only once.

### DIFF
--- a/packages/widgets/src/commandpalette.ts
+++ b/packages/widgets/src/commandpalette.ts
@@ -139,6 +139,20 @@ class CommandPalette extends Widget {
   }
 
   /**
+   * Adds command items to the command palette.
+   *
+   * @param items - An array of options for creating each command item.
+   *
+   * @returns The command items added to the palette.
+   */
+  addItems(items: CommandPalette.IItemOptions[]): CommandPalette.IItem[] {
+    const newItems = items.map(item => Private.createItem(this.commands, item));
+    newItems.forEach(item => this._items.push(item));
+    this.refresh();
+    return newItems;
+  }
+
+  /**
    * Remove an item from the command palette.
    *
    * @param item - The item to remove from the palette.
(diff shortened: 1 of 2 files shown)


jasongrout (approved):
I think this makes a lot of sense. Thanks!

jasongrout:
Leaving it open until next week for comments from others (this week a lot of people are off for holidays). If no one has objected or commented, and we haven't merged this next week, will you comment here to remind us to merge it?

jhamet93:
> Leaving it open until next week for comments from others (this week a lot of people are off for holidays). If no one has objected or commented, and we haven't merged this next week, will you comment here to remind us to merge it?

Yep absolutely!

telamonian (approved):
This also LGTM! Once this is in, there's probably a couple of places in core (eg `docmanager-extension`, `filebrowser-extension`, etc) with long lists of commands where this could be helpfully applied

blink1073 (approved):
Thanks!

blink1073:
Published:

```
 - @lumino/application: 1.11.1 => 1.12.0
 - @lumino/commands: 1.11.4 => 1.12.0
 - @lumino/datagrid: 0.14.1 => 0.15.0
 - @lumino/default-theme: 0.5.1 => 0.6.0
 - @lumino/virtualdom: 1.7.3 => 1.8.0
 - @lumino/widgets: 1.14.1 => 1.15.0
```

PR #136, opened 2020-11-28
Add style index.js files to optionally consume the CSS via a js module import

jasongrout:
The webpack css-loader plugin does not deduplicate css that is imported via the css @import. Making the css available via a js module import lets webpack do deduplication at the module level in a way that respects the dependency ordering.

Adding these index.js files should not affect anything in Lumino itself, but just make it easier to consume the Lumino CSS and have deduplication.

This will help fix a problem in JupyterLab, for example, where the Lumino widget CSS will be duplicated on the page dozens of times because many modules require the CSS files directly (so css-loader is used). With this change, modules will be able to require the style/index.js module, so Webpack itself will deduplicate the imports, and the Lumino CSS will only be put on the page once.

The default-theme package needs a more complicated index.js since it uses @import in its index.css (in this case, css-loader would make a copy of the css).

### DIFF
--- a/packages/default-theme/package.json
+++ b/packages/default-theme/package.json
@@ -19,7 +19,8 @@
     "src/*",
     "types/*",
     "images/*.png",
-    "style/*.css"
+    "style/*.css",
+    "style/index.js"
   ],
   "style": "style/index.css",
   "repository": {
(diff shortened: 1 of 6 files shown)


blink1073 (approved):
Thanks!

blink1073:
Released the following:

```
 - @lumino/application: 1.12.0 => 1.13.0
 - @lumino/datagrid: 0.15.0 => 0.16.0
 - @lumino/default-theme: 0.6.0 => 0.7.0
 - @lumino/dragdrop: 1.6.4 => 1.7.0
 - @lumino/widgets: 1.15.0 => 1.16.0
```

PR #137, opened 2020-12-03
Constrain tabs to their source DockPanel (opt-in)

### LINKED ISSUE #43: DockPanel inside a DockPanel

piersdeseilligny:
This pull request is a modification to [this one](https://github.com/jupyterlab/lumino/pull/52), by [jtpio](https://github.com/jtpio). I would have committed there, but obviously don't have write access.

It makes the constraining of tabs to their source dock panel an opt-in feature, through the `tabsConstrained` option and property, and fixes #43.

### DIFF
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -83,7 +83,7 @@ class DockPanel extends Widget {
 
     // Create the delegate renderer for the layout.
     let renderer: DockPanel.IRenderer = {
-      createTabBar: () => this._createTabBar(this._tabsMovable),
+      createTabBar: () => this._createTabBar(),
       createHandle: () => this._createHandle()
     };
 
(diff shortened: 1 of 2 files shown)


afshin:
@piersdeseilligny thank you for updating this logic! :fire:

This adds to the API surface area and modifies behavior, could you please add to the test suite as well to cover these new updates?

piersdeseilligny:
@afshin I unfortunately don't have much experience working with unit tests, but is this latest commit similar to what you're looking for?

blink1073 (approved):
LGTM, thanks!

jtpio:
Nice, thanks @piersdeseilligny for picking this up!

PR #140, opened 2020-12-09
Add text wrapping

ibdafna:
Text wrapping is enabled with a boolean flag (default is false). It will first attempt to split the text into words based on whitespace and wrap that way. If the text can not be split into words, it will apply "classic" wrapping based on characters. 

lumino_text_wrapping


ibdafna:
@jasongrout @blink1073 for awareness

blink1073 (approved):
Thanks!

blink1073:
Released as `@lumino/datagrid: 0.16.1 => 0.17.0`

Issue #144, opened 2020-12-15
Datagrid: Make repaintRegion protected


martinRenou:
The `repaintRegion` method is quite useful for scheduling a repaint of a subpart of the datagrid, exposing this API to subclasses would be nice in my case.

https://github.com/jupyterlab/lumino/blob/e348a0d5120db21099b8d65d002ea2a5f5e21dfe/packages/datagrid/src/datagrid.ts#L2015

I can come up with a PR, if nobody is against the idea.

afshin:
I am :+1: on making this method `protected`.

Just for posterity, can you describe a use case where this would be beneficial?

martinRenou:
This is useful when wanting to redraw a subpart of the grid upon an event like hovering on a cell.

I understand that this API used to be protected, as part of the `repaint` method, but the `repaint` method doesn't take parameters anymore and will repaint the entire grid, which is not optimized.

PR #145, opened 2020-12-15
Make _repaintRegion a protected method

martinRenou:
Fix https://github.com/jupyterlab/lumino/issues/144


afshin (approved):
Cheers @martinRenou!

PR #150, opened 2021-01-13
Allow passing of `tag` into `widget` constructor

telamonian:
This fix is related to jupyterlab/jupyterlab#9491

This PR adds `tag` to the `IOptions` object that can be passed into the `widget` constructor:

- if `node` is unset and:
  - `tag` is unset: the widget's node will be a new `div` element (same as before) 
  - `tag` is set: the widget's node will be a new element of tag `tag`
- if `node` is set, `tag` will always be ignored, since no new element construction will occur

### DIFF
--- a/packages/widgets/src/widget.ts
+++ b/packages/widgets/src/widget.ts
@@ -739,6 +739,14 @@ namespace Widget {
      * The default is a new `<div>`.
      */
     node?: HTMLElement;
+
+    /**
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM, thanks!

blink1073 (approved):
LGTM, thanks!

PR #155, opened 2021-03-08
Fix docs target for polling

### LINKED ISSUE #154: polling docs clobber disposable
The [tdoptions](https://github.com/jupyterlab/lumino/blob/master/packages/polling/tdoptions.json#L7) says:
```
  "out": "../../docs/api/disposable",
```

Will PR...

bollwyvl:
## References
- fixes #154

## Test Procedure
- `yarn docs`
- see that `docs/api/polling/index.html` exists and looks right
- see that `docs/api/disposable/index.html` exists and looks right

## Other Options
If not a bridge too far, I would happily these to this PR:

- add a `yarn docs:check` to verify the integrity of the docs build
- add to docs step on `.github/workflows`
  - the docs _do_ get built, but adding the `check` would avoid future issues
  - upload them as an artifact so they can be reviewed

### DIFF
--- a/packages/polling/tdoptions.json
+++ b/packages/polling/tdoptions.json
@@ -4,7 +4,7 @@
   "target": "es5",
   "module": "es5",
   "lib": ["lib.es2015.d.ts", "lib.es2015.collection.d.ts", "lib.es2015.promise.d.ts", "lib.dom.d.ts"],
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!  I'm about to make a release, so let's press with what we have.

PR #161, opened 2021-03-25
Change one BasicMouseHandler properties from private to protected 

ibdafna:

Whilst the BasicMouseHandler class can be extended, it has two private properties, making extensions tricky to implement. This PR makes the the _pressData property protected instead of private.

### DIFF
--- a/packages/datagrid/src/basicmousehandler.ts
+++ b/packages/datagrid/src/basicmousehandler.ts
@@ -613,7 +613,7 @@ class BasicMouseHandler implements DataGrid.IMouseHandler {
     return this._pressData;
   }
 
-  protected _disposed = false;
+  private _disposed = false;
   protected _pressData: PressData.PressData | null = null;
 }
 
--- a/packages/datagrid/src/basicmousehandler.ts
+++ b/packages/datagrid/src/basicmousehandler.ts
@@ -613,8 +613,8 @@ class BasicMouseHandler implements DataGrid.IMouseHandler {
     return this._pressData;
   }
 
-  private _disposed = false;
-  private _pressData: PressData.PressData | null = null;
+  protected _disposed = false;
+  protected _pressData: PressData.PressData | null = null;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ibdafna:
@jasongrout @blink1073

packages/datagrid/src/basicmousehandler.ts
```diff
@@ -613,8 +613,8 @@ class BasicMouseHandler implements DataGrid.IMouseHandler {
     return this._pressData;
   }
 
-  private _disposed = false;
-  private _pressData: PressData.PressData | null = null;
+  protected _disposed = false;
+  protected _pressData: PressData.PressData | null = null;
```
vidartf, line 617: Any inheriting class should be able to use the isDisposed/pressData methods, no?

vidartf: I mean, I understand why an inheritor might want to *write* to _pressData, but the `_disposed` property should not need to be directly written to ?

ibdafna: You are right - fixed now.

blink1073 (approved):
Thanks!  The windows CI failure is unrelated.  Once we get #165 merged I'll cut a release with this.

packages/datagrid/src/basicmousehandler.ts
```diff
@@ -606,15 +606,8 @@ class BasicMouseHandler implements DataGrid.IMouseHandler {
     return Private.cursorMap[handle];
   }
 
-  /**
-   * Get the current pressData
-   */
-  get pressData(): PressData.PressData | null {
-    return this._pressData;
-  }
-
```
vidartf, line 615: Doesn't this mean that this property will not be available on the public API anymore ?

nmichaud: Yeah agreed. Can this be added back?

blink1073: Anyone able to make a PR?

ibdafna: I'll submit one

ibdafna: @blink1073 to revert this, I'll need to restore the variable name of `pressData` back to `_pressData` as the `get pressData()` and `pressData` attribute have duplicate identifiers. Unless you have other suggestions?

ibdafna: Done in #167

PR #163, opened 2021-03-31
Ctrl-click to toggle single row or column selections

ibdafna:

This PR adds the ability to toggle single row or column selections when clicking on a row or column whilst holding the Ctrl key.

lumino_ctrl_click_toggle_selection


blink1073 (approved):
Thanks!

PR #167, opened 2021-04-09
Restore getter for _pressData

ibdafna:

Reverting the `_pressData` --> `pressData` change and getter removal in #161


blink1073 (approved):
Thanks!  I'll cut a release with this next time I am at a keyboard, which may not be until Monday

Issue #170, opened 2021-04-17
Typing of @lumino/widgets changed from 1.19 to 1.20


fcollonval:
## Description

The trouble comes from a change in the type file generated by the lumino package. In 1.19.0, the `Widget.layout` accessor was translated as property in the type declaration. But now in 1.20.0, the accessor is kept:

![image](https://user-images.githubusercontent.com/8435071/115112688-f46ab500-9f86-11eb-8359-a787eeb25c79.png)

This coupled with the new constrain (_Properties Overriding Accessors (and vice versa) is an Error_) on Typescript v4 will break code. The error is now seen on JupyterLab yet as the lock file is pulling version 1.19.0 of @lumino/widgets.

## Expected behavior

The type between minor version does not change.

Xref: https://github.com/jupyterlab/jupyterlab/issues/10088

> I hit this issue when upgrading the lock file on @jupyterlab/git: https://github.com/jupyterlab/jupyterlab-git/pull/926/checks?check_run_id=2359801487

fcollonval:
Unfortunately #172 did not correct this and @lumino/widgets@1.21.0 is still defining types as in 1.20.0.

fcollonval:
Actually the change in the produced typing file was introduced in TypeScript 3.7: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier

> To detect the issue around accessors, TypeScript 3.7 will now emit get/set accessors in .d.ts files so that in TypeScript can check for overridden accessors.

fcollonval:
So it seems we are stuck with Typescript 3.6, if we want to ensure dependent packages using Typescript 4+ can build without error.

PR #157, opened 2021-03-22
Add Sphinx Documentation

afshin:
Adds minimal sphinx documentation for the Lumino Project

Copies the infrastructure used in JupyterLab to add a Changelog and rendered API docs.

documentation screenshot 1

documentation screenshot 2

PR #172, opened 2021-04-19
Switch to TypeScript 3.9

jtpio:
Fixes jupyterlab/lumino#170

We will have to save TypeScript 4.0 for a major version bump.

For reference, it looks like TS was last updated in https://github.com/jupyterlab/lumino/pull/157


blink1073 (approved):
Thanks!

LTKy:
I still have trouble.
Why is the TS version in package filebrowser not switched?

PR #174, opened 2021-04-20
Fix tabindex for menu bar

marthacryan:
This change is for accessibility - it allows ambulatory users to access the menu bar without using a mouse. Not sure what the motivation behind setting tabindex to -1 was - @sccolbert it looks like you added that in, what was the purpose of that? 

## Code Changes
Changes the tabindex for the menu bar DOM element to be 0 instead of -1.

## User-facing Changes
Allows users to bring focus to the menu bar without using a mouse.

### DIFF
--- a/packages/widgets/src/menubar.ts
+++ b/packages/widgets/src/menubar.ts
@@ -937,7 +937,7 @@ namespace Private {
     /* </DEPRECATED> */
     node.appendChild(content);
     content.setAttribute('role', 'menubar');
-    node.tabIndex = -1;
+    node.tabIndex = 0;
     return node;
   }
 


blink1073 (approved):
Thanks!

marthacryan:
@sccolbert I'm a little unclear on what you mean by context - could you give an example of an action being triggered that would lose context?

sccolbert:
Lets say you trigger an action “paste” and you want that to paste from the
clipboard into the currently focused element. If the menu item is now the
focused element, you’ve lost the context of where you want to paste.

sccolbert:
And just an FYI, the menu hierachy is still traversable via keyboard, it
just doesnt use focus state to do it.

marthacryan:
Oh ok - I'm not sure if this has changed since it was implemented but my understanding is that the commands for notebook cells use the selected cells rather than the cells that are in focus (this would often be the same, but if you click outside of a notebook and change the focus it keeps the selected cell the same) - for example [here](https://github.com/jupyterlab/jupyterlab/blob/765ee24f8923b907dc2a3a6cf158ca40f3e9fab7/packages/notebook/src/actions.tsx#L905) you can see the paste function uses `notebook.activeCell`. I wasn't able to find any that used the actual focus to perform an action, but please let me know if I missed anything!

The menu hierarchy being traversable by keyboard is working, but the issue with this tabindex is reaching the menu without a mouse

PR #175, opened 2021-04-22
Switch back to TypeScript 3.6

jtpio:
Follow-up to #172 

Fixes https://github.com/jupyterlab/lumino/issues/170

Adds a hard-coded `api/index.html` file for now until we can upgrade TypeScript.


blink1073 (approved):
Thanks!

PR #177, opened 2021-04-24
Update poll instantiation logic to normalize frequence max to respect interval

afshin:
- Update `Poll` instantiation behavior to normalize the `frequency.max` value to be the larger of: `options.frequency.max`, `options.frequency.interval`, or the default `max` (30 * 1000 milliseconds).
- Update tests to include new behavior


blink1073 (approved):
👍

PR #178, opened 2021-04-30
Respect closable attribute on title

nmichaud:
Currently tabs on the tab bar always show the close icon, even if the tab isn't closable.


blink1073 (approved):
Thanks!

PR #179, opened 2021-05-04
Fix selection mode

ibdafna:

This PR reverts the changes from commit https://github.com/jupyterlab/lumino/commit/4a685fc4697910daf49ccc7fc6036668b16a8ed5 as they break the `selectionMode` parameter of `BasicSelectionModel`, resulting in 'cell' selection regardless of the parameter passed to the constructor:

```js
new BasicSelectionModel({
    dataModel: model6,
    selectionMode: "row", // won't work
  });
```
That change was arguably not needed to begin with as 'cell' selection mode will select full rows or columns when they are clicked on.


blink1073 (approved):
Thanks!

PR #186, opened 2021-05-25
Use Jupyter releaser

afshin:
- Adds `check_release` job 
- Updates release instructions
- Fixes dead links
- Adds changelog markers
- Adds script to format the changelog with individual package versions
- Adds script to tag the individual packages during draft release
   - We no longer tag during the version update, so the latest tag is the last full release (for the automated changelog)


blink1073 (approved):
Nice!

PR #187, opened 2021-05-25
Make focus consistent with active element in menus

marthacryan:
For the menu bar / menu widgets, these changes make the "active element" consistent with focus, both by putting the "active element" in focus and by setting the "active element" to the current focused element (i.e. if user uses tab to navigate the menu). 

This could use more thorough testing, I've been trying to find any edge cases that cause bugs with this but any extra testing would be appreciated!


blink1073:
Kicking CI

fcollonval (approved):
LGTM thanks @marthacryan

PR #190, opened 2021-06-09
Remove Chrome default scroll on focus

ibdafna:
This PR fixes an issue where Chrome automatically scrolls to the center of an element when the `.focus()` method is triggered on a Node element. This behaviour results in context menus being rendered out of position when they are opened on an element which is not focused.

For reference: https://www.chromestatus.com/feature/5745122025144320


blink1073 (approved):
Looks good, thank you!

PR #197, opened 2021-06-22
Simplify rendering logic for grids with merged cells

ibdafna:
This PR:
- fixes a bug where `config.width` was wrong for the inorder leftmost child of a merged cell hierarchy.
- restores optimization for grids with no merged cell regions
- simplifies rendering logic for grids with merged cell regions

~EDIT: Will confirm when ready.~


ibdafna:
The code in this PR is actually okay. I am seeing a rendering issue, but it's not caused by this change. I will, however, add the bugfix to this PR to reduce clutter.

blink1073 (approved):
Thanks!

PR #203, opened 2021-07-21
Don't sort context menu item by selector

### LINKED ISSUE #202: Don't sort context menu item by selector
### Problem

Currently the context menu elements are sorted _by selector_ and by rank. This breaks the separation of concern and create somewhat strange positioning like the `Paste` item in JupyterLab core:

> In the above example, it would also make more sense to put the `New...` items first (but this is blocked by the selector sort).

### Proposed Solution

It would make sense to sort only by rank (following the separation of concern rule). 

> This probably needs to be implemented as optional constructor argument to avoid breaking the API.

### Additional context

Code at play:

https://github.com/jupyterlab/lumino/blob/cb80c56bfba800e9a8e7c2ec16e8e96f5df36be5/packages/widgets/src/contextmenu.ts#L316-L322

fcollonval:
Fixes #202

### DIFF
--- a/packages/widgets/src/contextmenu.ts
+++ b/packages/widgets/src/contextmenu.ts
@@ -7,26 +7,15 @@
 |
 | The full license is in the file LICENSE, distributed with this software.
 |----------------------------------------------------------------------------*/
-import {
-  ArrayExt, each
-} from '@lumino/algorithm';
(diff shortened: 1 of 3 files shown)


fcollonval:
Error seems unrelated

blink1073 (approved):
Thank you!

PR #206, opened 2021-07-22
Add option to not group context menu item by target

fcollonval:
Follow-up of #203

To actually solve the issue #202, this adds an option to not group items by target when going up in the DOM hierarchy.


blink1073 (approved):
Thank you!

PR #210, opened 2021-07-23
Fixes for Check Release

fcollonval:
Fixes issues in check release job.
Adds Python and Pip Cache actions and fix link cache config.
Moves tagging of versions to before-draft-release to avoid confusing the second changelog check before release.


blink1073 (approved):
Thank you!

PR #217, opened 2021-08-05
Update dependencies

afshin:
Update several `devDependencies` and refresh `yarn.lock` to address security advisories.


blink1073 (approved):
Thank you!

PR #218, opened 2021-08-11
New renderer: HyperlinkRenderer

ibdafna:
This PR adds support for `HyperlinkRenderer` : a custom renderer based on `TextRenderer`, which adds logic for displaying and opening hyperlinks.

As it extends `TextRenderer`, parameters such as `font`, `color` and `background` can all be customised as expected. `HyperlinkRenderer` adds two additional parameters on top - `url` and `urlName`. Both take functions which control the logic of how to arrive at the actual `href` value for the URL, as well as any friendly name for displaying the hyperlink in the grid. In addition, the default text color and font weight for links rendered in `HyperlinkRenderer` are navy blue and **bold**.

Columns rendered with `HyperlinkRenderer` will have a pointer hover cursor to indicate they're links. To open a link, the _Ctrl_ key must be pressed before clicking. Not pressing the key will result in the existing selection behaviour.

lumino_hyperlink

### DIFF
--- a/examples/example-datagrid/src/index.ts
+++ b/examples/example-datagrid/src/index.ts
@@ -11,7 +11,7 @@ import 'es6-promise/auto';  // polyfill Promise on IE
 
 import {
   BasicKeyHandler, BasicMouseHandler, BasicSelectionModel, CellRenderer, CellGroup,
-  DataGrid, DataModel, JSONModel, TextRenderer, MutableDataModel, CellEditor, ICellEditor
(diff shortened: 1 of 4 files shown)


ibdafna:
Test seems to be unrelated.

ibdafna:
Ready for review!

blink1073 (commented):
Very nice, thank you!

blink1073 (approved):

PR #221, opened 2021-08-18
Add auto-resize function for column widths

ibdafna:
This PR adds a function which resizes column widths for `body` and `row` column headers (index and non-index columns) based on the text width of their column names. This allows users to not have to worry about manually resizing the grid columns upon instantiation or pre-determine column widths before instantiation. The API is quite simple takes two optional arguments: 

```typescript
DataGrid.fitColumnNames(area: DataGrid.ColumnFitType = 'all', numCols?: number): void;

type ColumnFitType = 'all' | 'row-header' | 'body';
```

`numCols` allow for capping of the number of columns to be resized.

lumino_autofit

### DIFF
--- a/packages/datagrid/src/datagrid.ts
+++ b/packages/datagrid/src/datagrid.ts
@@ -3773,6 +3773,53 @@ class DataGrid extends Widget {
   }
 
   /**
+   * Resizes row header columns so their text fits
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thank you!

PR #223, opened 2021-08-19
Add 'padding' argument to fitColumnNames function

ibdafna:

This PR adds an additional `padding` argument to the 'fitColumnNames' function added in #221. This will allow users to control how much padding the resize function will add to a given column text when resizing the column width. The default parameter is 15 pixels.


blink1073 (approved):
Thank you!

PR #230, opened 2021-09-14
Add linter

fcollonval:


vidartf:
I would probably add a git config to have `git blame` ignore the lint refactors: https://akrabat.com/ignoring-revisions-with-git-blame/

Other than that: I don't know if we should keep the files in review/api or not, but since they are auto-generated, I do not think we should lint them.

fcollonval:
Thanks for the review @vidartf 

> ignore the lint refactors

Done

> they are auto-generated, I do not think we should lint them.

thanks for the suggestion, i'll do that

blink1073 (approved):
Thank you!

blink1073:
Cutting a release now

PR #231, opened 2021-09-14
Use composition to improve tab switch

fcollonval:
Optionally use composition style for high speed tab switch

### DIFF
--- a/packages/widgets/src/widget.ts
+++ b/packages/widgets/src/widget.ts
@@ -175,7 +175,7 @@ export class Widget implements IMessageHandler, IObservableDisposable {
   set hiddenMode(value: Widget.HiddenMode) {
     if (value !== this._hiddenMode) {
       if (this.isHidden) {
-        if (value === Widget.HiddenMode.Class) {
+        if (value === Widget.HiddenMode.Display) {
           this.addClass('lm-mod-hidden');
           /* <DEPRECATED> */
           this.addClass('p-mod-hidden');
@@ -417,7 +417,7 @@ export class Widget implements IMessageHandler, IObservableDisposable {
     }
     this.clearFlag(Widget.Flag.IsHidden);
     this.node.removeAttribute('aria-hidden');
-    if (this.hiddenMode === Widget.HiddenMode.Class) {
+    if (this.hiddenMode === Widget.HiddenMode.Display) {
       this.removeClass('lm-mod-hidden');
       /* <DEPRECATED> */
       this.removeClass('p-mod-hidden');
@@ -452,7 +452,7 @@ export class Widget implements IMessageHandler, IObservableDisposable {
     }
     this.setFlag(Widget.Flag.IsHidden);
     this.node.setAttribute('aria-hidden', 'true');
-    if (this.hiddenMode === Widget.HiddenMode.Class) {
+    if (this.hiddenMode === Widget.HiddenMode.Display) {
       this.addClass('lm-mod-hidden');
       /* <DEPRECATED> */
       this.addClass('p-mod-hidden');
@@ -746,7 +746,7 @@ export class Widget implements IMessageHandler, IObservableDisposable {
   private _layout: Layout | null = null;
   private _parent: Widget | null = null;
   private _disposed = new Signal<this, void>(this);
-  private _hiddenMode: Widget.HiddenMode = Widget.HiddenMode.Class;
+  private _hiddenMode: Widget.HiddenMode = Widget.HiddenMode.Display;
 }
 
 /**
@@ -792,11 +792,11 @@ export namespace Widget {
     /**
      * Set a 'lm-mod-hidden' class to deal with the widget visibility
      */
-    Class = 0,
+    Display = 0,
     /**
      * Set 'transform' to 'scale(0)' to hide it
      */
-    Composition
+    Scale
   }
 
   /**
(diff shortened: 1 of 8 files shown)


packages/widgets/src/widget.ts
```diff
@@ -749,6 +789,29 @@ namespace Widget {
...
+   */
+  export enum HiddenMode {
+    /**
+     * Set a 'lm-mod-hidden' class to deal with the widget visibility
+     */
+    Class = 0,
+    /**
+     * Set 'transform' to 'scale(0)' to hide it
+     */
+    Composition
```
jasongrout, line 812: I'm curious why this is called the "Composition" mode? For example, "Transform" or "Scale" seem to be the terms actually used in the implementation.

jasongrout: Ah, is it saying that it is using hardware-accelerated composition to hide the window? There's presumably a few different ways to cause the browser to create composition layers, so I would still suggest making this mode name a little more specific. Actually, same with "Class" - setting a class could actually effect the transform method, so maybe "Class" should be "Display" or something to more clearly indicate how it will be hiding the widgets?

fcollonval: I use `Composition` to refer to the goal of using the [pixel composite phase](https://aerotwist.com/blog/the-anatomy-of-a-frame/) to deal with widget visibility (this can actually be achieve with `opacity: 0` too) instead of trigger more expensive actions like style recomputation.

But don't have any problem to use something closer to the implementation.

> I use `transform: scale(0)` instead of `opacity: 0` to ensure click event will be captured by the visible stacked panel.

jasongrout: It feels like those wanting to set these options will want to know more specifically the method used so that they can make performance decisions tied to the current browser performance. Also, if it eventually turns out there is a better way to do a composition-based hiding in some cases, it may have tradeoffs with doing scale-based hiding, so we'd want to keep both methods, which means we'd probably want names closer to the actual implementation.

fcollonval: > Actually, same with "Class" - setting a class could actually effect the transform method, so maybe "Class" should be "Display" or something to more clearly indicate how it will be hiding the widgets?

The current way (referred to by `Class` in this PR) has the nice effect to let the developer choosing the way it will deal with hiding the widget as you mentioned. So although changing the `display` style is the default rule applied, it could be overridden. When using the other mode is actually breaking that paradigm as the developer won't have any freedom. Unfortunately this performance trick works only if the CSS rules don't need to be updated.

But I don't have any strong opinion on the naming. So should I rename the enum to `Display` and `Scale` ?

jasongrout: @blois - you recently were investigating how increasing the number of layers in the browser can possibly increase or decrease performance in Colab. If I understand correctly, this change would mean in JLab each tab would be its own layer. Do you have any comments about how many layers a browser can effectively handle before seeing degradation in performance?

fcollonval: Related comment: on Chromium 90 using transform or opacity was doing the trick directly. But on Firefox 88, I needed to set the style `will-change` to ensure the style were not recomputed.

blois: My advice here is to proceed with caution- this definitely can introduce significant performance benefits.
 - Also consider using `contain: strict` (same cautions apply)
 - The additional layers can introduce typing lag for larger notebooks. Watch for regressions in 'Update layer tree' in Chrome devtools for this.
 - We had mixed results with performance improvements across platforms when leveraging layers. What was a 10x improvement on Linux could regress OSX by 10x. It was important to test broadly.
 - Elements in layers which overlap other elements may introduce additional layers. Adding `will-change` to one element can introduce a dozen layers. https://developer.chrome.com/docs/devtools/evaluate-performance/reference/#layer-borders can help.
 - We had occasional rendering flickers from the layers on OSX. Never got a clean repro for Chrome. We tried twice, two years apart to enable layers in one part of the UI and had to back off because of this. This hasn't stopped us from using layers elsewhere, just cautionary.

In short- layers can be used for performance, and we do, but right now we generally try to avoid them as our larger problem is too many layers.

fcollonval: Thanks a lot @blois to share your knowledge on this.

@jasongrout currently the `will-change` attribute, is set up on stacked panel when there is at least two panels (to add as few of them as possible). From the comment of Pete, what about activating this from user settings?

fcollonval:
@jasongrout would you mind having a second look at this one?

jasongrout:
@fcollonval - I added some documentation and code style suggestions in https://github.com/fcollonval/lumino/pull/1.

I have a few other comments I'll add here for discussion as well.

jasongrout:
> The current way (referred to by Class in this PR) has the nice effect to let the developer choosing the way it will deal with hiding the widget as you mentioned. So although changing the display style is the default rule applied, it could be overridden.

Our current CSS makes this hidden class an `!important` rule, so yes, technically they can override, but then they'd have to also use an important rule, which is awkward. I think it's easier to just say that this display:none CSS rule wasn't designed to be overridden.

jasongrout:
> The rendering action in the above test is decreased from ~250ms to ~100ms.

Nice!! Thanks for documenting it here too.

jasongrout:
> @jasongrout is this PR ready from your point of view?

Yes, thanks!

fcollonval:
Thanks a lot @jasongrout for the PRs and reviews.

PR #239, opened 2021-09-26
Optimise grid rendering

ibdafna:

This PR introduces some rendering optimisations by enabling blitting where possible and reducing the number of paint calls for the entire viewport. There is still room for performance tuning, which I'll hopefully have time to explore in the next few weeks.

I also added a helper function `regionHasMergedCells`. It's not used in this PR, but it will be used in the next PR.


blink1073 (approved):
Thank you!

Issue #151, opened 2021-02-16
Format arrow shortcuts as arrow glyphs


jasongrout:
If a shortcut has an arrow key, right now the text (e.g., "ArrowRight") is printed when the shortcut is formatted for display in https://github.com/jupyterlab/lumino/blob/4f2432290e874b040310d428593abe4afc336066/packages/commands/src/index.ts#L1133-L1165

A better convention, I think, is to replace arrows in the shortcut key with actual unicode arrow glyphs, i.e., "ArrowRight" with [unicode 2192](http://www.unicode-symbol.com/u/2192.html): →

jasongrout:
Here are some lists of more glyphs for specific keys on a mac: https://apple.stackexchange.com/questions/55727/where-can-i-find-the-unicode-symbols-for-mac-functional-keys-command-shift-e

vidartf:
Someone should check that the glyphs are accessible too :)

jasongrout:
It looks like there has been some work on this (thanks!). It looks like conversation around this is now spread across #252, #255, and #256. To have a consolidated place for the conversation, I'm writing my thoughts here.

I wasn't clear above that I meant to suggest that we just change the line at https://github.com/jupyterlab/lumino/blob/4f2432290e874b040310d428593abe4afc336066/packages/commands/src/index.ts#L1164
Instead of always returning `mods + parts.key`, we should instead probably format `parts.key` if `parts.key` is something like `ArrowUp`, `ArrowDown`, etc. In other words, that line would be `return mods + formattedKey` for `formattedKey` would default to `parts.key`, but would be the unicode arrow right symbol if `parts.key` was `ArrowUp`, etc.

It looks like #252 instead pursued the approach of treating arrow keys as essentially modifiers, which caused some issues noted in #255, and further work in this direction was undertaken in #256.

I propose we revert #255 and close #256 and instead just format `parts.key` in the line noted above if `parts.key` is `ArrowUp`, etc. I'm sorry my suggestion wasn't clearer on this in the original description.

[...]

PR #252, opened 2021-10-25
Add arrow glyphs #151

PlatinumCD:
Hi! This is my first time opening a pull request for Lumino. I'm currently writing a Jupyterlab Extension for the company I work at and I'm loving the Jupyterlab ecosystem.

I think the majority of my changes are self explanatory. Here are a few thoughts:
- The arrow order, left up right down, follows the order of unicode symbols.
- I transformed line 1118 into multiple lines because it was growing long.
- I considered changing the 'else if' chain into a switch statement but I did not follow through.

### DIFF
--- a/packages/commands/src/index.ts
+++ b/packages/commands/src/index.ts
@@ -1063,6 +1063,26 @@ export namespace CommandRegistry {
     shift: boolean;
 
     /**
+     * Whether `'ArrowLeft'` appears in the keystroke.
+     */
+    arrowLeft: boolean;
+
+    /**
+     * Whether `'ArrowUp'` appears in the keystroke.
+     */
+    arrowUp: boolean;
+
+    /**
+     * Whether `'ArrowRight'` appears in the keystroke.
+     */
+    arrowRight: boolean;
+
+    /**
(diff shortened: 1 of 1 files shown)


packages/commands/src/index.ts
```diff
@@ -1187,6 +1229,18 @@ export namespace CommandRegistry {
...
+      }
+      if (parts.arrowUp) {
+        mods += '\u2191 ';
+      }
+      if (parts.arrowRight) {
+        mods += '\u2192 ';
+      }
+      if (parts.arrowDown) {
+        mods += '\u2193 ';
+      }
```
krassowski, line 1243: This was only added to the Mac condition, this means that as-is the arrows are not displayed on other operating systems at all. I will fix that.

krassowski: See 526b1438bec0848a5ca91c6458e1e2390b805c99.

jasongrout:
I added some more context and thoughts at https://github.com/jupyterlab/lumino/issues/151#issuecomment-955652297, since that seems like the best place to consolidate the conversation across #151, #252, #255, and #256.

PR #253, opened 2021-10-25
Automated Changelog Entry for 2021.10.25 on master

blink1073:
Automated Changelog Entry for 2021.10.25 on master
```
npm workspace versions:
@lumino/example-datastore: 0.18.0
@lumino/example-accordionpanel: 0.5.0
@lumino/example-dockpanel-amd: 0.4.0
@lumino/example-dockpanel: 0.18.0
@lumino/example-dockpanel-iife: 0.4.0
@lumino/example-datagrid: 0.29.0
@lumino/default-theme: 0.20.0
@lumino/virtualdom: 1.14.0
@lumino/coreutils: 1.11.0
@lumino/domutils: 1.8.0
@lumino/widgets: 1.29.0
@lumino/commands: 1.18.0
@lumino/polling: 1.9.0
@lumino/datagrid: 0.33.0
@lumino/collections: 1.9.0
@lumino/algorithm: 1.9.0
@lumino/application: 1.26.0
@lumino/messaging: 1.10.0
@lumino/dragdrop: 1.13.0
@lumino/signaling: 1.10.0
@lumino/disposable: 1.10.0
@lumino/properties: 1.8.0
@lumino/datastore: 0.17.0
@lumino/keyboard: 1.8.0
```

After merging this PR run the "Full Release" Workflow on your fork of `jupyter_releaser` with the following inputs
| Input  | Value |
| ------------- | ------------- |
| Target | jupyterlab/lumino  |
| Branch  | master  |
| Version Spec | 2021.10.25 |
| Since | @lumino/algorithm@1.8.0 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,13 @@
 
 ### Enhancements made
 
-- Add arrow glyphs #151 [#252](https://github.com/jupyterlab/lumino/pull/252) ([@PlatinumCD](https://github.com/PlatinumCD))
-- Added PointerEvents to splitpanel [#251](https://github.com/jupyterlab/lumino/pull/251) ([@martaszmit](https://github.com/martaszmit))
+- Add arrow glyph handling to command registry [#252](https://github.com/jupyterlab/lumino/pull/252) ([@PlatinumCD](https://github.com/PlatinumCD))
+- Added `PointerEvents` handling to `SplitPanel` [#251](https://github.com/jupyterlab/lumino/pull/251) ([@martaszmit](https://github.com/martaszmit))
 - Ignore `keydown` events for modifier keys when accumulating key sequence [#245](https://github.com/jupyterlab/lumino/pull/245) ([@ph-ph](https://github.com/ph-ph))
 
 ### Bugs fixed
 
-- Update title in AccordionPanel [#249](https://github.com/jupyterlab/lumino/pull/249) ([@hbcarlos](https://github.com/hbcarlos))
+- Update title appropriately in `AccordionPanel` [#249](https://github.com/jupyterlab/lumino/pull/249) ([@hbcarlos](https://github.com/hbcarlos))
 
 ### Maintenance and upkeep improvements
 


jasongrout:
@blink1073 - I'm trying out releaser (for the first time! :). I followed the directions at https://jupyter-releaser.readthedocs.io/en/latest/get_started/making_first_release.html to try to generate a draft changelog. However, when I started the draft changelog workflow (with a version spec of 2021.11.04, similar to what you have here), I got an error in the npm version step: https://github.com/jasongrout/jupyter_releaser/runs/4108602970?check_suite_focus=true#step:4:141

Can you give me a pointer to where to dig for more info? Is this a known issue, or is this a problem with how I did things, or something else?

blink1073:
This repo is a bit different because we are using independent versioning, there is a manual pre-step to set the versions, outlined in [RELEASE.md](https://github.com/jupyterlab/lumino/blob/master/RELEASE.md#using-jupyter_releaser)

jasongrout:
Ah, okay, thanks. I'll look there.

jasongrout:
> This repo is a bit different because we are using independent versioning, there is a manual pre-step to set the versions, outlined in [RELEASE.md](https://github.com/jupyterlab/lumino/blob/master/RELEASE.md#using-jupyter_releaser)

So I need to push the new versions to master using that step, THEN run the draft changelog workflow?

jasongrout:
> So I need to push the new versions to master using that step, THEN run the draft changelog workflow?

Ah, yes, I see you pushed to master in https://github.com/jupyterlab/lumino/commit/6fa82516dcd70f9b5ba0646be648f94f6d2cc14d, just before merging this.

jasongrout:
I still see the error even after that prep step. It seems that running `npm version --git-tag-version false 2021.11.04` locally works with node 16, but not with node 14 (which is the node version the workflow is using).

blink1073:
We can update that [here](https://github.com/jupyter-server/jupyter_releaser/blob/25f51b41f2573db88eb1d8c45a7b32d4a382c73d/.github/actions/common/action.yml#L15), if you make a PR, I'll cut a release of releaser.

jasongrout:
> It seems that running `npm version --git-tag-version false 2021.11.04` locally works with node 16, but not with node 14 (which is the node version the workflow is using).

My fault, I didn't reset git changes before running the command with node 14. It's something else, I'll keep investigating.

jasongrout:
Looks like the problem was that I was giving `2021.11.04` as a version, but that isn't a valid semver version. Instead I needed to be doing `2021.11.4`.

blink1073:
Yeah, that has bitten me before too

jasongrout:
Perhaps that is our built-in burn-out prevention program? No releases the first 9 days of the month? :)

PR #256, opened 2021-10-30
Fix ignored arrow keys

### LINKED ISSUE #255: Change in arrow keys handling broke contract
## Description

#252 broke navigation with arrow keys, selecting cells with arrow keys and all other operations using arrow keys in JupyterLab (which for now is only visible on JupyterLab Desktop https://github.com/jupyterlab/jupyterlab-desktop/issues/320). This is because the addition of:

https://github.com/jupyterlab/lumino/blob/0f7378b62338868a53d0e6418d333f7e718d7411/packages/commands/src/index.ts#L1138-L1145

means that:

https://github.com/jupyterlab/lumino/blob/0f7378b62338868a53d0e6418d333f7e718d7411/packages/commands/src/index.ts#L1146-L1148

will not be executed for arrow keys, and all existing logic that relayed on arrow keys being there will fail.

## Reproduce

See https://github.com/jupyterlab/jupyterlab-desktop/issues/320.

## Expected behavior

1) this was released in a minor version, so we should fix it quickly; we should not change the contract in minor versions
2) arrow keys are not proper modifiers and should not be excluded from `key`

krassowski:
Fixes #255

### DIFF
--- a/packages/commands/src/index.ts
+++ b/packages/commands/src/index.ts
@@ -1121,6 +1121,17 @@ export namespace CommandRegistry {
     let arrowRight = false;
     let arrowDown = false;
     for (let token of keystroke.split(/\s+/)) {
+      // if an arrow is present it can be assigned to `key`
+      if (token === 'ArrowLeft') {
+        arrowLeft = true;
+      } else if (token === 'ArrowUp') {
+        arrowUp = true;
+      } else if (token === 'ArrowRight') {
+        arrowRight = true;
+      } else if (token === 'ArrowDown') {
+        arrowDown = true;
+      }
+      // but do not assign proper modifiers to `key`
       if (token === 'Accel') {
         if (Platform.IS_MAC) {
           cmd = true;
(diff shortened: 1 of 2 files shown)


krassowski:
The fix is good, but I also noted: https://github.com/jupyterlab/lumino/pull/252/files#r739683874 and will improve that as well here.

krassowski:
@blink1073 would it be a good idea to add Mac to CI matrix, or was it purposefully excluded?

https://github.com/jupyterlab/lumino/blob/0f7378b62338868a53d0e6418d333f7e718d7411/.github/workflows/tests.yml#L10-L11

jasongrout:
I added some more context and thoughts at https://github.com/jupyterlab/lumino/issues/151#issuecomment-955652297, since that seems like the best place to consolidate the conversation across #151, #252, #255, and #256.

blink1073:
> @blink1073 would it be a good idea to add Mac to CI matrix, or was it purposefully excluded?

I don't remember the context at the time, but yes, we should be testing against Mac.

PR #258, opened 2021-11-02
Format keyboard shortcuts according to OS conventions

### LINKED ISSUE #151: Format arrow shortcuts as arrow glyphs

### LINKED ISSUE #255: Change in arrow keys handling broke contract

jasongrout:
Fixes #151
Fixes #255
Closes #256

This formats keyboard shortcuts to OS conventions on Windows and macOS. It reverts #252 and supersedes #256.

I checked with VoiceOver, and it appears that these glyphs are recognized correctly on Catalina when speaking shortcuts on menu items, etc. (i.e., the curvy arrow is correctly pronounced "return" when reading the keyboard shortcut). I pulled the Windows formatting conventions by looking at how custom keyboard shortcuts are formatted in Microsoft Word (IIRC), but it would be good to double-check those.

### DIFF
--- a/packages/commands/src/index.ts
+++ b/packages/commands/src/index.ts
@@ -1063,26 +1063,6 @@ export namespace CommandRegistry {
     shift: boolean;
 
     /**
-     * Whether `'ArrowLeft'` appears in the keystroke.
-     */
-    arrowLeft: boolean;
-
-    /**
-     * Whether `'ArrowUp'` appears in the keystroke.
-     */
-    arrowUp: boolean;
-
-    /**
-     * Whether `'ArrowRight'` appears in the keystroke.
-     */
-    arrowRight: boolean;
-
-    /**
-     * Whether `'ArrowDown'` appears in the keystroke.
-     */
-    arrowDown: boolean;
-
-    /**
      * The primary key for the keystroke.
      */
     key: string;
@@ -1116,10 +1096,6 @@ export namespace CommandRegistry {
     let cmd = false;
     let ctrl = false;
     let shift = false;
-    let arrowLeft = false;
-    let arrowUp = false;
-    let arrowRight = false;
-    let arrowDown = false;
     for (let token of keystroke.split(/\s+/)) {
       if (token === 'Accel') {
         if (Platform.IS_MAC) {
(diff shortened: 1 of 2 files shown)


packages/commands/src/index.ts
```diff
@@ -1492,6 +1429,45 @@ namespace Private {
...
+      return MAC_DISPLAY.hasOwnProperty(key) ? MAC_DISPLAY[key] : key;
+    } else {
+      return WIN_DISPLAY.hasOwnProperty(key) ? WIN_DISPLAY[key] : key;
+    }
+  }
+
+  const MAC_DISPLAY: { [key: string]: string } = {
+    Backspace: '⌫',
+    Tab: '⇥',
+    Enter: '↩',
```
krassowski, line 1443: ⏎ might have a closer semantic meaning, but will often render as a less familiar icon

krassowski (approved):
Looks good to me, especially for the current use case of displaying the shortcuts as labels.

For better accessibility we could implement [aria-keyshortcuts](https://w3c.github.io/aria/#aria-keyshortcuts). I guess that would be another function taking the keystroke and returning the aria-label (or maybe `formatKeystroke` could return both `{label: string, aria: string}` in a future major version). If it makes sense I will open a new issue to track this accessibility improvement.

packages/commands/src/index.ts
jasongrout: In VoiceOver, your character ⏎ is read as "return symbol", whereas the one in the PR (↩) is read as "return". I think the one in the PR is the official one to use.

jasongrout:
> If it makes sense I will open a new issue to track this accessibility improvement.

+1

Huh, the Aria Shortcuts document seems to vindicate a decision we made about how shortcuts are specified that sometimes people call us out on as controversial or confusing:

> If the character used is determined by a modifier key, the author MUST specify the actual key used to generate the character, that is generated by the key, and not the resulting character. This convention enables the assistive technology to accurately convey what keys must be used to generate the shortcut. For example, on most U.S. English keyboards, the percent sign "%" can be input by pressing Shift+5. The correct way to specify this shortcut is "Shift+5". It is incorrect to specify "%" or "Shift+%".

jasongrout:
Thanks for the review @krassowski!

PR #259, opened 2021-11-02
Run tests on macOS.

jasongrout:
In #256, we realized that macOS tests were not running. This turns them on.

It also switches us to running tests on headless firefox by default.

For now, I exclude macos and firefox-headless, which seems to fail. I've opened https://github.com/jupyterlab/lumino/issues/262 to address this issue.


fcollonval (approved):
Thanks @jasongrout

jasongrout:
Interestingly, chrome on macOS passes, but firefox does not.

PR #276, opened 2021-12-26
fix(TabBar):  Event forwarding fails when hosted in a ShadowRoot

GordonSmith:

### DIFF
--- a/packages/widgets/src/docklayout.ts
+++ b/packages/widgets/src/docklayout.ts
@@ -53,6 +53,7 @@ export class DockLayout extends Layout {
     if (options.spacing !== undefined) {
       this._spacing = Utils.clampDimension(options.spacing);
     }
+    this._document = options.document || document;
     this._hiddenMode =
       options.hiddenMode !== undefined
         ? options.hiddenMode
@@ -1094,7 +1095,7 @@ export class DockLayout extends Layout {
    */
   private _createTabBar(): TabBar<Widget> {
     // Create the tab bar using the renderer.
-    let tabBar = this.renderer.createTabBar();
+    let tabBar = this.renderer.createTabBar(this._document);
 
     // Enforce necessary tab bar behavior.
     tabBar.orientation = 'horizontal';
@@ -1140,6 +1141,7 @@ export class DockLayout extends Layout {
   private _dirty = false;
   private _root: Private.LayoutNode | null = null;
   private _box: ElementExt.IBoxSizing | null = null;
+  private _document: Document | ShadowRoot;
   private _hiddenMode: Widget.HiddenMode;
   private _items: Private.ItemMap = new Map<Widget, LayoutItem>();
 }
@@ -1152,6 +1154,13 @@ export namespace DockLayout {
    * An options object for creating a dock layout.
    */
   export interface IOptions {
+    /**
+     * The document to use with the dock panel.
+     *
+     * The default is the global `document` instance.
+     */
+    document?: Document | ShadowRoot;
+
     /**
      * The method for hiding widgets.
      *
@@ -1181,7 +1190,7 @@ export namespace DockLayout {
      *
      * @returns A new tab bar for a dock layout.
      */
-    createTabBar(): TabBar<Widget>;
+    createTabBar(document?: Document | ShadowRoot): TabBar<Widget>;
 
     /**
      * Create a new handle node for use with a dock layout.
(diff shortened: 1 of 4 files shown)


packages/widgets/src/docklayout.ts
```diff
@@ -1181,7 +1190,7 @@ export namespace DockLayout {
      *
      * @returns A new tab bar for a dock layout.
      */
-    createTabBar(): TabBar<Widget>;
+    createTabBar(document: Document | ShadowRoot): TabBar<Widget>;
```
blink1073, line 1193: I think this argument should be optional

packages/widgets/src/tabbar.ts
```diff
@@ -156,6 +157,13 @@ export class TabBar<T> extends Widget {
    */
   readonly renderer: TabBar.IRenderer<T>;
 
+  /**
+   * The document to use with the tab bar.
+   *
+   * The default is the global `document` instance.
+   */
+  document: Document | ShadowRoot;
```
blink1073, line 165: I think this should be readonly, and only settable in the constructor.

blink1073 (changes requested):
Thanks!  Just a couple of suggestions.

packages/widgets/src/docklayout.ts
GordonSmith: Change made as suggested  (FWIW it only actually gets called from one place, within the DockPanel),

packages/widgets/src/tabbar.ts
GordonSmith: Change made as suggested.

GordonSmith:
@blink1073 FYI these changes are needed so that lumino widgets can be used within custom web components, which in turn can expose lumino widgets via "simple" html, see POC repo here (not currently published):  
* https://gordonsmith.github.io/hpcc-js-wc/components/splitter.html
* https://gordonsmith.github.io/hpcc-js-wc/components/dockpanel.html

This should greatly simplify the use of lumino widgets within frameworks like React, Vue and Angular.

blink1073:
> This should greatly simplify the use of lumino widgets within frameworks like React, Vue and Angular.

Very nice!

blink1073:
@GordonSmith, are you ready for me to make a release?

GordonSmith:
@blink1073 no real rush as I am building off local branches for now.

GordonSmith:
@blink1073 I might request a build if your not planning to do a release for a while?

Did you have any thoughts on the above examples?

blink1073:
The examples are cool!  Happy to feature them as part of the https://github.com/jupyterlab/lumino#external-examples if you want to submit a PR.  I'll cut a release on Monday (I don't like to make releases on a Friday).

packages/widgets/src/docklayout.ts
```diff
@@ -2184,7 +2193,7 @@ namespace Private {
     renderer: DockLayout.IRenderer
   ): TabLayoutNode {
     // Create the tab bar for the layout node.
-    let tabBar = renderer.createTabBar();
+    let tabBar = renderer.createTabBar(this._document);
```
hbcarlos, line 2196: Does this exist in this context?
This function is inside the Private namespace.

hbcarlos: I'm updating lumino in JupyterLab, and there is an issue with `this._document`, when changing to a single-mode document.
https://github.com/jupyterlab/jupyterlab/pull/11823

GordonSmith: Do you have an example of the breaking code - I couldn't see anything related in the failing visual regression test?

GordonSmith: Fix incomming.

GordonSmith: FYI:  https://github.com/jupyterlab/lumino/pull/281

hbcarlos: Hey @GordonSmith, thanks for the quick reply.

> Do you have an example of the breaking code

Yes, when in JupyterLab you change the view to a single-mode document, there is an error on this line:
https://github.com/jupyterlab/jupyterlab/blob/3b21d568035f7383dfd88332e051277130aea7d8/packages/application/src/shell.ts#L557

Following the trace, you can see that the error comes from
https://github.com/jupyterlab/lumino/blob/948eec9353d06ff9808dc127c888a49a9930ca75/packages/widgets/src/docklayout.ts#L2196

where `this` is null.

PR #277, opened 2021-12-29
Prevent opening an empty menu

hbcarlos:
Solves https://github.com/jupyterlab/jupyterlab/pull/11650

As commented in https://github.com/jupyterlab/jupyterlab/pull/11650#discussion_r766563571, there is a minor glitch on the menu when it is empty.
This PR prevents the menu bar from adding a menu to the DOM if the menu is empty.

### DIFF
--- a/examples/example-dockpanel/src/index.ts
+++ b/examples/example-dockpanel/src/index.ts
@@ -280,10 +280,15 @@ function main(): void {
   menu3.title.label = 'View';
   menu3.title.mnemonic = 0;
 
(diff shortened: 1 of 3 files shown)


hbcarlos:
I also added an empty menu in one of the examples.

fcollonval:
Would you mind adding a test too for this?

hbcarlos:
Sure. I'll add tests.

blink1073 (approved):
Nice, thanks!

PR #279, opened 2022-01-08
chore(Widget):  Simplify attach/detach sanity checking

GordonSmith:
Also makes compatible with shadowDom

### DIFF
--- a/packages/widgets/src/widget.ts
+++ b/packages/widgets/src/widget.ts
@@ -1070,10 +1070,10 @@ export namespace Widget {
     if (widget.parent) {
       throw new Error('Cannot attach a child widget.');
     }
-    if (widget.isAttached || document.body.contains(widget.node)) {
+    if (widget.isAttached || widget.node.isConnected) {
       throw new Error('Widget is already attached.');
     }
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks @GordonSmith 

Out of curiosity do you know if the `isConnected` is faster than contains?

> caniuse [report 93%](https://caniuse.com/?search=isConnected) for that feature.

GordonSmith:
@fcollonval I don't know if it is faster, but realistically it couldn't be slower? (not that it matters as this is test is prior modifiying  items in the DOM which will cause reflow etc. which will be orders of magnitudes slower than either...)
Also worth noting that the "contains" test fails for shadowDom elements...

fcollonval:
> I don't know if it is faster, but realistically it couldn't be slower?

Definitely not - it was purely for my information. I support at 100% what you are doing; I really think that moving to web components would be a great way forward to increase interest in Lumino widgets.

> Note2: It is not supported by ie11 if that is an issue?

No it is not an issue. As I checked support by browsers, I added a link if people are wondering.

blink1073 (approved):
Thank you!

PR #281, opened 2022-01-12
fix(DockLayout):  Invalid use of "this" in private namespace

GordonSmith:

### DIFF
--- a/packages/widgets/src/docklayout.ts
+++ b/packages/widgets/src/docklayout.ts
@@ -343,10 +343,14 @@ export class DockLayout extends Layout {
 
     // Create the root node for the new config.
     if (mainConfig) {
-      this._root = Private.realizeAreaConfig(mainConfig, {
-        createTabBar: () => this._createTabBar(),
-        createHandle: () => this._createHandle()
-      });
+      this._root = Private.realizeAreaConfig(
+        mainConfig,
+        {
+          createTabBar: () => this._createTabBar(),
+          createHandle: () => this._createHandle()
+        },
+        this._document
+      );
     } else {
       this._root = null;
     }
@@ -1482,13 +1486,14 @@ namespace Private {
    */
   export function realizeAreaConfig(
     config: DockLayout.AreaConfig,
-    renderer: DockLayout.IRenderer
+    renderer: DockLayout.IRenderer,
+    document: Document | ShadowRoot
   ): LayoutNode {
     let node: LayoutNode;
     if (config.type === 'tab-area') {
-      node = realizeTabAreaConfig(config, renderer);
+      node = realizeTabAreaConfig(config, renderer, document);
     } else {
-      node = realizeSplitAreaConfig(config, renderer);
+      node = realizeSplitAreaConfig(config, renderer, document);
     }
     return node;
   }
@@ -2190,10 +2195,11 @@ namespace Private {
    */
   function realizeTabAreaConfig(
     config: DockLayout.ITabAreaConfig,
-    renderer: DockLayout.IRenderer
+    renderer: DockLayout.IRenderer,
+    document: Document | ShadowRoot
   ): TabLayoutNode {
     // Create the tab bar for the layout node.
-    let tabBar = renderer.createTabBar(this._document);
+    let tabBar = renderer.createTabBar(document);
 
     // Hide each widget and add it to the tab bar.
     each(config.widgets, widget => {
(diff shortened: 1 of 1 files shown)


packages/widgets/src/docklayout.ts
```diff
@@ -2193,7 +2193,7 @@ namespace Private {
     renderer: DockLayout.IRenderer
   ): TabLayoutNode {
     // Create the tab bar for the layout node.
-    let tabBar = renderer.createTabBar(this._document);
+    let tabBar = renderer.createTabBar();
```
fcollonval, line 2196: Would it be better to be able to pass an optional document to `realizeTabAreaConfig` to be forward to `createTabBar`?

GordonSmith: Maybe - I am struggling to see where its used (as it doesn't seem to be hit in the tests and the old version of TypeScript isn't helpful).
I will have more time to dig into it in the morning, if your not looking for a quick fix today,

packages/widgets/src/dockpanel.ts
```diff
@@ -1400,7 +1400,7 @@ export namespace DockPanel {
      *
      * @returns A new tab bar for a dock panel.
      */
-    createTabBar(document?: Document | ShadowRoot): TabBar<Widget> {
+    createTabBar(document: Document | ShadowRoot): TabBar<Widget> {
```
fcollonval, line 1403: We should keep it optional as this is part of the public API.

```suggestion
    createTabBar(document?: Document | ShadowRoot): TabBar<Widget> {
```

fcollonval (changes requested):
Thanks a lot for finalizing this one.

We should not change the signature of `createTabBar` as its part of the public API. Otherwise this is looking good.

packages/widgets/src/docklayout.ts
GordonSmith: @fcollonval:  I changed the code as you suggested and did a bit more sanity checking...

Also I added TabPanel support to the web-components:  https://gordonsmith.github.io/hpcc-js-wc/src/layouts/lumino/tabPanel.html

fcollonval (approved):
LGTM thanks @GordonSmith

blink1073:
Thanks!  I'll cut a bugfix release now.

packages/widgets/src/docklayout.ts
```diff
@@ -343,10 +343,14 @@ export class DockLayout extends Layout {
...
     if (mainConfig) {
-      this._root = Private.realizeAreaConfig(mainConfig, {
-        createTabBar: () => this._createTabBar(),
-        createHandle: () => this._createHandle()
-      });
+      this._root = Private.realizeAreaConfig(
+        mainConfig,
+        {
+          createTabBar: () => this._createTabBar(),
+          createHandle: () => this._createHandle()
```
fcollonval, line 350: I'm sorry @GordonSmith I have a doubt due to this intermediate renderer definition. Actually, the `this._document` pass here is not use as these two callbacks are not passing through the received argument (and it does not make sense to allow a document other than `this._document` to be passed).
And as this is the only place `Private.realizeAreaConfig` is called, it sounds that your initial commit only removing `this._document` in `Private.realizeTabAreaConfig` is the only needed change, isn't it?

GordonSmith: I think its ok as they are just there to satisfy the IRenderer and resolve back onto itself:
```
  private _createTabBar(): TabBar<Widget> {
    // Create the tab bar using the renderer.
    let tabBar = this.renderer.createTabBar(this._document);

...
  }
``` 
IOW changing to this:
``` 
          createTabBar: (document?: Document | SahdowRoot) => this._createTabBar(this._document),
          createHandle: () => this._createHandle()
```
Ends up exactly the same?

You could argue that the following is better from a self-documenting point of view?
``` 
          createTabBar: (_document?: Document | SahdowRoot) => this._createTabBar(),
          createHandle: () => this._createHandle()
```

fcollonval: Thanks for the detailed answer. I agree with your latest proposal:

```
          createTabBar: (document?: Document | ShadowRoot) => this._createTabBar(),
          createHandle: () => this._createHandle()
```

It will indeed be a better self-documenting code. I opened https://github.com/jupyterlab/lumino/pull/285 to add this.

hbcarlos:
Thanks for the fix @GordonSmith!

PR #300, opened 2022-04-04
feature: Add commandWillExecute signal to CommandRegistry

thomasjm:
This signal would be useful to me.

Note: it looks like I need to update `review/api/commands.api.md` but I couldn't figure out how to do it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 <!-- <START NEW CHANGELOG ENTRY> -->
 
-- Add `commentWillExecute` signal to `CommandRegistry`.
-
 ## 2022.1.13
 
 ([Full Changelog](https://github.com/jupyterlab/lumino/compare/@lumino/application@1.28.0...f7d36a8bac6c44ee9928ea918d3473e297853658))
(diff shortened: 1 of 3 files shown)


blink1073 (commented):
The changestream entries are automated, mind removing this for now?

blink1073:
@thomasjm, thanks for this!  We're not maintaining the `review/api/` folder, since we're blocked by an upstream issue.

blink1073:
@fcollonval @jtpio thoughts on this one?

jasongrout:
Can we list some usecases where this signal would be important?

thomasjm:
For me, it would help to implement a command palette modal similar to the one in VSCode. I want to be able to know a command is about to be executed so I can take UI actions first (for example, closing the modal).

FWIW, another solution would be to add a signal or similar directly to the `CommandPalette` to announce when a command is about to be executed.

It's possible I can get this to work by just subclassing the `CommandRegistry`, let me try that first...

thomasjm:
BTW, I was able to do what I wanted by subclassing so you can close this if you like. Thanks!

blink1073:
Great, less API is better!

PR #301, opened 2022-04-28
Add handle moved signal to split panel

afshin:
The `handleMoved` signal is useful for capturing state changes in the layout of a `SplitPanel` and saving it (e.g. for layout restoration). This PR allows for removing the following code in JupyterLab:
```typescript
import { ISignal, Signal } from '@lumino/signaling';
import { SplitPanel as SPanel } from '@lumino/widgets';

/**
 * A deprecated split panel that will be removed when the phosphor split panel
 * supports a handle moved signal. See https://github.com/phosphorjs/phosphor/issues/297.
 */
export class SplitPanel extends SPanel {
  /**
   * Emits when the split handle has moved.
   */
  readonly handleMoved: ISignal<any, void> = new Signal<any, void>(this);

  handleEvent(event: Event): void {
    super.handleEvent(event);

    if (event.type === 'mouseup') {
      (this.handleMoved as Signal<any, void>).emit(undefined);
    }
  }
}
```
https://github.com/jupyterlab/jupyterlab/blob/dbdefed9db9332381fe4104bdf53ec314451951f/packages/settingeditor/src/splitpanel.ts

### DIFF
--- a/packages/widgets/src/splitpanel.ts
+++ b/packages/widgets/src/splitpanel.ts
@@ -351,7 +351,7 @@ export class SplitPanel extends Panel {
     this._pressData = null;
 
     // Emit the handle moved signal.
-    this._handleMoved.emit(undefined);
+    this._handleMoved.emit();
 
     // Remove the extra document listeners.
     document.removeEventListener('mouseup', this, true);
--- a/packages/widgets/src/splitpanel.ts
+++ b/packages/widgets/src/splitpanel.ts
@@ -119,7 +119,6 @@ export class SplitPanel extends Panel {
     return this._handleMoved;
   }
 
-
   /**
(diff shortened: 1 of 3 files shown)


packages/widgets/src/splitpanel.ts
```diff
@@ -341,6 +350,9 @@ export class SplitPanel extends Panel {
     this._pressData.override.dispose();
     this._pressData = null;
 
+    // Emit the handle moved signal.
+    this._handleMoved.emit(undefined);
```
fcollonval, line 354: You could drop the `undefined`, isn't it?

```suggestion
    this._handleMoved.emit();
```

fcollonval (approved):
Thanks @afshin

PR #304, opened 2022-05-30
Fix Accordion sizing issue

trungleduc:
## References
This PR fixes the sizing issue of the Accordion widget when a child is opened/closed multiple times.

## Code changes
- Add private method `_computeWidgetSize` to `AccordionPanel`  to cache and compute the size of widgets on title click event.
- Add boolean parameter `update`  to `SplitPanel.setRelativeSizes` to allow updating relative size without sending update message to the panel.

## User-facing changes
- Child widgets keep their sizes after being closed/opened.
- The resize behavior is changed, when a widget is opened or closed, the Accordion panel will try to resize only the last opened widget if possible.

### Old behavior

old

### New behavior

new

## Backwards-incompatible changes

N/A

### DIFF
--- a/packages/widgets/src/splitpanel.ts
+++ b/packages/widgets/src/splitpanel.ts
@@ -151,8 +151,8 @@ export class SplitPanel extends Panel {
    *
    * The actual geometry of the DOM nodes is updated asynchronously.
    */
-  setRelativeSizes(sizes: number[]): void {
-    (this.layout as SplitLayout).setRelativeSizes(sizes);
+  setRelativeSizes(sizes: number[], update = true): void {
+    (this.layout as SplitLayout).setRelativeSizes(sizes, update);
   }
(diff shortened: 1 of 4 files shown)


packages/widgets/src/splitpanel.ts
```diff
@@ -151,8 +151,8 @@ export class SplitPanel extends Panel {
    *
    * The actual geometry of the DOM nodes is updated asynchronously.
    */
-  setRelativeSizes(sizes: number[]): void {
-    (this.layout as SplitLayout).setRelativeSizes(sizes);
+  setRelativeSizes(sizes: number[], update = true): void {
```
fcollonval, line 154: Could you update the docstring here too?

fcollonval (approved):
Thanks a lot @trungleduc

PR #308, opened 2022-06-09
[memory-leak] Title.changed not cleared when owner is disposed

fcollonval:
### Reference

When working on adding memory leak test on JupyterLab in that [PR](https://github.com/jupyterlab/benchmarks/pull/100), it was highlighted that the signal connection to `Title.changed` where not cleared when the owner is disposed. This induces a memory leak due to connection stored in `Signal`.

---

In the open/close notebook scenario, the Launcher is removed from the memory leak by commenting the `title.changed` connection in [packages/running-extension/src/opentabs.ts](https://github.com/jupyterlab/jupyterlab/blob/master/packages/running-extension/src/opentabs.ts#L35).

### Code changes

- `Title` implements `IDisposable` and its the responsability of its owner to call `Title.dispose` (as it was added for `Widget` in this PR).

### DIFF
--- a/packages/widgets/tests/src/title.spec.ts
+++ b/packages/widgets/tests/src/title.spec.ts
@@ -11,7 +11,9 @@ import { expect } from 'chai';
 
 import { Title } from '@lumino/widgets';
 
-const owner = { name: 'Bob' };
+const owner = {
+  name: 'Bob'
+};
 
 describe('@lumino/widgets', () => {
   describe('Title', () => {
@@ -33,6 +35,19 @@ describe('@lumino/widgets', () => {
         });
         title.label = 'baz';
         expect(called).to.equal(true);
+      });
+
+      it('should clear signal if it is disposed', () => {
+        let called = false;
+        let title = new Title({ owner });
+        title.changed.connect((sender, arg) => {
+          called = true;
+        });
+
+        title.dispose();
+
+        title.label = 'baz';
+        expect(called).to.equal(false);
       });
     });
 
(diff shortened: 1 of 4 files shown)


fcollonval:
Another approach not changing the API would be to do the clean-up only with `Widget` for its `Title`. This will solve the memory leak seen in JupyterLab. But if the `Title` is used with other kind of owner by consumers, they may still be hit by this memory leak.

Another question: should we delete the `Title.changed` signal once the owner is disposed to ensure nobody will connect to it if the owner is disposed.

packages/widgets/tests/src/title.spec.ts
```diff
@@ -34,6 +36,19 @@ describe('@lumino/widgets', () => {
         title.label = 'baz';
         expect(called).to.equal(true);
       });
+
+      it('should clear signal if it is disposed', () => {
```
gabalafou, line 40: Minor comment.

I think it would help somebody reading this test code if the name of this test follows the pattern of the preceding test, since the code is patterned after that test:

```
it('should be emitted when the title state changes')
it('should not be emitted if it is disposed')
```

gabalafou (commented):
Looks good to me but I think somebody more familiar with Lumino should give it a pass

afshin:
Do you think the public setters (`label`, `mnemonic`, `icon`, `iconClass`, `iconLabel`) that emit a changed signal should test if `this.isDisposed` and `return` if necessary?

fcollonval:
> Do you think the public setters (`label`, `mnemonic`, `icon`, `iconClass`, `iconLabel`) that emit a changed signal should test if `this.isDisposed` and `return` if necessary?

@afshin I looked at other classes to see if we are testing `isDisposed` for setting attributes; we are never testing for it in setters. So I propose to not add those tests for `Title` neither.

fcollonval:
@afshin would you mind having a second look please?

afshin (approved):
Thank you, @fcollonval!

PR #314, opened 2022-07-06
Add the need to set _New Version Spec_ when using the releaser

fcollonval:


blink1073:
Thanks for the fixes and the release!

fcollonval:
You are welcome Steve - happy to share the maintenance burden.

PR #316, opened 2022-07-20
Support arguments to invoke a rate limited function with

afshin:


packages/polling/src/ratelimiter.ts
```diff
@@ -84,6 +87,11 @@ export abstract class RateLimiter<T, U> implements IRateLimiter<T, U> {
     return this.poll.stop();
   }
 
+  /**
+   * Arguments for the underlying function.
+   */
+  protected args: V | undefined = undefined;
```
trungleduc, line 93: Can we default `args` to `[]` so that the factory can call `fn(...this.args)` directly?

packages/polling/src/ratelimiter.ts
```diff
@@ -95,20 +103,33 @@ export abstract class RateLimiter<T, U> implements IRateLimiter<T, U> {
   protected poll: Poll<T, U, 'invoked'>;
 }
 
+export namespace RateLimiter {
+  export type Function<T, V extends any[]> =
+    | ((...args: V) => T | Promise<T>)
+    | (() => T | Promise<T>);
```
trungleduc, line 109: it does not seem necessary
```suggestion
```

packages/polling/src/ratelimiter.ts
afshin: This works even with it set to `undefined`.

packages/polling/src/ratelimiter.ts
afshin: You're right. Removed!

packages/polling/src/ratelimiter.ts
```diff
@@ -133,27 +159,33 @@ export class Throttler<T = any, U = any> extends RateLimiter<T, U> {
...
+  constructor(
+    fn: (...args: V) => T | Promise<T>,
+    options?: Throttler.IOptions | number
+  ) {
     super(fn, typeof options === 'number' ? options : options && options.limit);
-    let edge: 'leading' | 'trailing' = 'leading';
-    if (typeof options !== 'number') {
-      options = options || {};
-      edge = 'edge' in options ? options.edge! : edge;
+    if (typeof options !== 'number' && options && options.edge === 'trailing') {
```
fcollonval, line 167: Small compactor suggestion

```suggestion
    if (typeof options !== 'number' && options?.edge === 'trailing') {
```

fcollonval (approved):
Minor optional suggestion otherwise it looks good to me.

packages/polling/src/ratelimiter.ts
afshin: @fcollonval, I wish I could do that here. The version of TypeScript that Lumino uses doesn't have that feature. It hurts every time it comes up 🥲

afshin: Screen Shot 2022-07-21 at 11 25 32

afshin: Lumino uses TypeScript 3.6 and the chaining operator (`?.`) was introduced in 3.7. I don't know if we still need to be on this old version of the language though. Perhaps @blink1073 or @jasongrout remember the context and reason.

afshin: @fcollonval reminded me of this: https://github.com/jupyterlab/lumino/pull/175

Sorry for the noise, Jason and Steve!

PR #321, opened 2022-07-26
Add `collapse` and `expand` methods to `AccordionPanel`

fcollonval:
## Reference

Required for https://github.com/jupyterlab/jupyterlab/pull/12866

## Code Changes

Add API to programmatically expand or collapse a `AccordionPanel` widget. Before only user by clicking on the title could do that.

It is required for the extensions manager so we can show or hide the warning about installing third-party extensions depending on a settings parameter.

### DIFF
--- a/packages/widgets/src/accordionpanel.ts
+++ b/packages/widgets/src/accordionpanel.ts
@@ -69,12 +69,15 @@ export class AccordionPanel extends SplitPanel {
   /**
    * Collapse the widget at position `index`.
    *
+   * #### Notes
+   * If no widget is found for `index`, this will bail.
+   *
    * @param index Widget index
    */
   collapse(index: number): void {
     const widget = (this.layout as AccordionLayout).widgets[index];
 
-    if (!widget.isHidden) {
+    if (widget && !widget.isHidden) {
       this._toggleExpansion(index);
     }
   }
@@ -82,12 +85,15 @@ export class AccordionPanel extends SplitPanel {
   /**
    * Expand the widget at position `index`.
    *
+   * #### Notes
+   * If no widget is found for `index`, this will bail.
+   *
    * @param index Widget index
    */
   expand(index: number): void {
     const widget = (this.layout as AccordionLayout).widgets[index];
 
-    if (widget.isHidden) {
+    if (widget && widget.isHidden) {
       this._toggleExpansion(index);
     }
   }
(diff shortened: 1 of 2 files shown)


vidartf:
Would you mind writing a little bit more about why this is a good feature to have, and what it enables? It is hard to quickly parse that from the Lab PR.

Also, I see this was added to the Lumino 2 milestone, but I do not think this is backwards incompatible?

fcollonval:
> Would you mind writing a little bit more about why this is a good feature to have, and what it enables? It is hard to quickly parse that from the Lab PR (it has a diff of +2,238 −2,776 lines).

Done

> Also, I see this was added to the Lumino 2 milestone, but I do not think this is backwards incompatible?

Indeed this is compatible with the current version extending the API.

packages/widgets/src/accordionpanel.ts
```diff
@@ -66,6 +66,32 @@ export class AccordionPanel extends SplitPanel {
     widget.title.changed.connect(this._onTitleChanged, this);
   }
 
+  /**
+   * Collapse the widget at position `index`.
+   *
+   * @param index Widget index
+   */
+  collapse(index: number): void {
+    const widget = (this.layout as AccordionLayout).widgets[index];
```
afshin, line 75: We should handle when `widget` is `undefined` because `index` is wrong, I think.

packages/widgets/src/accordionpanel.ts
```diff
@@ -66,6 +66,32 @@ export class AccordionPanel extends SplitPanel {
...
+    }
+  }
+
+  /**
+   * Expand the widget at position `index`.
+   *
+   * @param index Widget index
+   */
+  expand(index: number): void {
+    const widget = (this.layout as AccordionLayout).widgets[index];
```
afshin, line 88: Same here

packages/widgets/src/accordionpanel.ts
```diff
@@ -296,6 +306,26 @@ export class AccordionPanel extends SplitPanel {
     }
   }
 
+  private _toggleExpansion(index: number) {
+    const title = this.titles[index];
+    const widget = (this.layout as AccordionLayout).widgets[index];
```
afshin, line 311: Presumably since you can control when this method is invoked it's not necessary to guard, here.

fcollonval: Agreed so I will not add guard here.

afshin (approved):
Thank you!

afshin:

Issue #58, opened 2020-03-07
Formally describing command args and results


bollwyvl:
Commands are critical to the interaction between many parts of a lumino app, yet are very loosely-typed, and generally opaque to implementers and users. This limits the discovery, effective user interface, and reusability of commands. While these problems are likely unsolveable at compile time, runtime validation and discovery would improve all of these for the user of a lumino application, with minimal effort to implementers.

### Discovery
At present, commands are difficult to discover. Yes, one can `listCommands` to get their ids, and then iterate over all of their option values, and try calling them.

When commands are further abstracted, it becomes basically impossible to determine whether a command execution will be well-formed, and what is just nonsense that will fail in a fully-spun-up application.

### User Interface
From the application perspective, commands are fire-and-forget, delegating the responsibility to individual implementations. Many commands would benefit from even minimal UI to modify their behavior, and could be, by default, presented in a single, consistent way vs some commands "just working", some opening dialogs, and other unpredictable/inconsistently accessible things that can occur.

## Payoff
With well-typed and documented commands, it becomes possible to generate machine-readable representations of possible command states (to-be-executed, was-executed, etc). A command palette, thus enhanced, could provide significant value.

Additionally, with discoverable commands, hybrid gui/cli applications, a la [clui](https://blog.repl.it/clui) become possible.

Finally, the ability to understand each command's arguments and result make it possible to serialize longer chains of commands that reference each other's output: `DockPanel`'s layout serialization represents this in one, particular way, but more generalized approaches could truly make lumino an engine for lisp machine/smalltalk-style self-extensibility and introspection.

## Out of scope

Selecting the validation description itself is probably neither viable nor useful: there still exists no standards-body adopted description-of-JSON-in-JSON, much less one implemented in browsers. While JSON Schema is the _de facto_ representation, with `ajv` the _de facto_ browser validator, it's worth exploring other options... once we have something that can accept any of it, with the usual caveats of performance, security, stability, and cross-language compatibility. 

- jddf: https://tools.ietf.org/html/draft-ucarion-jddf-05
- json-e: https://github.com/taskcluster/json-e

## Proposal

I don't _think_ any of these would require breaking changes, but it may be worth targeting/tabling this work until such a time as it would make sense. At any rate, I would be able to commit to implementing these feature, just not on what timeline.

- `@lumino/coreutils`
  - add `JSONValidationExt`
[...]

PR #322, opened 2022-07-28
Add `describedBy` to command

fcollonval:
## Reference

Partially implements #58

## Code changes

Add optional `describedBy` attributes to provide a JSON Schema for the commands arguments

**Question** Should we drop `usage` in favor of this computation description?


packages/commands/src/index.ts
```diff
@@ -151,6 +151,18 @@ export class CommandRegistry {
...
+  /**
+   * Get the description for a specific command.
+   *
+   * @param id - The id of the command of interest.
+   *
+   * @returns The description for the command.
+   */
+  describedBy(id: string): { args: ReadonlyJSONObject | null } {
+    let cmd = this._commands[id];
+    return cmd ? cmd.describedBy : { args: null };
```
afshin, line 163: Should we merge this after #319 so we can use the chaining operator?

fcollonval: Sure

fcollonval:
@afshin would you mind having a second look please?

afshin (approved):
Thank you!

cc: @bollwyvl

bollwyvl:
This is very rad, sorry I didn't see it earlier! This is very much a step up in the discoverability of commands.

One wrinkle to this: perhaps `describedBy`, much like the other pieces, could be allowed to be an `ReadonlyJSONObject | CommandFunc<ReadonlyJSONObject>` where the functional form accepted the current `args`... I can even imagine it being an async function, as fetching information might need to occur. 

Especially when driving UI, being able to add "late" content can be super helpful if there are certain `args` that would benefit from a dynamic `oneOf` here and there, but shouldn't block command registration. As a strawman, `kernel:start` could have an enumeration of all of the current kernel ids.

bollwyvl:
😁

fcollonval:

PR #327, opened 2022-08-04
Remove mouse event handlers from SplitPanel

afshin:
Continuation of Lumino 2 deprecated code removal: this PR removes mouse event handlers from `SplitPanel` and leaves only pointer event handlers.


blink1073 (approved):
LGTM!

fcollonval (approved):
Thanks @afshin

PR #331, opened 2022-08-04
Allow describedBy to be an async function

fcollonval:
Follow-up of https://github.com/jupyterlab/lumino/pull/322#issuecomment-1205383882

cc @bollwyvl

### DIFF
--- a/packages/commands/src/index.ts
+++ b/packages/commands/src/index.ts
@@ -156,11 +156,18 @@ export class CommandRegistry {
    *
    * @param id - The id of the command of interest.
    *
+   * @param args - The arguments for the command.
+   *
    * @returns The description for the command.
    */
-  describedBy(id: string): { args: ReadonlyJSONObject | null } {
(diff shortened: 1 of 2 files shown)


afshin:
Maybe instead of special-casing `describedBy` to be either a `Promise` or a `CommandFunc`, we should have all command functions potentially return a promise?


```typescript
export type CommandFunc<T> = (args: ReadonlyPartialJSONObject) => T | Promise<T>;
````

Or maybe that's a bad idea ... commands need to run synchronously, so maybe this special-casing is sound.

What do you think?

fcollonval:
Good question. I'm a bit reluctant to do that because most of them are for displaying information (in particular to build menus or command buttons). So I would rather not apply it as very slow asynchronous attributes may prevent displaying part of the UI or render in a weird way with popping e.g..

afshin (approved):

afshin:
Thank you!

PR #345, opened 2022-08-08
Remove `BPlusTree` class from `@lumino/collections`

afshin:
This PR removes the `BPlusTree` class from `@lumino/collections`. Its primary use was the `@lumino/datastore` package. It also does not have any unit tests, which we would have needed to add if we wanted to maintain it long-term.

If someone wants to use it, they can either access it from Lumino 1.x packages or they can copy/vendor the source code itself as it is contained within one file: https://github.com/jupyterlab/lumino/blob/1.x/packages/collections/src/bplustree.ts


fcollonval (approved):
LGTM

PR #348, opened 2022-08-13
Use Rollup for tests and examples

### LINKED ISSUE #332: Remove webpack from Lumino examples and code base
> As commented in https://github.com/jupyterlab/lumino/pull/330, it will be nice to drop webpack and to use rollup for the tests to remove any inconsistency between the packager used for publishing and for testing.
&mdash; @fcollonval https://github.com/jupyterlab/lumino/issues/325#issuecomment-1206518610

gabalafou:
Closes #332.

## Test plan

For each package, for example @lumino/algorithm:

```sh
cd packages/algorithm
yarn build:test
yarn test:firefox
yarn test:firefox-headless
yarn test:chrome
yarn test:chrome-headless
```

One way to check that each package (not examples, not tests) builds exactly the same as before is to build the packages from the main branch, then build them with this PR and compare the two. You can leverage git to do it:

```
git checkout -b build-from-main main
git clean -xfd
echo node_modules/ > .gitignore
yarn && yarn build && yarn build:examples && yarn build:test
git add .
git commit --no-verify -m"build"
git merge rollup
git checkout -b build-from-rollup rollup
git clean -xfd
echo node_modules/ > .gitignore
yarn && yarn build && yarn build:examples && yarn build:test
git add .
git commit --no-verify -m"build"
git diff build-from-main --stat
```

You'll see that the bundles for the examples and the tests have changed. That's to be expected because they were previously built with Webpack. But you shouldn't see any differences for the main packages, which were built with Rollup before this PR (and are still built with Rollup after this PR).


gabalafou:
I'm opening this draft pull request with one package in order to catch any problems before extending to all of the other packages.

I will soon also convert a single example from the examples folder to use rollup instead of webpacke.

packages/algorithm/package.json
```diff
@@ -26,9 +26,9 @@
     "api": "api-extractor run --local --verbose",
     "build": "npm run build:src && rollup -c",
     "build:src": "tsc --build",
-    "build:test": "tsc --build tests && cd tests && webpack",
+    "build:test": "npm run clean:test && tsc --build tests && cd tests && rollup -c",
```
gabalafou, line 29: Should clean be part of the build step? If you don't run clean and delete the `tests/tsconfig.tsbuildinfo` file then the build step will do nothing.

packages/algorithm/package.json
```diff
@@ -26,9 +26,9 @@
...
     "clean": "rimraf ./lib && rimraf *.tsbuildinfo && rimraf ./types && rimraf ./dist",
-    "clean:test": "rimraf tests/build",
+    "clean:test": "rimraf tests/lib && rimraf tests/tsconfig.tsbuildinfo",
```
gabalafou, line 31: Must delete the `.tsbuildinfo` file or else the build command will do nothing.

packages/algorithm/package.json
```diff
@@ -59,9 +60,7 @@
     "terser": "^5.7.1",
     "tslib": "^2.4.0",
     "typedoc": "~0.23.9",
-    "typescript": "~4.7.3",
-    "webpack": "^5.74.0",
-    "webpack-cli": "^4.10.0"
```
gabalafou, line 64: I'm guessing these are the only two webpack-related dependencies to delete.

packages/algorithm/tests/rollup.config.js
```diff
@@ -0,0 +1,10 @@
+import nodeResolve from '@rollup/plugin-node-resolve';
+import commonjs from '@rollup/plugin-commonjs';
+
+export default {
+  input: './lib/index.spec.js',
+  output: {
+    file: './lib/bundle.test.js'
+  },
+  plugins: [commonjs(), nodeResolve()]
```
gabalafou, line 9: Without `nodeResolve` plugin, the test files (*.spec.ts) cannot import from `chai` or `@lumino/algorithm`.

Without `commonjs` plugin, the `chai` import doesn't work. You get the following error:

```
Error: 'default' is not exported by ../../../node_modules/chai/index.js, imported by ../../../node_modules/chai/index.mjs
```

Note that with these two plugins, you will get a warning about circular dependencies in Chai, but it's just a warning, not an error, and the tests will build, run, and pass.

Does the order of these plugins matter?

packages/algorithm/package.json
```diff
@@ -26,9 +26,9 @@
...
     "build": "npm run build:src && rollup -c",
     "build:src": "tsc --build",
-    "build:test": "tsc --build tests && cd tests && webpack",
+    "build:test": "npm run clean:test && tsc --build tests && cd tests && rollup -c",
     "clean": "rimraf ./lib && rimraf *.tsbuildinfo && rimraf ./types && rimraf ./dist",
-    "clean:test": "rimraf tests/build",
+    "clean:test": "rimraf tests/lib && rimraf tests/tsconfig.tsbuildinfo",
     "docs": "typedoc --options ../../typedoc.js src",
     "minimize": "terser dist/index.js -c -m --source-map \"content='dist/index.js.map',url='index.min.js.map'\" -o dist/index.min.js",
     "test": "npm run test:firefox-headless",
```
gabalafou, line 34: Should we remove `test:ie`?

gabalafou (commented):
done self-reviewing

examples/example-accordionpanel/style/index.css
```diff
@@ -1,8 +1,8 @@
-/* 
+/*
  Copyright (c) Jupyter Development Team.
  Distributed under the terms of the Modified BSD License.
 */
-@import '~@lumino/default-theme/style/index.css';
+@import '@lumino/default-theme/style/index.css';
```
gabalafou, line 5: I don't really know what I'm doing when it comes to CSS loaders, but after just trying a bunch of different [CSS Rollup plugins](https://github.com/rollup/awesome#css) and various code changes I found that the following worked: 

1. Use `rollup-plugin-styles` 
2. Change any CSS `@import` in the chain to use `@` rather than `~@` syntax.

yarn.lock
```diff
@@ -1188,6 +1188,14 @@
     estree-walker "^1.0.1"
```
gabalafou, line 1188: All `yarn.lock` changes should be related to the changes in the package.json files in this PR.

gabalafou (commented):
Done self-reviewing second commit (`example-accordionpanel`)

packages/algorithm/package.json
blink1073: Yeah, we don't support `ie` anymore :smile:.  Ideally we'd run with [karma-webkit-launcher](https://github.com/google/karma-webkit-launcher) since we claim to support webkit, chrome, and firefox in JupyterLab.

examples/example-accordionpanel/style/index.css
blink1073: Makes sense to me

packages/algorithm/package.json
blink1073: Yeah I agree

packages/algorithm/tests/rollup.config.js
blink1073: I wouldn't worry the plugin order for tests

packages/algorithm/package.json
gabalafou, line 29: I wasn't sure if I should add `clean:test` here, but without it the `build:test` script/command is not idempotent because it creates a `tsconfig.tsbuildinfo` file, which I found prevents a second run of the `build:test` command from picking up intervening changes in the test code.

gabalafou:
Thanks @blink1073 for reviewing the test-related Rollup changes. I feel confident now extending my changes in @lumino/algorithm tests to the other packages' tests in the monorepo.

packages/algorithm/package.json
fcollonval, line 31: FYI you can pass multiple targets to `rimraf`:

```suggestion
    "clean:test": "rimraf tests/lib tests/tsconfig.tsbuildinfo",
```

examples/example-accordionpanel/rollup.config.js
```diff
@@ -0,0 +1,12 @@
+import nodeResolve from '@rollup/plugin-node-resolve';
```
fcollonval, line 1: Would you mind refactoring all rollup config (including the one currently set in the packages) to target common files at the root level as it is done for typescript and api-extractor configuration. This will ease maintenance.

yarn.lock
```diff
@@ -7400,6 +7509,13 @@ rollup@^2.77.2:
   optionalDependencies:
     fsevents "~2.3.2"
 
+rollup@^2.77.3:
```
fcollonval, line 7512: Before merging, we should check that this entry disappears as it should merge with the previous one. But it is not a blocker.

fcollonval (commented):
Thanks a lot @gabalafou 

I added some suggestions before you move forward.

gabalafou:
@fcollonval, could you take a look at my [last commit add2f47](https://github.com/jupyterlab/lumino/pull/348/commits/add2f4711f7697fa7d3b80d734c09bd7c573990e), is it something like what you had imagined?

One question I have: should I move all of the rollup dependencies out of the various `package.json`s and into the root package.json?

rollup.examples.config.js
```diff
@@ -0,0 +1,16 @@
+import nodeResolve from '@rollup/plugin-node-resolve';
+import styles from 'rollup-plugin-styles';
+
+const globals = id =>
```
gabalafou, line 4: I copied this line from rollup.src.config because it fixed a warning or error message that I saw from the `rollup` command.

rollup.src.config.js
```diff
@@ -0,0 +1,38 @@
+import nodeResolve from '@rollup/plugin-node-resolve';
+import sourcemaps from 'rollup-plugin-sourcemaps';
+import postcss from 'rollup-plugin-postcss';
```
gabalafou, line 3: These dependencies are defined in e.g. `packages/algorithm/package.json`... but should they be under `devDependencies` in the repo's root package.json (or somewhere else)?

packages/algorithm/rollup.config.js
```diff
@@ -1,40 +1,6 @@
-import nodeResolve from '@rollup/plugin-node-resolve';
-import sourcemaps from 'rollup-plugin-sourcemaps';
-import postcss from 'rollup-plugin-postcss';
+import { createRollupConfig } from '../../rollup.src.config';
```
gabalafou, line 1: For consistency, perhaps the other two base Rollup config files should also output a config-generating function versus the config object itself?

packages/algorithm/rollup.config.js
```diff
@@ -1,40 +1,6 @@
...
-      }),
-      sourcemaps(),
-      postcss({
-        extensions: ['.css'],
-        minimize: true
-      })
-    ]
-  }
-];
+export default config;
```
gabalafou, line 6: By the way, I cross-checked this by comparing the results of from building with the old file versus building with the new file, as you can see from my console log below:

```
% cd packages/algorithm
% diff dist/index.es6.js prev_dist/index.es6.js                                                                                                                                                                                                                        rollup
% diff dist/index.es6.js.map prev_dist/index.es6.js.map                                                                                                                                                                                                                rollup
% diff dist/index.js prev_dist/index.js                                                                                                                                                                                                                                rollup
% diff dist/index.js.map prev_dist/index.js.map     
```

rollup.src.config.js
fcollonval: let's keep them in each package as it is where they are used.

fcollonval:
> should I move all of the rollup dependencies (the plugins) out of the various `package.json`s and into the root package.json?

@gabalafou it is better to let them in each package where they are used.

yarn.lock
gabalafou: Nope. Still hasn't disappeared.

gabalafou: When I added rollup to the package.json file for each example, it set: `"rollup": "^2.77.3"`.

rollup.examples.config.js
```diff
@@ -0,0 +1,18 @@
+import nodeResolve from '@rollup/plugin-node-resolve';
+import styles from 'rollup-plugin-styles';
+
+export const globals = id =>
```
gabalafou, line 4: There's something I find a bit counter-intuitive about these base config files. For example, I tripped up here because I thought I could replace this line with an import like so:

```js
import { globals } from './rollup.src.config';
```

But that caused the build script in the examples to fail. That's because importing from that other Rollup config brought along its dependencies, which are not included in the package.json for each of the example packages. In other words, for example, `rollup-plugin-postcss` is not included as a dependency for any of the example-* packages, but is a dependency in the other packages.

yarn.lock
fcollonval: Would you mind running `yarn why rollup` to find why we have two versions?

rollup.examples.config.js
fcollonval: This is expected - a solution would be to define a base config that only uses common dep. But here this is probably not worth the trouble just for the `globals` definition.

fcollonval (commented):
Thanks @gabalafou 

We only have to figure out why there are still two versions of rollup in the yarn.lock. And we should be good to go.

yarn.lock
gabalafou: Here is my console log for `yarn why rollup`:

```
yarn why rollup                                                                                                                                                                                                                                                                              rollup
yarn why v1.22.19
[1/4] 🤔  Why do we have the module "rollup"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "rollup@2.77.2"
info Has been hoisted to "rollup"
info Reasons this module exists
   - "workspace-aggregator-511ef2db-8ac6-4e9e-8474-e5094676397f" depends on it
   - Hoisted from "_project_#@lumino#algorithm#rollup"
   - Hoisted from "_project_#@lumino#application#rollup"
   - Hoisted from "_project_#@lumino#collections#rollup"
   - Hoisted from "_project_#@lumino#commands#rollup"
   - Hoisted from "_project_#@lumino#coreutils#rollup"
   - Hoisted from "_project_#@lumino#datagrid#rollup"
   - Hoisted from "_project_#@lumino#disposable#rollup"
   - Hoisted from "_project_#@lumino#domutils#rollup"
   - Hoisted from "_project_#@lumino#dragdrop#rollup"
   - Hoisted from "_project_#@lumino#keyboard#rollup"
   - Hoisted from "_project_#@lumino#messaging#rollup"
   - Hoisted from "_project_#@lumino#polling#rollup"
   - Hoisted from "_project_#@lumino#properties#rollup"
   - Hoisted from "_project_#@lumino#signaling#rollup"
   - Hoisted from "_project_#@lumino#virtualdom#rollup"
   - Hoisted from "_project_#@lumino#widgets#rollup"
info Disk size without dependencies: "6.43MB"
info Disk size with unique dependencies: "6.43MB"
info Disk size with transitive dependencies: "6.43MB"
info Number of shared dependencies: 0
=> Found "@lumino/example-accordionpanel#rollup@2.77.3"
info This module exists because "_project_#@lumino#example-accordionpanel" depends on it.
info Disk size without dependencies: "6.43MB"
info Disk size with unique dependencies: "6.43MB"
info Disk size with transitive dependencies: "6.43MB"
info Number of shared dependencies: 0
=> Found "@lumino/example-datagrid#rollup@2.77.3"
info This module exists because "_project_#@lumino#example-datagrid" depends on it.
info Disk size without dependencies: "6.43MB"
info Disk size with unique dependencies: "6.43MB"
info Disk size with transitive dependencies: "6.43MB"
info Number of shared dependencies: 0
=> Found "@lumino/example-dockpanel#rollup@2.77.3"
info This module exists because "_project_#@lumino#example-dockpanel" depends on it.
info Disk size without dependencies: "6.43MB"
info Disk size with unique dependencies: "6.43MB"
info Disk size with transitive dependencies: "6.43MB"
info Number of shared dependencies: 0
✨  Done in 0.85s.
```

fcollonval:
Kicking the CI

fcollonval (approved):
thanks a lot @gabalafou

PR #354, opened 2022-08-18
Update and fix docs

gabalafou:
This PR contains changes to improve and fix the docs.

You can see the results of these changes by going to: https://gabalafoulumino.readthedocs.io.

You should also be able to build and try the docs locally:

```sh
conda env create -f docs/environment.yml -y
conda activate lumino_documentation
cd docs
make html
open build/html/index.html
```


packages/algorithm/package.json
```diff
@@ -29,7 +29,6 @@
     "build:test": "tsc --build tests && cd tests && webpack",
     "clean": "rimraf ./lib && rimraf *.tsbuildinfo && rimraf ./types && rimraf ./dist",
     "clean:test": "rimraf tests/build",
-    "docs": "typedoc --options ../../typedoc.js src",
```
gabalafou, line 32: I decided to get rid of the `docs` script in each individual package because it doesn't really make sense to me to build the docs for a single package in isolation of the others, especially since some of them depend on others in the monorepo.

Please let me know if my reasoning here is wrong.

typedoc.js
```diff
@@ -1,29 +1,9 @@
...
-  entryPoints,
-  exclude,
+  entryPoints: '.',
+  entryPointStrategy: 'packages',
+  exclude: ['examples/**', '**/tests/**', 'packages/default-theme'],
+  includeVersion: true,
   name: '@lumino',
   out: 'docs/source/api',
-  readme: 'README.md',
-  tsconfig: 'tsconfigdoc.json'
```
fcollonval, line 28: I guess this means we can remove `tsconfigdoc.json`

packages/algorithm/package.json
```diff
@@ -39,6 +38,10 @@
     "test:ie": "cd tests && karma start --browsers=IE",
     "watch": "tsc --build --watch"
   },
+  "typedoc": {
+    "entryPoint": "./src/index.ts",
+    "displayName": "algorithm"
```
gabalafou, line 43: Without setting display name like this, the breadcrumbs in the rendered docs end up looking like: `@lumino / @lumino/algorithm / member`. You'll have to check the rendered docs, and see for yourself if you like it prefer it this way or not.

typedoc.js
gabalafou, line 28: I couldn't find any other reference in the codebase to `tsconfigdoc.json`. Should I delete that file?

fcollonval (changes requested):
Thanks @gabalafou 

This looks great on https://lumino--354.org.readthedocs.build/en/354/

Just one comment (see code) and a suggestion: it will be nice to switch to pydata-sphinx-theme to align with the other Jupyter projects (see example in [JupyterLab migration](https://github.com/jupyterlab/jupyterlab/pull/11803)). But that could be a follow-up.

typedoc.js
```diff
@@ -1,29 +1,9 @@
...
+  readme: 'none'
```
gabalafou, line 8: Setting `readme` to `none` means that the index page shows the list of packages rather than the repo readme, which allows us to get rid of the index-copy step in the conf.py file.

docs/source/conf.py
```diff
@@ -104,56 +104,43 @@ def build_api_docs(out_dir):
...
+    yarn = [shutil.which('yarn')]
+    check_call(yarn, cwd=root)
+    check_call(yarn + ["docs"], cwd=root)
 
     dest_dir = osp.join(out_dir, "api")
     print(f"Copying {docs_api} -> {dest_dir}")
     if osp.exists(dest_dir):
         shutil.rmtree(dest_dir)
     shutil.copytree(docs_api, dest_dir)
-    shutil.copy(osp.join(HERE, 'api_index.html'), osp.join(dest_dir, 'index.html'))
```
gabalafou, line 126: If we set `readme` to `none` in the Typedoc config file then we don't need to do this copy step.

gabalafou (commented):
done self-reviewing, left comments for reviewer

typedoc.js
gabalafou: Oh you made this comment while I was self-reviewing this PR. I removed the file in a subsequent commit.

typedoc.js
fcollonval: I found the package first page for this PR: https://lumino--354.org.readthedocs.build/en/354/api/modules/algorithm.html
much better than the current one: https://lumino.readthedocs.io/en/stable/api/algorithm/index.html

So definitely :100: for this new setup

Issue #353, opened 2022-08-17
Update `dragdrop` package to use `PointerEvent` and sub-class `DragEvent`


afshin:
[initMouseEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/initMouseEvent) is deprecated. We should update the `@lumino/dragdrop` package to use [`new PointerEvent(...)`](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent) instances or sub-classes of `PointerEvent`.

afshin:
[initMouseEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/initMouseEvent) is deprecated. We should update the `@lumino/dragdrop` package to use [`new PointerEvent(...)`](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent) instances or sub-classes of `PointerEvent`.

Issue #340, opened 2022-08-08
Author Lumino 1 → Lumino 2 migration guide


afshin:
While we produce change logs automatically, we'll need something more substantial with examples and some narrative to help users migrate their code with the Lumino update. This document will complement the JupyterLab 3 → JupyterLab 4 migration guide for extension authors.

afshin:
I think handling this by looking at open and closed issues in the [Lumino 2 milestone](https://github.com/jupyterlab/lumino/milestone/1) might be good because you'll see everything that's being modified or removed, there.

afshin:
The top-level comment for this PR https://github.com/jupyterlab/lumino/pull/346 lists out all the API changes that PR contains and is relevant here.

afshin:
There is now a `migration.md` file in the docs that includes the documentation for iterators.

PR #355, opened 2022-08-18
Update `dragdrop` to use `PointerEvent` and `DragEvent`

afshin:
- This PR updates `@lumino/dragdrop` to use [`new PointerEvent(...)`](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent) instances or sub-classes of `DragEvent` instead of [initMouseEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/initMouseEvent), which is deprecated.
- This PR **moves** the `DropAction` and `SupportedActions` type definitions into the `Drag` namespace. This does not seem to affect any downstream Lumino or JupyterLab code.
- This PR **deprecates** the `IDragEvent` interface and suggests client code use `Drag.Event` instead.

Fixes https://github.com/jupyterlab/lumino/issues/353
Partially fixes https://github.com/jupyterlab/lumino/issues/340

### `@lumino/dragdrop`
| | `export` | name | note |
|-|-|-|-|
| ❌ | `type` | `DropAction` | Moved to `Drag.DropAction` |
| ❌ | `type` | `SupportedActions` | Moved to `Drag.SupportedActions` |
| ☑️ | `interface` | `IDragEvent` | `@deprecated`, use `Drag.Event` instead |

### DIFF
--- a/review/api/dragdrop.api.md
+++ b/review/api/dragdrop.api.md
@@ -17,14 +17,30 @@ export class Drag implements IDisposable {
     get isDisposed(): boolean;
     readonly mimeData: MimeData;
     protected moveDragImage(clientX: number, clientY: number): void;
-    readonly proposedAction: DropAction;
+    readonly proposedAction: Drag.DropAction;
     readonly source: any;
-    start(clientX: number, clientY: number): Promise<DropAction>;
-    readonly supportedActions: SupportedActions;
+    start(clientX: number, clientY: number): Promise<Drag.DropAction>;
+    readonly supportedActions: Drag.SupportedActions;
 }
 
 // @public
 export namespace Drag {
+    export type DropAction = 'none' | 'copy' | 'link' | 'move';
+    export class Event extends DragEvent {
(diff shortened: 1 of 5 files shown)


review/api/dragdrop.api.md
```diff
@@ -34,23 +50,13 @@ export namespace Drag {
...
-
-// @public
-export interface IDragEvent extends MouseEvent {
-    dropAction: DropAction;
-    readonly mimeData: MimeData;
-    readonly proposedAction: DropAction;
-    readonly source: any;
-    readonly supportedActions: SupportedActions;
+// @public @deprecated (undocumented)
+export interface IDragEvent extends Drag.Event {
```
fcollonval, line 57: As this is deprecating `IDragEvent`, I think we should do the same for `DropAction` and `SupportedActions` as they are part of `IDragEvent` interface.

Otherwise this looks good.

fcollonval (approved):
LGTM

PR #373, opened 2022-08-23
Fix tab trap in menubar

gabalafou:
There are a few tab traps in the JupyterLab UI. The worst offender is the top menu bar. The code change in this pull request is the most straightforward way to fix the menu bar tab trap in JupyterLab.

In terms of backwards compatibility, I have a hard time imagining that downstream dependents on Lumino would rely on the tab key event getting caught and swallowed (`event.preventDefault` + `event.stopPropagation`) by the Lumino menu bar, but it is something to consider.

### DIFF
--- a/packages/widgets/src/menubar.ts
+++ b/packages/widgets/src/menubar.ts
@@ -409,6 +409,9 @@ export class MenuBar extends Widget {
 
   /**
    * Handle the `'keydown'` event for the menu bar.
+   *
+   * #### Notes
+   * All keys are trapped except the tab key that is ignored.
    */
   private _evtKeyDown(event: KeyboardEvent): void {
     // Fetch the key code for the event.
@@ -419,7 +422,7 @@ export class MenuBar extends Widget {
       return;
     }
 
-    // A menu bar handles all keydown events.
+    // A menu bar handles all other keydown events.
     event.preventDefault();
     event.stopPropagation();
 
--- a/packages/widgets/src/menubar.ts
+++ b/packages/widgets/src/menubar.ts
@@ -411,12 +411,17 @@ export class MenuBar extends Widget {
    * Handle the `'keydown'` event for the menu bar.
    */
   private _evtKeyDown(event: KeyboardEvent): void {
+    // Fetch the key code for the event.
+    let kc = event.keyCode;
+
+    // Do not trap the tab key.
+    if (kc === 9) {
+      return;
+    }
+
     // A menu bar handles all keydown events.
     event.preventDefault();
     event.stopPropagation();
-
-    // Fetch the key code for the event.
-    let kc = event.keyCode;
 
     // Enter, Up Arrow, Down Arrow
     if (kc === 13 || kc === 38 || kc === 40) {
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Lumino
 
 [![Build Status](https://github.com/jupyterlab/lumino/workflows/Tests/badge.svg?branch=main)](https://github.com/jupyterlab/lumino/actions?query=branch%3Amain+workflow%3A%22Tests%22)
-[![Documentation Status](https://readthedocs.org/projects/jupyterlab/badge/?version=stable)](http://lumino.readthedocs.io/en/stable/)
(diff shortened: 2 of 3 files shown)


gabalafou:
Note that this PR does not fix all of the tab traps in JupyterLab. There are others, these two in particular:

- when a notebook is open, code cells become tab traps (technically, you can `ESC`-key out of the code cell and tab backwards through the UI, but I couldn't figure out a way to go forward past the code cell)
- opening the command-line terminal also creates a tab trap unless you know to press `CTRL` + `D`

gabalafou:
## Testing

Testing this change against the JupyterLab UI is a little bit tricky right now.

When I tried to link `@lumino/widgets`, I got errors when trying to do a dev build of the JupyterLab frontend code. I think that's because the JupyterLab frontend code has not yet been updated to be compatible with the latest Lumino code.

So I had to cherry pick the commit from this PR onto a branch based on the Lumino 1.x branch, and then do the linking.

In terms of the command line, it looks like the following.

First, with the Lumino repo as your current working directory:
```
git checkout -b no-tab-traps 1.x
git cherry-pick 911f4b067f27d88be3d9707e33ad4eac746009b7
yarn 
yarn build
cd packages/widgets
yarn link
```

Then, with the JupyterLab repo as current working directory (and assuming you have already pulled latest changes and installed everything):
```
jlpm link @lumino/widgets
jlpm build
jupyter lab --dev-mode
```

packages/widgets/src/menubar.ts
```diff
@@ -411,13 +411,18 @@ export class MenuBar extends Widget {
    * Handle the `'keydown'` event for the menu bar.
```
fcollonval, line 411:
```suggestion
   * Handle the `'keydown'` event for the menu bar.
   *
   * #### Notes
   * All keys are trapped except the tab key that is ignored.
```

packages/widgets/src/menubar.ts
```diff
@@ -411,13 +411,18 @@ export class MenuBar extends Widget {
...
   private _evtKeyDown(event: KeyboardEvent): void {
+    // Fetch the key code for the event.
+    let kc = event.keyCode;
+
+    // Do not trap the tab key.
+    if (kc === 9) {
+      return;
+    }
+
     // A menu bar handles all keydown events.
```
fcollonval, line 422:
```suggestion
    // A menu bar handles all other keydown events.
```

fcollonval (changes requested):
Just adding more information in the code comments. Otherwise it looks good to me.

fcollonval (approved):
Thanks @gabalafou

packages/widgets/src/menubar.ts
```diff
@@ -409,15 +409,23 @@ export class MenuBar extends Widget {
...
-
     // Fetch the key code for the event.
     let kc = event.keyCode;
 
+    // Do not trap the tab key.
+    if (kc === 9) {
+      return;
+    }
+
+    // A menu bar handles all other keydown events.
```
gabalafou, line 425: I wonder if the logic here is sound.

Originally this handler only called `preventDefault` and `stopPropagation` if it actually handled the key. But then the handler was rewritten to [handle mnemonics](https://github.com/jupyterlab/lumino/commit/6b506e2010248b9c1a348c73e374787675672326). I lightly suspect that the decision to call prevent and stop on all keys was done to make the handler more readable, without too much thought given to possible negative consequences.

I wrote this PR conservatively, meaning I kept everything else about this function the same EXCEPT when the tab key is pressed. But I guess it's worth asking whether the handler should be refactored to only intercept key events that it actually does something with.

I could imagine a handler that looks more like the following:

```js
keyDown() {
  let kc = event.keyCode;
  let endEvent = () => { 
    event.preventDefault();
    event.stopPropagation();
  };
  // Enter, Up Arrow, Down Arrow
  if (kc === 13 || kc === 38 || kc === 40) {
    endEvent();
    this.openActiveMenu();
    return;
  }
  // ...
}
```

In other words, it would go back to looking like the original implementation.

afshin (approved):
Thank you!

packages/widgets/src/menubar.ts
afshin: Let's pursue this train of thought in another PR if necessary so we can merge this one and iterate.

fcollonval:

PR #377, opened 2022-08-25
Dynamic extensions reloading

### LINKED ISSUE #278: Dynamic extensions reloading: `IPlugin.deactivate`, `Application.registerPlugin` and `Application.deactivatePlugin` proposal
### Problem

The dynamic extensions provided by [`jupyterlab-plugin-playground`](https://github.com/jupyterlab/jupyterlab-plugin-playground) cannot be registered more than once with the same id. This forces the user to reload the page every time, or at least change the id each time they activate the plugin, which is really annoying for the prototyping workflows as intended for the plugin playground.

Neither normal JupyterLab/Lumino extensions can be unloaded on runtime. It might be by design - maybe there was an assumption that reloading websites is cheap - but the cost is suboptimal for some more recent use cases like:
- JupyterLab Desktop where refresh = close and reopen app
- JupyterLite = the bundle weights much more and kernels will take long to load

To allow to reload extension with new updated code we need to unload it first. We could remove the plugin manually from the plugin map (inelegantly accessing private properties of `Application`, which is what is currently done in https://github.com/jupyterlab/jupyterlab-plugin-playground/pull/28) but when we register module anew, the old one remains active, with all of its commands, widgets, toolbars and menus attached.

### Proposed Solution

There should be three additions to the API are needed:
- `Application.unregisterPlugin()` method to remove the plugin from the plugin map and complement the existing `registerPlugin()`
- an optional `deactivate` method which plugins may decide to implement. If it is provided, it should undo all the changes the `activate` method made, including removing the added commands, menus, etc. to complement the current [`activate()`](https://github.com/jupyterlab/lumino/blob/578509cae6af5567975a6ed2e6bc9d8a1927a078/packages/application/src/index.ts#L84-L101) method.
- `Application.deactivatePlugin` method to call the `deactivate` method on the plugin and all of it dependants

The `unregisterPlugin` would roughly be:

```typescript
/**
 * @param id - id of the plugin to unregister
 * @param force - whether to unregister even if active
 */
async unregisterPlugin(id: str, force?: boolean) {
  const data = this._pluginMap[id];
  if (data.activated && !force) {
     reject('plugin is still active');
  }
  delete this._pluginMap[plugin.id];
}
```

`deactivatePlugin` would reject if plugin does not have `deactivate()` method:

```typescript
/**
 * Will deactivate the plugin and its dependant if and only if the plugin
 * and its dependants all support `deactivate`.
 * @returns other dependant plugins deactivated as a consequence
 */
deactivatePlugin(id: string): IPlugin[] {
  if (!plugin.deactivate) {
     reject('This plugin does not support deactivation');
  }
  // find an optimal deactivation order for dependants, probably some DAG flattened to an array
  const pluginDependantsInDAGOrder = this._findDependants();
  for (const dependant of pluginDependantsInDAGOrder) {
     if (!dependant.deactivate) {
         reject('Plugin dependant does not support deactivation: ' + dependant.id)
     }
  }
  for (const dependant of pluginDependantsInDAGOrder) {
     await this.deactivatePlugin(dependant.id);
  }
  await plugin.deactivate()
  return pluginDependantsDAG;
}
```

It will be up to plugin authors to adopt the `deactivate` method. This would be a backwards-compatible change.

The signature of `deactivate` could be:

```typescript
deactivate?: (app: T, ...args: any[]) => void | Promise<void>; 
```

Because plugins are also service providers we would want to unregister the associated service _after_ the `deactivate()` call completed (allowing the plugin to cleanly shut the service down to avoid memory leaks), thus we want to allow the plugins to return a promise.

If `deactivate` is not present the service provided by the plugin should never be removed.

Questions:
- [ ] I think that the return type should be `void` because we would know what the associated service is; any other takes?
- [ ] what should happen if `deactivate()` is present but throws or rejects? Should we still unregister the service?

### Additional context

- VSCode Extensions API have both [`activate` and `deactivate`](https://code.visualstudio.com/api/get-started/extension-anatomy#extension-entry-file) functions as entry points. The `deactivate()` method is optional in VSCode.
- InteliJ Platform Plugin SDK has [plugin `load`/`unload` events](https://plugins.jetbrains.com/docs/intellij/dynamic-plugins.html#plugin-loadunload-events) for its [dynamic extensions](https://plugins.jetbrains.com/docs/intellij/dynamic-plugins.html#restrictions); if I read the documentation correctly it also disallows use of any entry points which cannot by loaded/unloaded dynamically for dynamic plugins

Adding a `deactivate` which enables hot reloading of dynamic extensions is a pre-requisite for using plugin-playground (also cf https://github.com/jupyterlab/jupyterlab/issues/8866) to develop extensions inside JupyterLab itself (https://github.com/jupyterlab/jupyterlab/issues/7469).

Finally, deregistration wrapper will also need to be added downstream in JupyterLab where `registerPluginModule` is implemented.

fcollonval:
Fixes #278

## Code changes

There are three additions to the API :
- `Application.unregisterPlugin()` method to remove the plugin from the plugin map and complement the existing `registerPlugin()`
- an optional `deactivate` method which plugins may decide to implement. If it is provided, it should undo all the changes the `activate` method made, including removing the added commands, menus, etc. to complement the current [`activate()`](https://github.com/jupyterlab/lumino/blob/578509cae6af5567975a6ed2e6bc9d8a1927a078/packages/application/src/index.ts#L84-L101) method.
- `Application.deactivatePlugin` method to call the `deactivate` method on the plugin and all of it dependants

The `unregisterPlugin` signature is:

```typescript
/**
 * @param id - id of the plugin to unregister
 * @param force - whether to unregister even if active
 */
async unregisterPlugin(id: str, force?: boolean)
```

`deactivatePlugin` reject if plugin or one of its dependents does not have `deactivate()` method:

```typescript
/**
 * Will deactivate the plugin and its dependant if and only if the plugin
 * and its dependants all support `deactivate`.
 * @returns other dependant plugins deactivated as a consequence
 */
deactivatePlugin(id: string): Promise<string[]>
```

> The return object is a list of plugin ids rather than a IPlugin list to avoid security hole that returning IPlugin will open (the activate function could be altered by deactivating and reactivating a Plugin with the proposed API in #278)

It is up to plugin authors to adopt the `deactivate` method. This is a backwards-compatible change.

The signature of `deactivate` is:

```typescript
deactivate?: (app: T, ...args: any[]) => void | Promise<void>; 
```

:warning: deactivating a plugin will deactivate all its dependents optional or not. In theory, we could reactivate the one having a optional relationship. I propose to keep it like this and to let sub class application decide what to do - the rational is that we anyway have to modify some API in JupyterLab itself to support dynamic reload.

---

Side effect I replace the use of objects by mapping because deleting a object key in JS is famous to reduce performance.

### DIFF
--- a/packages/application/src/index.ts
+++ b/packages/application/src/index.ts
@@ -373,10 +373,16 @@ export class Application<T extends Widget> {
           }
         })
       );
-      dependent.service = null;
-      dependent.activated = false;
       if (promise) {
-        promises.push(promise);
+        promises.push(
+          promise.then(() => {
+            dependent.service = null;
+            dependent.activated = false;
+          })
+        );
+      } else {
+        dependent.service = null;
+        dependent.activated = false;
       }
     }
     await Promise.all(promises);
@@ -835,7 +841,7 @@ namespace Private {
    * @returns A list of dependent plugin ids in order of deactivation
    *
    * #### Notes
-   * The final item of the return list is always the plugin of interest.
+   * The final item of the returned list is always the plugin of interest.
    */
   export function findDependents(
     id: string,
--- a/packages/application/src/index.ts
+++ b/packages/application/src/index.ts
@@ -302,11 +302,8 @@ export class Application<T extends Widget = Widget> {
     // Resolve the optional services for the plugin.
     const optional = plugin.optional.map(t => this.resolveOptionalService(t));
 
-    // Create the array of promises to resolve.
-    const promises = required.concat(optional);
-
     // Setup the resolver promise for the plugin.
-    plugin.promise = Promise.all(promises)
+    plugin.promise = Promise.all([...required, ...optional])
       .then(services => plugin!.activate.apply(undefined, [this, ...services]))
       .then(service => {
         plugin!.service = service;
@@ -516,7 +513,7 @@ export class Application<T extends Widget = Widget> {
     Promise.all(promises).then(() => {
       this.attachShell(hostID);
       this.addEventListeners();
-      this._delegate.resolve(undefined);
+      this._delegate.resolve();
     });
 
     // Return the pending delegate promise.
--- a/packages/application/src/index.ts
+++ b/packages/application/src/index.ts
@@ -210,7 +210,7 @@ export class Application<T extends Widget> {
     }
 
     // Add the plugin to the plugin map.
-    this._pluginMap[data.id] = data;
+    this._pluginMap.set(data.id, data);
   }
 
   /**
@@ -228,6 +228,24 @@ export class Application<T extends Widget> {
   }
 
   /**
+   *
+   * @param id - id of the plugin to unregister
+   * @param force - whether to unregister the plugin even if it is active
+   */
+  unregisterPlugins(id: string, force?: boolean): void {
+    const data = this._pluginMap.get(id);
+    if (!data) {
+      return;
+    }
+
+    if (data.activated && !force) {
+      throw new Error('Plugin is still active.');
+    }
+
+    this._pluginMap.delete(id);
+  }
+
+  /**
    * Activate the plugin with the given id.
    *
    * @param id - The ID of the plugin of interest.
@@ -237,7 +255,7 @@ export class Application<T extends Widget> {
    */
   activatePlugin(id: string): Promise<void> {
     // Reject the promise if the plugin is not registered.
-    let data = this._pluginMap[id];
+    let data = this._pluginMap.get(id);
     if (!data) {
       return Promise.reject(new Error(`Plugin '${id}' is not registered.`));
     }
@@ -264,15 +282,15 @@ export class Application<T extends Widget> {
     // Setup the resolver promise for the plugin.
     data.promise = Promise.all(promises)
       .then(services => {
-        return data.activate.apply(undefined, [this, ...services]);
+        return data!.activate.apply(undefined, [this, ...services]);
       })
       .then(service => {
-        data.service = service;
-        data.activated = true;
-        data.promise = null;
+        data!.service = service;
+        data!.activated = true;
+        data!.promise = null;
       })
       .catch(error => {
-        data.promise = null;
+        data!.promise = null;
         throw error;
       });
 
@@ -299,7 +317,7 @@ export class Application<T extends Widget> {
    * the required services for the user's plugins will be resolved
    * automatically when the plugin is activated.
    */
-  resolveRequiredService<U>(token: Token<U>): Promise<U> {
+  async resolveRequiredService<U>(token: Token<U>): Promise<U> {
     // Reject the promise if there is no provider for the type.
     let id = this._serviceMap.get(token);
     if (!id) {
@@ -307,13 +325,14 @@ export class Application<T extends Widget> {
     }
 
     // Resolve immediately if the plugin is already activated.
-    let data = this._pluginMap[id];
+    let data = this._pluginMap.get(id)!;
     if (data.activated) {
       return Promise.resolve(data.service);
     }
 
     // Otherwise, activate the plugin and wait on the results.
-    return this.activatePlugin(id).then(() => data.service);
+    await this.activatePlugin(id);
+    return data.service;
   }
 
   /**
@@ -335,7 +354,7 @@ export class Application<T extends Widget> {
    * the optional services for the user's plugins will be resolved
    * automatically when the plugin is activated.
    */
-  resolveOptionalService<U>(token: Token<U>): Promise<U | null> {
+  async resolveOptionalService<U>(token: Token<U>): Promise<U | null> {
     // Resolve with `null` if there is no provider for the type.
     let id = this._serviceMap.get(token);
     if (!id) {
@@ -343,20 +362,19 @@ export class Application<T extends Widget> {
     }
 
     // Resolve immediately if the plugin is already activated.
-    let data = this._pluginMap[id];
+    let data = this._pluginMap.get(id)!;
     if (data.activated) {
       return Promise.resolve(data.service);
     }
 
     // Otherwise, activate the plugin and wait on the results.
-    return this.activatePlugin(id)
-      .then(() => {
-        return data.service;
-      })
-      .catch(reason => {
-        console.error(reason);
-        return null;
-      });
+    try {
+      await this.activatePlugin(id);
+      return data.service;
+    } catch (reason) {
+      console.error(reason);
+      return null;
+    }
   }
 
   /**
@@ -396,11 +414,13 @@ export class Application<T extends Widget> {
     let startups = Private.collectStartupPlugins(this._pluginMap, options);
 
     // Generate the activation promises.
-    let promises = startups.map(id => {
-      return this.activatePlugin(id).catch(error => {
+    let promises = startups.map(async id => {
+      try {
+        return await this.activatePlugin(id);
+      } catch (error) {
         console.error(`Plugin '${id}' failed to activate.`);
         console.error(error);
-      });
+      }
     });
 
     // Wait for the plugins to activate, then finalize startup.
@@ -634,7 +654,7 @@ namespace Private {
   /**
    * A type alias for a mapping of plugin id to plugin data.
    */
-  export type PluginMap = { [id: string]: IPluginData };
+  export type PluginMap = Map<string, IPluginData>;
 
   /**
    * A type alias for a mapping of service token to plugin id.
@@ -645,7 +665,7 @@ namespace Private {
    * Create a new plugin map.
    */
   export function createPluginMap(): PluginMap {
-    return Object.create(null);
+    return new Map<string, IPluginData>();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


fcollonval:
cc @jasongrout @krassowski

blink1073 (approved):
API changes LGTM!

afshin:
It took me several tries to `git checkout` this branch name 🤣

```zsh
git checkout "fix/278-Dynamic-extensions-reloading-\`IPlugindeactivate\`-\`ApplicationregisterPlugin\`-and-\`ApplicationdeactivatePlugin\`-proposal"
```

packages/application/src/index.ts
```diff
@@ -613,7 +714,20 @@ namespace Private {
     /**
      * The function which activates the plugin.
      */
-    readonly activate: (app: any, ...args: any[]) => any;
+    readonly activate: (
+      app: Application<Widget>,
+      ...args: Token<any>[]
```
afshin, line 719: We don't pass `Token` instances into the `activate` function, we pass the resolved item (which can by `any` thing).

packages/application/src/index.ts
```diff
@@ -613,7 +714,20 @@ namespace Private {
...
+    ) => Token<any> | Promise<Token<any>>;
```
afshin, line 720: The `activate` function returns `any` item that the `Token` represents, not the `Token` itself.

packages/application/src/index.ts
```diff
@@ -613,7 +714,20 @@ namespace Private {
...
+
+    /**
+     * The optional function which deactivates the plugin.
+     */
+    readonly deactivate:
+      | ((
+          app: Application<Widget>,
+          ...args: Token<any>[]
```
afshin, line 728: The `args` here follow the same pattern, they are not `Token` instances.

packages/application/src/index.ts
```diff
@@ -396,11 +495,13 @@ export class Application<T extends Widget> {
     let startups = Private.collectStartupPlugins(this._pluginMap, options);
 
     // Generate the activation promises.
-    let promises = startups.map(id => {
-      return this.activatePlugin(id).catch(error => {
+    let promises = startups.map(async id => {
```
afshin, line 498: The original version here seems more straightforward because it doesn't wrap the `activate` promise in an additional `await`.

packages/application/src/index.ts
```diff
@@ -299,21 +398,22 @@ export class Application<T extends Widget> {
...
     let id = this._serviceMap.get(token);
     if (!id) {
       return Promise.reject(new Error(`No provider for: ${token.name}.`));
     }
 
     // Resolve immediately if the plugin is already activated.
-    let data = this._pluginMap[id];
+    let data = this._pluginMap.get(id)!;
     if (data.activated) {
       return Promise.resolve(data.service);
```
afshin, line 411: We can just return `data.service` since you've converted it to `async`

packages/application/src/index.ts
```diff
@@ -299,21 +398,22 @@ export class Application<T extends Widget> {
    * the required services for the user's plugins will be resolved
    * automatically when the plugin is activated.
    */
-  resolveRequiredService<U>(token: Token<U>): Promise<U> {
+  async resolveRequiredService<U>(token: Token<U>): Promise<U> {
     // Reject the promise if there is no provider for the type.
     let id = this._serviceMap.get(token);
     if (!id) {
       return Promise.reject(new Error(`No provider for: ${token.name}.`));
```
afshin, line 405: We can just `throw` since you've converted it to `async`

afshin:
I opened [a PR against your fork](https://github.com/fcollonval/lumino/pull/5) which contains some of the review comments I made above and a few other changes. I didn't want to push directly on your branch in case you are in the middle of something or in case you disagree.

packages/application/src/index.ts
fcollonval: True, thanks I was trying to improve typing but yep, that change is definitely inappropriate.

packages/application/src/index.ts
fcollonval: I use my editor refactoring tool to switch to `async-await` as I find that syntax easier to read - especially for people unfamiliar with Promises. So I did not look too much to those - thanks for doing it.

fcollonval:
> It took me several tries to `git checkout` this branch name rofl

autogenerated name from issue title... sorry about that

> I opened https://github.com/fcollonval/lumino/pull/5 which contains some of the review comments I made above and a few other changes. I didn't want to push directly on your branch in case you are in the middle of something or in case you disagree.

Thanks a lot for the review and the commit to improve this PR. I have only one blocking comment about your PR regarding not using sequential deactivation of the plugins: https://github.com/fcollonval/lumino/pull/5/files#r956955307

afshin:
Thank you!

packages/application/src/index.ts
```diff
@@ -186,31 +210,31 @@ export class Application<T extends Widget> {
...
    * If the plugin provides a service which has already been provided
    * by another plugin, the new service will override the old service.
    */
   registerPlugin(plugin: IPlugin<this, any>): void {
-    // Throw an error if the plugin id is already registered.
-    if (plugin.id in this._pluginMap) {
-      throw new Error(`Plugin '${plugin.id}' is already registered.`);
+    // Throw an error if the plugin ID is already registered.
+    if (this._plugins.has(plugin.id)) {
+      throw new TypeError(`Plugin '${plugin.id}' is already registered.`);
```
vidartf, line 222: "The TypeError object represents an error when an operation could not be performed, typically (but not exclusively) when a value is not of the expected type."

This doesn't really seem appropriate here?

packages/application/src/index.ts
```diff
@@ -222,62 +246,133 @@ export class Application<T extends Widget> {
...
+    // Find the optimal deactivation order for plugins downstream of this one.
+    const manifest = Private.findDependents(id, this._plugins, this._services);
+    const downstream = manifest.map(id => this._plugins.get(id)!);
+
+    // Check that all downstream plugins can deactivate.
+    for (const plugin of downstream) {
+      if (!plugin.deactivate) {
+        throw new TypeError(
+          `Plugin ${plugin.id}#deactivate() method missing (depends on ${id})`
+        );
```
vidartf, line 356: This exception doesn't seem to correspond to the PR description: "an optional deactivate method which plugins may decide to implement. If it is provided, it should undo all the changes the activate method made, including removing the added commands, menus, etc. to complement the current activate() method."

afshin: Which part is problematic, do you mean because `deactivate` is optional?

This only happens if you try to deactivate something and either the thing you're deactivating or one of its downstream plugins cannot deactivate. Is this behavior incorrect?

vidartf: I guess there is an additional line further down in the OP that mentions it, so probably ok.

packages/application/src/index.ts
vidartf: I just wondered why a `TypeError` was chosen for all the errors thrown in this PR. The only argument for it would seem to be that it adds some context clues, and if the context isn't that you passed something of the wrong type, then it is only confusing.

afshin: There are three `ReferenceError`s, two `TypeError`s, one *new* `Error` and a couple old ones.

afshin: It was an attempt to make the errors more semantic, but if you thing it detracts, there's no technical need for them to be any specific variety of error.

vidartf: It was mostly that my impression was that `TypeError` and `ReferenceError` has certain *specific* meanings (the type of something was wrong, typically via `instanceof` check; and an unassigned variable was attempted to be dereferenced).

afshin: I was checking the MDN pages for the errors when I picked these errors.

> The [**ReferenceError**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ReferenceError) object represents an error when a variable that doesn't exist (or hasn't yet been initialized) in the current scope is referenced.

My rationale was that not being registered yet is akin to "hasn't yet been initialized".

> A [**TypeError**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) may be thrown when:
> 
>    * an operand or argument passed to a function is incompatible with the type expected by that operator or function; or
>    * when attempting to modify a value that cannot be changed; or
>    * when attempting to use a value in an inappropriate way.

I think, for example, if a plugin does not have a `deactivate` method, trying to deactivate it is using it in an inappropriate way.

---

I don't feel dogmatic about these choices. If you feel they are inappropriate, I'm happy to change them back to generic `Error` instances. But I do think a reasonable reading of the documentation for these errors allows some wiggle room as well.

Do you think we should change it?

fcollonval:

jtpio:
> I won't backport it because `Map` does not exist in older es version.

So for reference this seems to have been backported anyway in https://github.com/jupyterlab/lumino/pull/485?

PR #378, opened 2022-08-26
Iterate on iterators

afshin:
Using [`yield*`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield*) where relevant means we can deprecate several more functions:
* `chain<T>(...)`
* `empty<T>(...)`
* `once<T>(...)`
* `repeat<T>(...)`

cc: @vidartf


blink1073 (approved):
Nice, thanks!

fcollonval (approved):
Thanks

PR #382, opened 2022-08-29
Fix service type

afshin:


.github/workflows/license-header.yml
```diff
@@ -32,6 +32,19 @@ jobs:
...
+            packages/**/*.js
+            packages/**/*.ts
+            packages/**/*.tsx
+
+      - name: Did things change?
+        if: steps.changed-files.outputs.any_changed == 'true'
+        shell: bash -l {0}
+        run: |
+          echo "Yes! Some things changed!"
+          echo | git status --short --branch
```
fcollonval, line 47: @afshin did you encounter an unexpected file being committed or is this just a fail-safe?

Note: to be sure to display the changes detected by the previous action, I would recommend using the following snippet:

https://github.com/jupyterlab/lumino/blob/d6a592ce57f3f683335e4e890afb75b540135e1f/.github/workflows/check-api-changes.yml#L64-L67

fcollonval (approved):
Thanks @afshin

.github/workflows/license-header.yml
afshin: No new files are committed, but for some reason the next job (the one that does the `git push`) fails, presumably because there is nothing to push? I'm not 100% sure what is happening. I reverted my modifications to the CI job because I was just experimenting.

vidartf:
For those of us trying to follow along: What exactly does this change do? What problems did the old code cause, and how is this a fix?

vidartf:
Oh, I see now that this is a cleanup after https://github.com/jupyterlab/lumino/pull/377.

afshin:
> For those of us trying to follow along: What exactly does this change do? What problems did the old code cause, and how is this a fix?
> 
> Oh, I see now that this is a cleanup after https://github.com/jupyterlab/lumino/pull/377.

Right, this puts the signatures back to what they originally were.

Issue #339, opened 2022-08-08
Add a README file to each package


afshin:
We should have at least a basic `REAMDE.md` file in each Lumino package. In addition to helping new contributors navigate this mono-repo, it also allows the NPM page for each Lumino package to display some useful information to consumers of the package.

gabalafou:
As far as I can tell, none of the packages currently have a `README.md`. If somebody could make a README for one package, I could use that as a template to make more for other packages.

fcollonval:
Something like the following would be nice

---

# @lumino/application

A Lumino package to create web application composed of plug-able plugins.

By default, it contains a `CommandRegistry` using [@lumino/commands](../commands), a `ContextMenu` and a view called _shell_ that must inherits from `Widget` class (the two later classes are defined in [@lumino/widgets](../widgets)).

gabalafou:
Thanks for the template @fcollonval, I'll work on extending this template to all of the Lumino packages. I don't have permission to assign this task to myself. Could you assign it to me? Thank you!

afshin:
@gabalafou I've added you to the "Triage" team for the `jupyterlab` GitHub org, you should be able to assign and label things, etc., now

gabalafou:
I'm punting on this for another week because I wanted to finish https://github.com/jupyterlab/lumino/issues/332.

PR #386, opened 2022-08-31
Make `Poll` an async iterable

### LINKED ISSUE #385: Make `Poll` an async iterator
`Poll` instances are already multi-paradigm, supporting ticks that are captured via promises and also signals.

With the introduction of async iterators (`ES2018`), `Poll` instances should implement a `Symbol.asyncIterator` method to support that method of capturing a stream of ticks as well.

afshin:
In this PR, the `IPoll` interface and the `Poll` class extend `AsyncIterable` (by implementing a [`Symbol.asyncIterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator) method) to support [`for-await...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of) loops.

This PR also adds a `README.md` file for the `@lumino/polling` package.

This functionality may be a natural fit for streams of events from the new [Jupyter event system](https://github.com/jupyter-server/jupyter_server/issues/780), other `WebSocket` connections, or other stream sources.

cc: @Zsailer @andrii-i @davidbrochart @3coins @vidartf (since you all have some interest in events and/or iterators)

---

### Example usage
The tests in this PR are illustrative. Here is one of them with some annotation:

Here, we set up the testing state variables and create a new `Poll` instance.
```typescript
const total = 2;
let i = 0;
poll = new Poll({
  auto: false,
  factory: () => Promise.resolve(++i > total ? poll.dispose() : void 0),
  frequency: { interval: Poll.IMMEDIATE },
  name: '@lumino/polling:Poll#[Symbol.asyncIterator]-3'
});
const expected = `started ${'resolved '.repeat(total)}disposed`;
const ticker: IPoll.Phase<any>[] = [];
```
Then the poll is started:
```typescript
void poll.start();
```
Instead of connecting to the `ticked` signal or awaiting the `tick` promise, we can now use a `for-await...of` loop:
```typescript
for await (const state of poll) {
  ticker.push(state.phase);
  if (poll.isDisposed) {
    break;
  }
}
```
And we check to make sure the results are as expected:
```typescript
// ticker and expected are both: 'started resolved resolved disposed'
expect(ticker.join(' ')).to.equal(expected);
```
### Note for consumers of async iterators
The `Poll` class itself only uses `ES6` (and `DOM`) types in its `lib` collection **but** in order to use `for-await...of` loops in TypeScript, you will need to use `ES2018` or above in your `lib` array in `tsconfig.json`.

---

Fixes #385 
cf. partially addresses https://github.com/jupyterlab/lumino/issues/339

### DIFF
--- a/packages/polling/src/poll.ts
+++ b/packages/polling/src/poll.ts
@@ -157,10 +157,9 @@ export class Poll<T = any, U = any, V extends string = 'standby'>
    * Return an async iterator that yields every tick.
    */
   async *[Symbol.asyncIterator](): AsyncIterableIterator<IPoll.State<T, U, V>> {
-    yield this.state;
     while (!this.isDisposed) {
-      await this.tick.catch(() => undefined);
       yield this.state;
+      await this.tick.catch(() => undefined);
     }
   }
 
--- a/packages/polling/src/poll.ts
+++ b/packages/polling/src/poll.ts
@@ -154,6 +154,17 @@ export class Poll<T = any, U = any, V extends string = 'standby'>
   }
 
   /**
+   * Return an async iterator that yields every tick.
+   */
+  async *[Symbol.asyncIterator](): AsyncIterableIterator<IPoll.State<T, U, V>> {
+    yield this.state;
+    while (!this.isDisposed) {
+      await this.tick.catch(() => undefined);
+      yield this.state;
+    }
+  }
+
+  /**
    * Dispose the poll.
    */
   dispose(): void {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -800,7 +800,7 @@ github_url: 'https://github.com/jupyterlab/lumino/blob/main/CHANGELOG.md'
     @lumino/example-datagrid@0.11.0
     @lumino/datagrid@0.14.0
 
-- Make private \_drawCornerHeaderRegion protected drawCornerHeaderRegion [#116](https://github.com/jupyterlab/lumino/pull/116) ([@lmcnichols](https://github.com/lmcnichols))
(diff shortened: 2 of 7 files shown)


blink1073:
Maybe add those notes to the README for polling?

afshin:
I was going to ask "what README?" but the question answers itself 🤣

blink1073 (approved):
Nice, thanks!

packages/polling/src/poll.ts
```diff
@@ -153,6 +153,17 @@ export class Poll<T = any, U = any, V extends string = 'standby'>
...
 
+  /**
+   * Return an async iterator that yields every tick.
+   */
+  async *[Symbol.asyncIterator](): AsyncIterableIterator<IPoll.State<T, U, V>> {
+    yield this.state;
+    while (!this.isDisposed) {
+      await this.tick.catch(() => undefined);
+      yield this.state;
+    }
```
fcollonval, line 164: Should we do it the other way around to always check for disposable state:

```suggestion
    while (!this.isDisposed) {
      yield this.state;
      await this.tick.catch(() => undefined);
    }
```

afshin: This suggestion will work, but it will never `yield` the `disposed` state and one of the tests I wrote will fail.

There is a real question of whether we *want* to `yield` the disposed state, but that was my motivation for writing it this way.

afshin: Hm, we do *not* emit a `ticked` signal when the poll is disposed. And we do not `resolve` the `tick` promise, we `reject` it. So maybe we should not `yield` when disposed either.

afshin: I think I talked myself into it.

fcollonval (approved):
Thanks

PR #391, opened 2022-09-06
Fix bug that prevents any startup plugins from activating

afshin:
This  makes the current version of `@lumino/application` unusable.


blink1073:
Releasing now

Issue #392, opened 2022-09-06
Use of requestAnimationFrame means activities cannot take place in background


thetorpedodog:
## Description

We run a JupyterHub instance on top of Kubernetes. Launching a new JupyterLab notebook instance takes a minute or so, meaning users often switch to another tab and do something else while waiting for the notebook to load. However, other setup activities, like loading notebook files, starting kernels, and more does not actually take place until the user switches back. This may take another minute or two (during which they are likely to switch away again).

I traced this delay back to the fact that Lumino uses `requestAnimationFrame` to defer tasks. This does not run while the tab is in the background:

https://github.com/jupyterlab/lumino/blob/2d9b7aae5fd35772327f5b4e787e5aa435953296/packages/polling/src/poll.ts#L15-L18

## Reproduce

The full user experience won’t reproduce locally, but you can get a flavor of it and see the symptoms:

1. Launch a Jupyter Lab server (v3.4.5 will work), but don’t open the browser (`jupyter lab --no-browser`)
2. Open a new browser tab and open the developer tools, so you can capture network requests and activity.
3. Paste the URL that Jupyter Lab spit out at you. Hit enter and immediately switch to another tab (holding CTRL while hitting both ENTER and then PGUP (or PGDN) immediately after worked well for me).
4. Chill for a bit.
5. Return to the Jupyter Lab tab. Note that none of your notebooks etc. have loaded.

The long wait ↹’d in this screenshot is the time that I had another tab open.

![screenshot of dev tools showing a long wait](https://user-images.githubusercontent.com/6622003/188706662-eb93af22-a268-47f0-8fca-7bb6a63531b5.png)

## Expected behavior

I would expect my session to restore itself and other things to load even if I had my tab open in the background.

## Caveats and ideas

There probably is existing code that depends upon the existing “don‘t do work in the background” behavior. These are a few ideas for how to deal with that, listed from what seems (from my outsider’s perspective) from better to worse/harder:

- One solution might be to schedule deferred tasks immediately with `setTimeout(...)` if a `requestAnimationFrame` has not happened yet, and then once the first animation frame has been requested, use that instead.
- The existing code does say that the (nonstandard) `setImmediate` is an option if `requestAnimationFrame` does not exist. `setTimeout` (with no timeout) is the standard version of that and it could be used.
- Perform deferred tasks either on the next animation frame or on some longer timeout.
[...]

PR #395, opened 2022-09-07
Switch to using `setTimeout` for all deferred processing.

thetorpedodog:
As noted on issue #392, using `requestAnimationFrame` means that no deferred tasks can take place while the user has the browser window in the background. This can lead to a bad experience for long-lasting tasks, particularly setup tasks (when a user might start JupyterLab, switch somewhere else when it loads, and then come back).

This change instead uses `setTimeout` (with a timeout of zero, either implicitly or explicitly), which immediately puts the task on the event loop to be run (subject to less severe browser throttling). This allows these tasks to occur in the background (and should hopefully not be too taxing on user resources).

https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#timeouts_in_inactive_tabs

> Firefox Desktop and Chrome both have a minimum timeout of 1 second > for inactive tabs.

### DIFF
--- a/packages/polling/src/poll.ts
+++ b/packages/polling/src/poll.ts
@@ -326,13 +326,7 @@ export class Poll<T = any, U = any, V extends string = 'standby'>
   private _state: IPoll.State<T, U, V>;
   private _tick = new PromiseDelegate<this>();
   private _ticked = new Signal<this, IPoll.State<T, U, V>>(this);
-  /*
-   * If set, the timeout handle for this poller. We use this complex formulation
-   * for the type since the NodeJS environment has a different return type
-   * (`NodeJS.Timeout`) from the browser environment (`number`) and we want this
-   * to compile in both places.
-   */
-  private _timeout?: ReturnType<typeof setTimeout>;
+  private _timeout?: ReturnType<typeof setTimeout>; // Support node and browser.
 }
 
 /**
@@ -401,8 +395,7 @@ export namespace Poll {
     standby?: Standby | (() => boolean | Standby);
   }
   /**
-   * An interval value that indicates the poll should tick immediately,
-   * i.e. 0 msec.
+   * An interval value (0ms) that indicates the poll should tick immediately.
    */
   export const IMMEDIATE = 0;
 
--- a/packages/polling/src/poll.ts
+++ b/packages/polling/src/poll.ts
@@ -10,22 +10,6 @@ import { ISignal, Signal } from '@lumino/signaling';
 import { IPoll } from './index';
 
 /**
- * A function to defer an action immediately.
- */
-const defer =
-  typeof requestAnimationFrame === 'function'
-    ? requestAnimationFrame
-    : setImmediate;
-
-/**
- * A function to cancel a deferred action.
- */
-const cancel: (timeout: any) => void =
-  typeof cancelAnimationFrame === 'function'
-    ? cancelAnimationFrame
-    : clearImmediate;
-
-/**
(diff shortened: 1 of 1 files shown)


thetorpedodog:
@meeseeksdev tag bug

packages/polling/src/poll.ts
```diff
@@ -347,7 +326,13 @@ export class Poll<T = any, U = any, V extends string = 'standby'>
...
   private _tick = new PromiseDelegate<this>();
   private _ticked = new Signal<this, IPoll.State<T, U, V>>(this);
-  private _timeout: any = -1;
+  /*
+   * If set, the timeout handle for this poller. We use this complex formulation
+   * for the type since the NodeJS environment has a different return type
+   * (`NodeJS.Timeout`) from the browser environment (`number`) and we want this
+   * to compile in both places.
+   */
+  private _timeout?: ReturnType<typeof setTimeout>;
```
fcollonval, line 335: nice

packages/polling/src/poll.ts
```diff
@@ -241,17 +224,18 @@ export class Poll<T = any, U = any, V extends string = 'standby'>
     this._tick = scheduled;
 
     // Clear the schedule if possible.
-    if (last.interval === Poll.IMMEDIATE) {
-      cancel(this._timeout);
-    } else {
-      clearTimeout(this._timeout);
-    }
+    clearTimeout(this._timeout);
```
fcollonval, line 227: Should we test for non-undefined timeout id?

```suggestion
    if (typeof this._timeout === 'number') {
      clearTimeout(this._timeout);
    }
```

fcollonval (approved):
Thanks @thetorpedodog

fcollonval:
We probably want to backport this to Lumino 1.x to include it in a JupyterLab patch release

thetorpedodog:
in this case I will pull it out of the drafts so it can be merged!

packages/polling/src/poll.ts
thetorpedodog: for the record I [learned this from Stack Overflow](https://stackoverflow.com/a/56239226/39808)

thetorpedodog:
> We probably want to backport this to Lumino 1.x to include it in a JupyterLab patch release

Just did a test and the patch applies cleanly, so this can be automated post-submission: https://github.com/TileDB-Inc/lumino-dev/commit/1c2e89eb31bf9edd09c2db05d9cb192cb3bebb49

afshin:
Thank you, @thetorpedodog! Great work tracking down this issue from several layers downstream.

I changed your comment on the `_timeout` variable to make a terse in-line comment instead of a docstring style just to keep it more idiomatic with the rest of the library, I hope that's okay!

packages/polling/src/poll.ts
afshin, line 335:
```suggestion
  private _timeout?: ReturnType<typeof setTimeout>; // Support node and browser.
```

packages/polling/src/poll.ts
```diff
@@ -416,7 +401,8 @@ export namespace Poll {
     standby?: Standby | (() => boolean | Standby);
   }
   /**
-   * An interval value that indicates the poll should tick immediately.
+   * An interval value that indicates the poll should tick immediately,
+   * i.e. 0 msec.
```
afshin, line 405:
```suggestion
   * An interval value (0ms) that indicates the poll should tick immediately.
```

afshin (approved):

blink1073:

Issue #325, opened 2022-08-01
Update and prune dependencies in Lumino 2


afshin:
We should audit the `dependencies` and `devDependencies` of the packages (and the meta-package) and make sure they are up to date. We have a lot of warnings in the build logs and console output and we are using outdated versions of tools like ESLint, among others.

### Relevant issues and PRs
- [x] #319
- [x] #330
- [x] #332
- [x] #396

fcollonval:
As commented in #330, it will be nice to drop webpack and to use rollup for the tests to remove any inconsistency between the packager used for publishing and for testing.

blink1073:
The only remaining warning is:

```
warning "@jupyterlab/buildutils > verdaccio > clipanion@3.1.0" has unmet peer dependency "typanion@*".
```

I'm closing this one as complete.

PR #396, opened 2022-09-09
Remove some webpack artifacts from `devDependencies`

afshin:
cf. https://github.com/jupyterlab/lumino/issues/325


blink1073 (approved):
Thanks!

Issue #400, opened 2022-09-09
Handle deferred processing when `document` is hidden


afshin:
## Context

We merged a recent PR from @thetorpedodog (thank you!) modifying the `Poll` class to stop using `requestAnimationFrame`/`setImmediate` in favor of `setTimeout` every time:
- https://github.com/jupyterlab/lumino/pull/395
in order to resolve:
- https://github.com/jupyterlab/lumino/issues/392

## Problem

Upon further reflection and talking through this with @blink1073 and @fcollonval, I think this may not solve the original issue that Paul is trying to solve. Because:

- `Poll.IMMEDIATE` was a sentinel value (`0`) that told the `Poll` to use `requestAnimationFrame`/`setImmediate`
- However that was only happening when polling was explicitly so to happen immediately. This happens nowhere in JupyterLab core as far as I can see.

Most polls are set to reasonably long intervals. The more likely reason is two-fold:
- Use of the `when-hidden` standby mode for `Poll` instances takes place immediately when the window is hidden.
- Even if crucial `Poll` instance tick one more time, all `Widget` rendering in Lumino applications happens using the `MessageLoop`, which uses `requestAnimationFrame`/`setImmediate`.

## Proposal

- Let's release JupyterLab `4.0.0a29` with @thetorpedodog's fix to see if it resolves the issue. My suspicion is that it will not.
- Let's release JupyterLab `4.0.0a30` with the following changes:
  - Create a `schedule()` and `unschedule()` abstraction utility to use in `Poll` and `MessageLoop` that picks the defer function based on whether the client is a browser and whether it is hidden.
  - Modifying the behavior of the `when-hidden` standby to tick a few more times after being hidden (perhaps a default value of `1` lingering tick that can be changed by a client)

This way there are two easy to validate versions of JupyterLab that attempt to resolve this issue that we can compare.

thetorpedodog:
> Modifying the `MessageLoop` to use `setTimeout` when the window is hidden but to use `requestAnimationFrame`/`setImmediate` otherwise

Isn’t this essentially equivalent to just using `setTimeout(..., 0)` all the time?

Also: leaving myself a note to trace where the `requestAnimationFrame` stuff came from, because I am almost certain it went through the `Poll` path, though it might have gone through the message loop and *then* a Poll.

afshin:
@vidartf,

> Note: The PR was backported to the 1.x branch. Will that be an issue?

We will not release a patch release of the 1.x branch until we have an answer to this.

[...]

PR #402, opened 2022-09-11
Handle deferred processing when `document` is hidden

afshin:
This PR does the following.
- [x] Modifies the behavior of the polls' `when-hidden` standby to tick once after being hidden (the default `options.linger` value of `1` time can be overridden at instantiation time)
- [x] Modifies the `MessageLoop` to use a new promise-based implementation of `schedule` for immediate deferral to support always running in background tabs
- [x] Modifies `Poll` to export a protected `hidden` attribute to indicate to sub-classes when the document is hidden.

We can backport this to 1.x as well.

**CC:** @thetorpedodog – I kept your suggested `setTimeout` for `Poll` instances because almost all polls have a nonzero `interval` anyway and it simplifies that code. I updated `MessageLoop` to work in the background without having to resort to `setTimeout`.

---

Fixes https://github.com/jupyterlab/lumino/issues/400


packages/coreutils/src/schedule.ts
```diff
@@ -0,0 +1,49 @@
...
+  return setImmediate(fn);
+}
+
+/**
+ * Unschedules a function invocation if possible.
+ *
+ * @param handle The handle for the callback to unschedule.
+ */
+export function unschedule(handle: ScheduleHandle): void {
+  if (BROWSER && document.visibilityState === 'hidden') {
```
blink1073, line 42: I think we need store the original state on the handle, because the document visibility can change between schedule and unschedule.

afshin: Ugh. I thought about this a lot and I don't want to do it because of the performance hit of tracking the type of function that was invoked. It seems unlikely that in however long it takes to execute an animation frame, the page goes hidden just before the execution completes.

- Keep in mind that this is meant to be near instantaneous and cancel is not guaranteed.
- `unschedule` will not throw an error if it gets an errant handle.

I don't know, what do you think?

blink1073: Yeah that's fair

blink1073 (approved):
Thanks!

vidartf:
> [...] schedule uses `setTimeout` when the window is hidden and uses `requestAnimationFrame`/`setImmediate` otherwise

What are the nuances of this? If a window is visible, and I `requestAnimationFrame` and then hide the window, will the trigger fire the next time I show the window? If that is the case, it seems like this will cause some inconsistent behavior?

Also: Do we know what was the motivation for choosing `requestAnimationFrame` rather than `setTimeout` for the message loop originally? It seems like not driving the page when it is hidden is considerate of system resources, and this will change that.

afshin:
I think the primary original motivation wasn't about consideration of system resources so much as it was about responsiveness. Consider this example somebody created that is very illustrative: https://codepen.io/YOONBYEONGIN/pen/RwxrEPN

It seems pretty unlikely that you *could* hide the window in the time between when you invoke `requestAnimationFrame` and the time the function actually fires.

vidartf:
> It seems pretty unlikely that you could hide the window in the time between when you invoke requestAnimationFrame and the time the function actually fires.

@afshin It might be unlikely, but if it happens once a month for a heavy user, and the consequence is catastrophic, then it still matters. So the question the becomes, what will be the failure mode if one callback gets stuck waiting on an animation frame while other callbacks keep firing with timeouts? Do we end up processing messages out of order? I don't think "probably nothing bad will happen" is an ok answer in this case. Waiting a bit for something after resuming a browser tab is in general a more ok user experience than undefined behavior (you can fit a lot into "undefined behavior").

afshin:
I should have stated that more strongly: you have <1ms to switch tabs from the time you have invoked an animation frame to the time it fires. The only time `unschedule()` is invoked by the message loop logic, it is in the `.flush()` function, which is invoked by Lumino code only to support older IE and Edge. It's not actually invoked otherwise, so I am not coming up with a scenario I can imagine where it causes loop events ever firing in the incorrect order or leading to a catastrophic outcome.

I'll give it some more thought.

Does anyone else who has considered this part of the code base have any thoughts?

afshin:
>  if one callback gets stuck waiting on an animation frame while other callbacks keep firing with timeouts

Hm. I'll think about if this is possible -- and if so, what to do.

afshin:
After a lot of back and forth, this is ready for review.

blink1073 (approved):
LGTM!

packages/polling/src/poll.ts
```diff
@@ -395,7 +422,7 @@ export namespace Poll {
     standby?: Standby | (() => boolean | Standby);
   }
   /**
-   * An interval value (0ms) that indicates the poll should tick immediately.
+   * An interval value that indicates the poll should tick immediately.
```
fcollonval, line 425: Probably worth mentioning the unit of the value.

```suggestion
   * An interval value in ms that indicates the poll should tick immediately.
```

packages/polling/src/poll.ts
```diff
@@ -500,4 +515,44 @@ namespace Private {
...
+      return false;
+    }
+    document.addEventListener('visibilitychange', () => {
+      hidden = document.visibilityState === 'hidden';
+    });
+    document.addEventListener('pagehide', () => {
+      hidden = document.visibilityState === 'hidden';
+    });
+    return document.visibilityState === 'hidden';
+  })();
```
fcollonval, line 540: To avoid external modification (aka hidden should not be modified except by the listeners), what about doing something like:

```suggestion
  let _hidden = (() => {
    if (typeof document === 'undefined') {
      return false;
    }
    document.addEventListener('visibilitychange', () => {
      _hidden = document.visibilityState === 'hidden';
    });
    document.addEventListener('pagehide', () => {
      _hidden = document.visibilityState === 'hidden';
    });
    return document.visibilityState === 'hidden';
  })();
  
  export function isHidden(): boolean {
    return _hidden;
  }
```

fcollonval (changes requested):
Thanks @afshin 

I have two comments.

packages/polling/src/poll.ts
afshin: I think that because `hidden` is already in the `Private` namespace, it is already protected from interference: it's true that something else in `poll.ts` could modify but nothing externally can change it.

It follows the same convention as other exported variables in private namespaces we use (e.g., in `@lumino/keyboard` and `@lumino/signaling`.

fcollonval: Thanks for the comment

fcollonval (approved):

fcollonval:
@afshin should we move forward with this one?

packages/messaging/src/index.ts
```diff
@@ -195,6 +195,29 @@ export type MessageHook =
...
+  const schedule = (
+    resolved =>
+    (fn: () => unknown): (() => void) => {
+      let rejected = false;
+      resolved.then(() => !rejected && fn());
+      return () => {
+        rejected = true;
+      };
+    }
+  )(Promise.resolve());
```
vidartf, line 219: As mentioned in last weeks lab dev meeting, using a Promise or async will schedule the task as a microtask, compared to `requestAnimationFrame` / `setTimeout` etc that will schedule it as a macro task. This difference is likely to be significant (@afshin has the details).

afshin: It is safe in this circumstance because message propagation is both lightweight and synchronous. This is the reason why the same technique is unavailable to us in `Poll`.

afshin: > @afshin should we move forward with this one?

Yes, I wanted to give @vidartf a chance to look, but I'm happy to move ahead.

PR #406, opened 2022-09-14
Improve title on accordion label

fcollonval:


fcollonval:

vidartf:
For those of us following along on the repo: What issue does this fix? Why is it important to have the title fall back to the label if the caption is not present? Please make some notes in either the PR or the commit so that the rest of us can understand what is going on, or if someone in the future is wondering why this change was made, they can understand.

PR #411, opened 2022-09-14
Fix resizing and mode switching in `DockPanel`

afshin:
This PR:

- Fixes resizing in `DockPanel`
- Fixes switching between `single-document` and `multiple-document` in `DockLayout`
- Fixes the broken shortcut characters in the example pages
- Adds an empty test to the `docklayout.spec.ts` file

This came up in the Lumino Gitter channel:

https://gitter.im/jupyterlab/lumino?at=6321ba7faa091774290be83a

### DIFF
--- a/packages/widgets/src/docklayout.ts
+++ b/packages/widgets/src/docklayout.ts
@@ -145,10 +145,8 @@ export class DockLayout extends Layout {
    * #### Notes
    * This iterator includes the generated tab bars.
    */
-  *[Symbol.iterator](): IterableIterator<Widget> {
-    if (this._root) {
-      yield* this._root.iterAllWidgets();
-    }
+  [Symbol.iterator](): IterableIterator<Widget> {
+    return this._root ? this._root.iterAllWidgets() : Private.empty;
   }
 
   /**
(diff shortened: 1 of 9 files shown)


packages/widgets/src/docklayout.ts
```diff
@@ -306,10 +306,10 @@ export class DockLayout extends Layout {
       mainConfig = null;
     }
 
-    // Create iterators over the old content.
-    let oldWidgets = this.widgets();
-    let oldTabBars = this.tabBars();
-    let oldHandles = this.handles();
+    // Create a snapshot of the old content.
```
afshin, line 309: This used to work because the old `IIterator` instances were instantiated *before* `this._root` was set to `null` and afforded iteration even after line 215, but now that `this.tabBars()` is itself an iterator, it yields nothing after `this._root` has been set to `null`, so we must instead capture a snapshot of the old content before nullifying the root.

packages/widgets/src/docklayout.ts
```diff
@@ -306,10 +306,10 @@ export class DockLayout extends Layout {
...
+    let oldWidgets = Array.from(this.widgets());
+    let oldTabBars = Array.from(this.tabBars());
+    let oldHandles = Array.from(this.handles());
```
afshin, line 312: This used to work because the old `IIterator` instances were instantiated *before* `this._root` was set to `null` and afforded iteration even after line 315, but now that `this.tabBars()` is itself an iterator, it yields nothing after `this._root` has been set to `null`, so we must instead capture a snapshot of the old content before nullifying the root.

cc: @fcollonval

blink1073 (approved):
LGTM!

fcollonval (approved):
Thanks a lot for digging this one up @afshin 

Let's merge and I'll do an alpha release.

PR #415, opened 2022-09-15
Follow on to #411 addressing `yield*`

afshin:
This PR:
- adds a section in the migration guide for `yield*` usage
- un-deprecates `empty()`

cc: @vidartf, you might find this and https://github.com/jupyterlab/lumino/pull/411 interesting


fcollonval (approved):
Thanks

PR #420, opened 2022-09-21
Add section to README: Learning Resources

gabalafou:
This video series was mentioned in the semi-weekly Juptyer Accessibility call, so I thought it should be added to the README.


blink1073 (approved):
Thanks!

PR #421, opened 2022-09-22
Add DockLayout tests

### LINKED ISSUE #410: Write more tests for `DockPanel`
`DockPanel` tests are very sparse and `DockLayout` tests are totally empty. It would help us catch errors before shipping if we add some tests here.

3coins:
Fixes #410

### DIFF
--- a/packages/widgets/tests/src/docklayout.spec.ts
+++ b/packages/widgets/tests/src/docklayout.spec.ts
@@ -1,8 +1,227 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
+import { Message } from '@lumino/messaging';
+import { DockLayout, TabBar, Widget } from '@lumino/widgets';
+import { expect } from 'chai';
+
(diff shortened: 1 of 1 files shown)


packages/widgets/tests/src/docklayout.spec.ts
```diff
@@ -1,8 +1,264 @@
...
+    });
+    describe('#removeWidget()', () => {
+      it.skip('should have some tests');
+    });
+    describe('#hitTestTabAreas()', () => {
+      it.skip('should have some tests');
+    });
+    describe('#attachWidget()', () => {
+      const hook = new LogHook();
+      expect(hook.messages).to.be.empty;
```
fcollonval, line 257: Let's keep it no code as there is no test

```suggestion
```

fcollonval (changes requested):
Thanks @3coins 

I have a minor suggestion but the rest looks good.

fcollonval (approved):
Thanks for this one @3coins

PR #446, opened 2022-10-20
Datagrid: Do not prevent page scroll if we are not actually scrolling the grid

martinRenou:
The Datagrid was always preventing the propagation of the wheel event, preventing the scrolling of the page when using the wheel on top of a grid that does not need scrolling. This PR fixes it by only stopping the propagation if the grid actually needs scrolling.

cc. @ibdafna 

https://user-images.githubusercontent.com/21197331/196924993-bb0c246d-e598-4597-b30e-5c80684f264a.mp4

### DIFF
--- a/packages/datagrid/src/basicmousehandler.ts
+++ b/packages/datagrid/src/basicmousehandler.ts
@@ -682,8 +682,23 @@ export class BasicMouseHandler implements DataGrid.IMouseHandler {
         throw 'unreachable';
     }
 
-    // Scroll by the desired amount.
-    grid.scrollBy(dx, dy);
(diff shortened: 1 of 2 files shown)


fcollonval (approved):
Thanks @martinRenou

ibdafna:
@fcollonval I think this change is missing from yesterday's release?

fcollonval:
@ibdafna Monday release was targeting 1.x not 2.x. Is this PR for 1.x too? As I did not follow closely the changes may for Lumino 2 on the datagrid and the milestone was not set I did not backport it.

Issue #448, opened 2022-10-26
DataGrid is slow when cells contain a lot of (long) text


krassowski:
## Description

Rendering, and scrolling within DataGrid is very lagy when large text is present in cells. This is due to `CanvasRenderingContext2D.measureText` invocations.

- Firefox: 8.5 seconds
- Chrome: 11.5 seconds

Firefox:

![Screenshot from 2022-10-27 00-15-55](https://user-images.githubusercontent.com/5832902/198156234-df8e5c3c-ebcd-4f69-b570-e9c573e0a5bf.png)

![Screenshot from 2022-10-27 00-16-16](https://user-images.githubusercontent.com/5832902/198156218-885fd8e7-c89c-4785-9101-c0e08665457d.png)

## Reproduce

https://github.com/jupyterlab/jupyterlab-plugin-playground with:

<details>

```ts
import {
  JupyterFrontEnd,
  JupyterFrontEndPlugin,
} from '@jupyterlab/application';
import { MainAreaWidget } from '@jupyterlab/apputils';
import { DataGrid, DataModel, BasicKeyHandler, BasicMouseHandler, BasicSelectionModel } from '@lumino/datagrid';

class LargeDataModel extends DataModel {
  rowCount(region: DataModel.RowRegion): number {
    return region === 'body' ? 1000 : 1;
  }

  columnCount(region: DataModel.ColumnRegion): number {
    return region === 'body' ? 10 : 1;
  }

  data(region: DataModel.CellRegion, row: number, column: number): any {
    if (region !== 'body') {
      return `${row}, ${column}`;
    }
    return 'x'.repeat(10*1024);
  }
}

const extension: JupyterFrontEndPlugin<void> = {
  id: 'datagrid',
  autoStart: true,
  activate: (
    app: JupyterFrontEnd
  ) => {
    const model = new LargeDataModel();
    const content = new DataGrid();
  
    content.dataModel = model;
    // (handlers are not required for reproduction)
    content.keyHandler = new BasicKeyHandler();
    content.mouseHandler = new BasicMouseHandler();
    content.selectionModel = new BasicSelectionModel({
      dataModel: model
    });
    const widget = new MainAreaWidget({content});
    widget.id = 'datagrid-example';
    widget.title.label = 'Slow Datagrid Example';
    app.shell.add(widget, 'main');
  },
};

export default extension;
```

</details>

This is a bit pathological, but I got worse with just one column of long strings (CSS content of lab one sheet per row)

[...]

PR #451, opened 2022-10-31
Disable text eliding by default

ibdafna:
Addresses #448

### DIFF
--- a/packages/datagrid/src/hyperlinkrenderer.ts
+++ b/packages/datagrid/src/hyperlinkrenderer.ts
@@ -162,6 +162,12 @@ export class HyperlinkRenderer extends TextRenderer {
     gc.textAlign = hAlign;
     gc.textBaseline = 'bottom';
 
+    // Terminate call here if we're not eliding or wrapping text
+    if (elideDirection === 'none' && !wrapText) {
+      gc.fillText(text, textX, textY);
+      return;
+    }
+
     // The current text width in pixels.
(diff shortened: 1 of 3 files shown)


fcollonval (commented):
Thanks @ibdafna could you run `yarn api` to make the CI check happy :wink: ?

Does it solves the issue or should we let the issue opened?

krassowski:
I would say that this does not fix the issue, since we can make vast algorithmic improvements (and even keep eliding enabled by default if we do that). While, it certainly helps to be able to disable eliding, ideally it could be disabled/enabled on per-column basis.

ibdafna:
@krassowski This is a good start, and it does address the issue. We can open separate PRs for improving the algorithms of text eliding and wrapping. But I'd like to iterate again, there is no easy way to significantly reduce the time elide takes as it needs to make at least one call per cell to the measureText function. You can certainly enable or disable this on a per column basis by setting a custom TextRenderer for columns you want it enabled or disabled for.

ibdafna:
@fcollonval `yarn api` doesn't generate any file changes?

fcollonval:
> yarn api doesn't generate any file changes?

I don't know if you did [cacc6ac](https://github.com/jupyterlab/lumino/pull/451/commits/cacc6ac2c0097cc5ba538989b15e68ae9e60618f) manually then (thanks for that). But I think the remaining error was a typo in that commit. I fixed it. :crossed_fingers: for the CI

PR #458, opened 2022-11-01
Audit `DataGrid` public API in preparation for 0.x => 2.0

afshin:
In this PR:

- All public package versions are set to `2.0.0-alpha.6`
- Moves the standalone `Notification` class for `DataGrid` instances that use `CellEditor` into the latter's namespace: `CellEditor.Notification`


blink1073:
I believe from a releaser standpoint we need to set `version-cmd` [here](https://github.com/jupyterlab/lumino/blob/15fb5d14136b9d465d744c3373bd347fe1f40bf0/package.json#L61) to `npm run update:versions` and update the release instructions.

blink1073:
Need to pass a `-y` flag I think: `? Are you sure you want to create these versions? (ynH)`

blink1073:
It is `--yes`

fcollonval:
I'm not convinced we should go that road. It was not needed earlier and we also decide it was not needed for JupyterLab (although dependency cycles are still to be removed).

I'm in favor to bump all packages for v2. But going forward in v2 branch I don't see the benefit to use fixed version as most of the time is will bumped packages for no reasons.

RELEASE.md
```diff
@@ -4,8 +4,7 @@
 
 The recommended way to make a release is to use [`jupyter_releaser`](https://github.com/jupyter-server/jupyter_releaser#checklist-for-adoption).
 
-Because `lumino` uses independent versions, the versioning must be
-done manually as follows:
+`lumino` uses a fixed version for all packages. The version is defined in `lerna.json`. To update package versions, run:
```
fcollonval, line 7:
```suggestion
Because `lumino` uses independent versions, the versioning must be
```

.github/workflows/tests.yml
```diff
@@ -106,6 +106,23 @@ jobs:
...
+          with open("package.json") as f:
+            data = json.load(f)
+
+          data["version"] = datetime.now().strftime("%Y.%m.%d")
+          data["jupyter-releaser"]["options"]["version-cmd"] = "npm run update:versions -- --yes"
+          
+          with open("package.json", "w") as f:
+            data = json.dump(f)
+          
+
```
fcollonval, line 125:
```suggestion
```

vidartf:
The diff now contains a lot of changes to datagrid, which doesn't seem related to the title of this PR?

afshin:
> The diff now contains a lot of changes to datagrid, which doesn't seem related to the title of this PR?

Yeah, I should make two PRs, probably.

Basically the name `Notification` is a DOM concept `window.Notification` so before making `DataGrid` a post-1.0 version, that part of the API needs updating. This was meant to be an all-in-one version PR for all the packages, but since `@lumino/datagrid` requires some changes, maybe I should take all the version stuff out of this PR and just use this branch to update `DataGrid`.

afshin:
cc: @martinRenou @ibdafna

martinRenou (approved):
LGTM! Thank you

PR #462, opened 2022-11-08
Add async iterable `Stream` class that inherits from `Signal`

afshin:
This PR adds a `Stream` class that `extends Signal` and implements `IStream`, which means that in addition to being a signal, it is also an async iterable. It can be used as a drop-in replacement for any `Signal` (or `ISignal`) in an application if the author wishes to expose an async iterable API to complement signal connection/disconnection.

This PR also exposes a `protected blocked: number` property of a `Signal` so that sub-classes can implement correct blocking behavior.

### Subscribe to a stream via `connect`/`disconnect`:
```ts
// Instantiate a stream.
const stream = new Stream<unknown, string>({});

// Connect a handler to the stream.
const handler = (_, args) => console.log(`emission is: ${args}`);
stream.connect(handler);

stream.emit('foo');
// [OUTPUT] emission is: foo

stream.emit('bar');
// [OUTPUT] emission is: bar

stream.emit('baz');
// [OUTPUT] emission is: baz

// Disconnect the stream handler.
stream.disconnect(handler);
```

### Subscribe to a stream as an async iterable:
```ts
// Instantiate a stream.
const stream = new Stream<unknown, string>({});

// Iterate over stream as an iterable.
void (async () => {
  for await (const emission of stream) {
    console.log(`emission is: ${emission}`);
  }
  console.log('stream stopped');
})();

stream.emit('foo');
// [OUTPUT] emission is: foo

stream.emit('bar');
// [OUTPUT] emission is: bar

stream.emit('baz');
// [OUTPUT] emission is: baz

stream.stop();
// [OUTPUT] stream stopped
```

### Generate an async iterator over the stream.
```ts
// Instantiate a stream.
const stream = new Stream<unknown, string>({});

// Loop through the stream iterator.
void (async () => {
  const it = stream[Symbol.asyncIterator]();
  let emission: IteratorResult<string, any>;
  while (!(emission = await it.next()).done) {
    console.log(`emission is: ${emission.value}`);
  }
  console.log('stream stopped');
})();

stream.emit('foo');
// [OUTPUT] emission is: foo

stream.emit('bar');
// [OUTPUT] emission is: bar

stream.emit('baz');
// [OUTPUT] emission is: baz

stream.stop();
// [OUTPUT] stream stopped
```

### DIFF
--- a/packages/signaling/package.json
+++ b/packages/signaling/package.json
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@lumino/algorithm": "^2.0.0-alpha.6",
+    "@lumino/coreutils": "^2.0.0-alpha.6",
     "@lumino/properties": "^2.0.0-alpha.6"
   },
   "devDependencies": {
--- a/packages/signaling/src/index.ts
+++ b/packages/signaling/src/index.ts
@@ -86,6 +86,13 @@ export interface ISignal<T, U> {
 }
 
 /**
+ * An object that is both a signal and an async iterable iterator.
+ */
+export interface IStream<T, U>
+  extends ISignal<T, U>,
+    AsyncIterableIterator<U> {}
+
+/**
  * A concrete implementation of `ISignal`.
  *
  * #### Example
@@ -345,10 +352,7 @@ export namespace Signal {
 /**
  * A stream with the characteristics of a signal and an async iterator.
  */
-export class Stream<T, U>
-  extends Signal<T, U>
-  implements AsyncIterableIterator<U>
-{
+export class Stream<T, U> extends Signal<T, U> implements IStream<T, U> {
   /**
    * Return an async iterator that yields every emission.
    */
--- a/packages/signaling/src/index.ts
+++ b/packages/signaling/src/index.ts
@@ -355,7 +355,7 @@ export class Stream<T, U> extends Signal<T, U> implements IStream<T, U> {
    * Return an async iterator that yields every emission.
    */
   async *[Symbol.asyncIterator](): AsyncIterableIterator<U> {
-    let pending = this.pending;
+    let pending = this._pending;
     while (true) {
       try {
         const { args, next } = await pending.promise;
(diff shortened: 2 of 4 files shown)


afshin:
cc: @jasongrout @vidartf

jasongrout:
Cool. If I understand it correctly, you're kind of long-polling the signal?

afshin:
I wouldn't have used the term long polling, but maybe it is apt, because there is always a `pending` promise outstanding and it is resolved every time there is an emission. Is that what you mean?

fcollonval:
Thinking a bit more about event life cycle, could you try the following scenario:

A signal listener, emit the same signal that triggers it (only the first time it get triggers to avoid infinite recursion).

When are the iterator/iterable displayed?

> The goal here is to check a validation scenario for instance, in which one listener _correct_ a model it is listening. So that the same signal will be triggered twice in a synchronous callback stack.

My guess is that the iterator/iterable will not be displayed between the two executions of the listener as they are asynchronous. If this is the case, we should warn user of potential inconsistency when using that feature as the state that triggered a promise resolution may not exist any more.

afshin:
@fcollonval You're absolutely right. There is a problem here. I'm going to put this PR back into draft mode to think about it.

afshin:
@fcollonval I was thinking about it realized that maybe `.next()` is not necessary and that it is enough for `Stream` to be an async iterable. If you *truly* want an iterator, you can still call `Stream#[Symbol.asyncIterator]()`, but this is insufficient to fix the bug, *in se*.

The other thing I did was to make sure you *never* lose a promise by having the internal `pending` promise always resolve to the *next* promise. This way no promises get lost. It's the same promise-chaining technique `Poll` uses.

Thanks for finding it. It hurt my head for a while 🥲

fcollonval (approved):
Thanks @afshin

vidartf:
@afshin Right, so let's say we have widgets A and B. B listens to a stream on A via the new API. Then the owner of widget A figures out that it is no longer needed, and disposes of it. Will the infinite loop on B prevent the cleanup of A? Will it keep it around until B itself is cleaned up? Or will it be kept around forever?

afshin:
I think you're exposing the need for a `Stream#stop()` perhaps.

afshin:
@vidartf I added a `.stop()` method and updated the code snippets above to show how it can be used. I would say in the `dispose()` method of a class that has a stream, we should recommend that people stop the stream.

PR #464, opened 2022-11-12
When rejecting the internal promise in a `Stream`, catch failures

afshin:
This is a minor bug that might manifest in testing suites or in circumstances where a `Stream` instance is stopped when nobody is subscribed to its async iterator.


fcollonval (approved):
Thanks

PR #465, opened 2022-11-14
Improve the menubar accessibility

scmmmh:
This PR will improve accessibility issues with the menubar component, bringing it closer to being WCAG2.1 compliant. In particular it will
* Set the tabindex properties as specified;
* Handle the focus setting as specified, both when focus first moves into the menubar and also when focus moves out of it;

This is my first bit of work with the lumino codebase, so all comments, particularly around architecture are much appreciated.

I haven't added any tests yet, that is the next step.

### DIFF
--- a/examples/example-menubar/package.json
+++ b/examples/example-menubar/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@lumino/example-menubar",
-  "version": "0.9.0-alpha.6",
+  "version": "0.1.0-alpha.6",
   "private": true,
   "scripts": {
     "build": "tsc && rollup -c",
--- a/examples/example-menubar/package.json
+++ b/examples/example-menubar/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "@lumino/example-menubar",
+  "version": "0.9.0-alpha.6",
+  "private": true,
+  "scripts": {
+    "build": "tsc && rollup -c",
+    "clean": "rimraf build"
+  },
+  "dependencies": {
+    "@lumino/default-theme": "^1.0.0-alpha.6",
+    "@lumino/messaging": "^2.0.0-alpha.6",
+    "@lumino/widgets": "^2.0.0-alpha.6"
+  },
+  "devDependencies": {
+    "@rollup/plugin-node-resolve": "^13.3.0",
+    "rimraf": "^3.0.2",
+    "rollup": "^2.77.3",
+    "rollup-plugin-styles": "^4.0.0",
+    "typescript": "~4.7.3"
+  }
+}
--- a/examples/example-menubar/src/index.ts
+++ b/examples/example-menubar/src/index.ts
@@ -1,0 +1,104 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import { CommandRegistry } from '@lumino/commands';
+import { Menu, MenuBar, PanelLayout, Widget } from '@lumino/widgets';
+
+import '../style/index.css';
+
+class Application extends Widget {
+  constructor() {
+    super({ tag: 'main' });
+  }
+}
+
+class SkipLink extends Widget {
+  static createNode(): HTMLElement {
+    const node = document.createElement('a');
+    node.setAttribute('href', '#content');
+    node.innerHTML = 'Skip to the main content';
+    node.classList.add('lm-example-skip-link');
+    return node;
+  }
+
+  constructor() {
+    super({ node: SkipLink.createNode() });
+  }
+}
+
+class Article extends Widget {
+  static createNode(): HTMLElement {
+    const node = document.createElement('div');
+    node.setAttribute('id', 'content');
+    node.setAttribute('tabindex', '-1');
+    const h1 = document.createElement('h1');
+    h1.innerHTML = 'Menubar Example';
+    node.appendChild(h1);
+    const label = document.createElement('label');
+    label.appendChild(
+      document.createTextNode('A textarea to demonstrate the tab handling.')
+    );
+    const textarea = document.createElement('textarea');
+    textarea.setAttribute('autocomplete', 'off');
+    label.appendChild(textarea);
+    node.appendChild(label);
+    return node;
+  }
+
+  constructor() {
+    super({ node: Article.createNode() });
+  }
+}
+
+function addMenuItem(
+  commands: CommandRegistry,
+  menu: Menu,
+  command: string,
+  label: string,
+  log: string
+): void {
+  commands.addCommand(command, {
+    label: label,
+    execute: () => {
+      console.log(log);
+    }
+  });
+  menu.addItem({
+    type: 'command',
+    command: command
+  });
+}
+
+function main(): void {
+  const app = new Application();
+  const appLayout = new PanelLayout();
+  app.layout = appLayout;
+
+  const skipLink = new SkipLink();
+  appLayout.addWidget(skipLink);
+
+  const menubar = new MenuBar();
+  const commands = new CommandRegistry();
+
+  const fileMenu = new Menu({ commands: commands });
+  fileMenu.title.label = 'File';
+  addMenuItem(commands, fileMenu, 'new', 'New', 'File > New');
+  addMenuItem(commands, fileMenu, 'open', 'Open', 'File > Open');
+  addMenuItem(commands, fileMenu, 'save', 'Save', 'File > Save');
+  menubar.addMenu(fileMenu);
+
+  const editMenu = new Menu({ commands: commands });
+  editMenu.title.label = 'Edit';
+  addMenuItem(commands, editMenu, 'cut', 'Cut', 'Edit > Cut');
+  addMenuItem(commands, editMenu, 'copy', 'Copy', 'Edit > Copy');
+  addMenuItem(commands, editMenu, 'paste', 'Paste', 'Edit > Paste');
+  menubar.addMenu(editMenu);
+  appLayout.addWidget(menubar);
+
+  const article = new Article();
+  appLayout.addWidget(article);
+
+  Widget.attach(app, document.body);
+}
+
+window.onload = main;
--- a/examples/example-menubar/src/index.ts
+++ b/examples/example-menubar/src/index.ts
@@ -49,14 +49,9 @@ class Article extends Widget {
     const h1 = document.createElement('h1');
     h1.innerHTML = 'MenuBar Example';
     node.appendChild(h1);
-    const label = document.createElement('label');
-    label.appendChild(
-      document.createTextNode('A textarea to demonstrate the tab handling.')
-    );
-    const textarea = document.createElement('textarea');
-    textarea.setAttribute('autocomplete', 'off');
-    label.appendChild(textarea);
-    node.appendChild(label);
+    const button = document.createElement('button');
+    button.innerHTML = 'A button you can tab to out of the menubar';
+    node.appendChild(button);
     return node;
   }
 
@@ -96,7 +91,6 @@ function main(): void {
   app.layout = appLayout;
 
   const skipLink = new SkipLink();
-  appLayout.addWidget(skipLink);
 
   const menubar = new MenuBar();
   const commands = new CommandRegistry();
@@ -114,9 +108,11 @@ function main(): void {
   addMenuItem(commands, editMenu, 'copy', 'Copy', 'Edit > Copy');
   addMenuItem(commands, editMenu, 'paste', 'Paste', 'Edit > Paste');
   menubar.addMenu(editMenu);
-  appLayout.addWidget(menubar);
 
   const article = new Article();
+
+  appLayout.addWidget(skipLink);
+  appLayout.addWidget(menubar);
   appLayout.addWidget(article);
 
   Widget.attach(app, document.body);
--- a/examples/example-menubar/src/index.ts
+++ b/examples/example-menubar/src/index.ts
@@ -6,13 +6,22 @@ import { Menu, MenuBar, PanelLayout, Widget } from '@lumino/widgets';
 
 import '../style/index.css';
 
+/**
+ * Wrapper widget containing the example application.
+ */
 class Application extends Widget {
   constructor() {
     super({ tag: 'main' });
   }
 }
 
+/**
+ * Skip link to jump to the main content.
+ */
 class SkipLink extends Widget {
+  /**
+   * Create a HTMLElement that statically links to "#content".
+   */
   static createNode(): HTMLElement {
     const node = document.createElement('a');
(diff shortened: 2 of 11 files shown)


scmmmh:
I've added tests and the PR is now ready for review.

examples/example-menubar/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "version": "0.9.0-alpha.6",
+  "private": true,
+  "scripts": {
+    "build": "tsc && rollup -c",
+    "clean": "rimraf build"
+  },
+  "dependencies": {
+    "@lumino/default-theme": "^1.0.0-alpha.6",
+    "@lumino/messaging": "^2.0.0-alpha.6",
+    "@lumino/widgets": "^2.0.0-alpha.6"
```
gabalafou, line 12: I realized I'm actually quite confused about how this builds the example with your local modifications to the source code. I think it doesn't do so automatically, so in order to test this new example page with your source code changes, I think I have to execute the following commands:

```shell
cd lumino
gh pr checkout 465
cd packages/widgets
yarn run build
yarn link
cd ../../examples/example-menubar
yarn install
yarn link @lumino/widgets
yarn run build
open index.html
```

packages/widgets/src/menubar.ts
```diff
@@ -417,17 +423,20 @@ export class MenuBar extends Widget {
     // Fetch the key code for the event.
     let kc = event.keyCode;
 
-    // Do not trap the tab key.
+    // Reset the active index on tab (but not reverse tab), but do not trap the tab key.
     if (kc === 9) {
+      if (!event.shiftKey) {
+        this.activeIndex = -1;
```
gabalafou, line 429: Sorry, even with the comment I'm still not quite sure I get what this line is doing. 

I thought if it's not resetting the `activeIndex`, then I should be able to do the following on the example page:

1. `Tab` to the menu bar
2. Right arrow key over to Edit (it's highlighted now)
3. Click on the text area input ("Edit" is still highlighted)
4. Press `Shift` + `Tab` and...
    **Expected** behavior: "Edit" remains highlighted and if I press `Enter` it should open the Edit menu. 
    **Actual** behavior: the File menu items becomes highlighted and if I press `Enter` it opens the File menu.

packages/widgets/src/menubar.ts
```diff
@@ -591,6 +600,17 @@ export class MenuBar extends Widget {
     }
   }
 
+  /**
+   * Handle the `'focus'` event for the menu bar.
+   */
+  private _evtFocus(event: FocusEvent): void {
```
gabalafou, line 606: Why do we wish to reset the active menu item when the menu bar receives focus?

I would like to make the Lumino menubar behave more like the [example menubar in the ARIA Authoring Practices Guide](https://www.w3.org/WAI/ARIA/apg/example-index/menubar/menubar-navigation).

packages/widgets/tests/src/menubar.spec.ts
```diff
@@ -760,6 +780,45 @@ describe('@lumino/widgets', () => {
...
+          bar.dispose();
+        });
+
+        it('should loose focus on tab key', () => {
+          let bar = createUnfocusedMenuBar();
+          bar.contentNode.focus();
+          expect(
+            bar.contentNode.contains(document.activeElement) &&
+              bar.contentNode !== document.activeElement
+          ).to.equal(true);
```
gabalafou, line 801: I'm not sure this test actually tests that the menu bar loses focus. Perhaps the test should be renamed, or perhaps these 4 lines should be modified and moved below line 803 (the event dispatch).

```suggestion
          expect(bar.node.contains(document.activeElement)).toBe(false); 
```

remembering the odd property that a node.contains itself, so `document.activeElement.contains(document.activeElement)` will always be true I think.

packages/widgets/tests/src/menubar.spec.ts
```diff
@@ -760,6 +780,45 @@ describe('@lumino/widgets', () => {
...
+        it('should loose focus on shift-tab key', () => {
+          let bar = createUnfocusedMenuBar();
+          bar.contentNode.focus();
+          expect(
+            bar.contentNode.contains(document.activeElement) &&
+              bar.contentNode !== document.activeElement
+          ).to.equal(true);
+          let event = new KeyboardEvent('keydown', { keyCode: 9 });
+          bar.contentNode.dispatchEvent(event);
+          bar.contentNode.focus();
```
gabalafou, line 817: What should `bar.activeIndex` equal before this call to `.focus()`?

packages/widgets/tests/src/menubar.spec.ts
```diff
@@ -760,6 +780,45 @@ describe('@lumino/widgets', () => {
...
+        });
+
+        it('should loose focus on shift-tab key', () => {
+          let bar = createUnfocusedMenuBar();
+          bar.contentNode.focus();
+          expect(
+            bar.contentNode.contains(document.activeElement) &&
+              bar.contentNode !== document.activeElement
+          ).to.equal(true);
+          let event = new KeyboardEvent('keydown', { keyCode: 9 });
```
gabalafou, line 815: Did you forget the shift modifier?

gabalafou (changes requested):
I'm so stoked to see this PR, thanks @scmmmh!

Is there any way we could correct the asymmetry between tabbing forward versus backward (shift-tabbing)? Right now, what I'm seeing with your changes is that when I tab through the example page, it focuses on the skip link first, then the File menu item, then the text area. But starting from the text area and shift-tabbing backwards, I get: File menu item, then the entire menubar, then the skip link. Tabbing forward, the user does not see the menubar receive focus, but tabbing backwards they do. Does my description make sense? I can make a video if it's not clear, just lemme know! 😃 

I think this is a step in the right direction, but ultimately I want to make the Lumino menubar behave more like the [ARIA example menubar](https://www.w3.org/WAI/ARIA/apg/example-index/menubar/menubar-navigation). There are some differences between that implementation and the Lumino menubar after this PR, which I would like to correct, for example:

1. Pressing the escape key takes the user out of the menubar in Lumino, whereas in the ARIA example, it closes any open menus but keeps them in the menubar. (Tab key is how to get out.)
2. Pressing the tab key while inside an open Lumino menu does nothing, whereas in the ARIA example it navigates you out of the menubar.
3. When you leave the Lumino menubar it forgets the last menu item that you were on, whereas the ARIA example remembers.

And so on and so forth.

But I think this PR incrementally moves in a direction aligned with a future better UX, so I don't think it needs to be held back by any of those outstanding discrepancies.

The only reason I'm not approving it right now is because I have some questions about the code. I also cannot review the version numbering stuff in the package.json file. @afshin could you take a look at that or ping somebody who knows how to review it?

packages/widgets/tests/src/menubar.spec.ts
```diff
@@ -760,6 +780,45 @@ describe('@lumino/widgets', () => {
...
+            bar.contentNode.contains(document.activeElement) &&
+              bar.contentNode !== document.activeElement
+          ).to.equal(true);
+          let event = new KeyboardEvent('keydown', { keyCode: 9, bubbles });
+          bar.contentNode.dispatchEvent(event);
+          expect(bar.activeIndex).to.equal(-1);
+          bar.dispose();
+        });
+
+        it('should loose focus on shift-tab key', () => {
```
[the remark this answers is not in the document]
scmmmh: :laughing: No, this is just the one spelling mistake that I continue to make. I've always misspelled that and it looks like I always will.

examples/example-menubar/src/index.ts
```diff
@@ -0,0 +1,125 @@
...
+  static createNode(): HTMLElement {
+    const node = document.createElement('div');
+    node.setAttribute('id', 'content');
+    node.setAttribute('tabindex', '-1');
+    const h1 = document.createElement('h1');
+    h1.innerHTML = 'MenuBar Example';
+    node.appendChild(h1);
+    const label = document.createElement('label');
+    label.appendChild(
+      document.createTextNode('A textarea to demonstrate the tab handling.')
```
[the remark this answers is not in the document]
scmmmh: I'll add a comment. It is basically there so that there is some content that can be tabbed to. Otherwise the tab out of the menubar takes you into the browser UI and that can often be a bit harder to see where it has gone (I guess an a11y issue in itself).

examples/example-menubar/src/index.ts
```diff
@@ -0,0 +1,125 @@
...
+  editMenu.title.label = 'Edit';
+  addMenuItem(commands, editMenu, 'cut', 'Cut', 'Edit > Cut');
+  addMenuItem(commands, editMenu, 'copy', 'Copy', 'Edit > Copy');
+  addMenuItem(commands, editMenu, 'paste', 'Paste', 'Edit > Paste');
+  menubar.addMenu(editMenu);
+  appLayout.addWidget(menubar);
+
+  const article = new Article();
+  appLayout.addWidget(article);
+
```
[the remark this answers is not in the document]
scmmmh: No problem. Will change.

examples/example-menubar/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@lumino/example-menubar",
+  "version": "0.9.0-alpha.6",
```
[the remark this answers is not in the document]
scmmmh: Yes. Happy to change to whatever is desired.

scmmmh:
Thanks for all the detailed feedback. I'll iterate the code.

The main reason for the divergence from the menubar example is that I was following the spec, which says "when a menubar receives focus, keyboard focus is placed on the first item". Looking at the example and your comments, it has become clear to me that there is an ambiguity here, whether the "menubar" is the actual HTML element with the `role="menubar"` or the component as a whole. In the first interpretation the HTML element must have `tabindex="0"` to be focusable. As a result you have this slightly different behaviour tabbing forward and backward and also a lot of the code I have is to handle the resulting edge cases. In the second interpretation, none of that is necessary and only the menuitems are focusable. Since that is the interpretation used by the example, I'll rework the code to follow that one. It will make the code a good bit simpler, so that'll be nice.

I agree that the long-term aim is to make the menubar work exactly like the ARIA menubar and this is just a first, incremental step towards that. I just wanted to take some first steps in the codebase and get some feedback, before I start making bigger changes.

examples/example-menubar/package.json
scmmmh: I hadn't even thought about that. I just run

```
cd packages/widgets
yarn run build
cd ../../examples/example-menubar
yarn run build
```

and it works :-).

packages/widgets/src/menubar.ts
scmmmh: This is caused by the interpretation of what it means for the "menubar to receive focus", as I put in the main reply. I'll switch to the interpretation used in the example, which will also make the code much simpler.

packages/widgets/tests/src/menubar.spec.ts
scmmmh: Yes. I was fiddling with getting the tab behaviour to work in the test and in the process of that removed too much :-).

scmmmh:
If somebody could add a label, then that would reduce the number of failure e-mails I receive and would be much appreciated :-).

blink1073:
> If somebody could add a label, then that would reduce the number of failure e-mails I receive and would be much appreciated :-).

Done, and thanks for working on this!

scmmmh:
> Done, and thanks for working on this!

Happy to help. I'm responsible for accessibility in my department, so every issue I fix is a complaint that doesn't come to me :smiley:

packages/widgets/src/menubar.ts
```diff
@@ -727,6 +737,7 @@ export class MenuBar extends Widget {
   }
 
   private _activeIndex = -1;
+  private _focusIndex = 0;
```
gabalafou, line 740: I'm a little worried about adding another state variable to this component. Why do we need to allow the active index to diverge from the focus index?

packages/widgets/src/menubar.ts
```diff
@@ -417,17 +427,18 @@ export class MenuBar extends Widget {
     // Fetch the key code for the event.
     let kc = event.keyCode;
 
-    // Do not trap the tab key.
+    // Reset the active index on tab, but do not trap the tab key.
     if (kc === 9) {
+      this.activeIndex = -1;
```
gabalafou, line 432: This introduces an inconsistency that makes me uncomfortable. The active index gets reset when you tab out of the menubar but not when you click out. On the example page, if you tab to the File menu, then click on the button, the File menu stays shaded. But if you tab to the File menu, then tab to the button, the File menu loses its shading.

gabalafou (changes requested):
This is so great! I really love the direction this is going in!

When trying this out locally, I noticed that if I get rid of the focus index and also get rid of the line that resets the active index, I think I get the same behavior minus the inconsistency I mention in one of my inline comments. 

As such, I'm marking my review as "request changes"—not because I'm sure the code needs to change but because I have questions. Thank you for your patience with the review!

packages/widgets/src/menubar.ts
```diff
@@ -772,6 +783,11 @@ export namespace MenuBar {
      */
     readonly active: boolean;
 
+    /**
+     * The index of the item in the list of items.
```
gabalafou, line 787: I think this description was meant for a different variable? `focusable` is a boolean, not an index.

While we're at it, could we rename this variable? To be, focusable means "able to be focussed" but I think what this variable really means is more like "should be made focusable" (by setting `tabindex` to 0). 

What do we think about `makeFocusable` or `enableFocus` or `enableTabIndex` or...?

packages/widgets/src/menubar.ts
gabalafou: After discussing this over a call, I better understand the need for another state variable, but could you leave some comments explaining what these variables mean and why we need both of them?

gabalafou (approved):
I would be okay with merging this code as-is, but I left some inline comments with some suggested changes.

packages/widgets/src/menubar.ts
gabalafou: I can live with this inconsistency for now, but we will need to address it in the future.

scmmmh:
I've updated the code in line with your comments.

gabalafou (approved):

scmmmh:
Do I need to do anything to make the API changes test pass?

blink1073:
> Do I need to do anything to make the API changes test pass?

I believe you need to call `yarn run api` and then commit the changes.

fcollonval (approved):
Thanks a lot @scmmmh and thanks @gabalafou for the review

fcollonval:

scmmmh:
> Thanks a lot @scmmmh and thanks @gabalafou for the review

Happy to help. More soon, I hope.

Mark

Issue #472, opened 2022-11-23
Dock panels that have dock panels inside become very difficult/impossible to rearrange


drcd1:
## Description

When using DockPanels inside a DockPanel, with the 'tabsConstrained' option, it is very difficult to re-organize the top level's layout.
This issue is related to #43 . While #43 was solved by blocking invalid moves and drags, it doesn't show the "valid" possible moves. Below you can see an animation reproducing this issue.

![Animation](https://user-images.githubusercontent.com/22733172/203542122-f1faaa8a-25f7-45b1-bfca-8e43241eb37f.gif)

Here, the user tries to move panel 2 under panel 3. Most of the overlay positions are blocked, because the lower level dockpanels take priority. The user then is able to place panel 3 at the very very thin margins. This (correctly) does not yield the intended layout, and then the original configuration becomes impossible to recover, as the user is only able to place the panels at the margins: all other positions are blocked. 

## Reproduce

1. Create a ui interface with DockPanels inside DockPanels. Don't forget to set 'constrainedTabs' so that the panels
2. Attempt to drag one of the top level tabs to reorganize the top level layout.
3. Notice how this is almost impossible to do, as the 'drag' only sees the bottom layer, and (correctly) marks the move as invalid.

## Expected behavior
When dragging a top-level tab in a DockPanel, one would expect to see the overlays and act changes only for the top level, and not have their actions interpreted as if they wanted to place the tabs inside the inner DockPanel.

## Proposed solution

When handling drag events in the DockPanel, don't stop propagation if they're invalid.
I've tested this and it seems to work. 
I will make a pull request with these changes

drcd1:
Here's a modified version of one of the examples that illustrates the problem:
[example-dockpanel-amd.zip](https://github.com/jupyterlab/lumino/files/10075546/example-dockpanel-amd.zip)

fcollonval:
Closing as fixed by #473

PR #473, opened 2022-11-23
Fix drag-and-drop of nested dock panel

drcd1:
This pull request applies the proposed solution to issue #472, by allowing drag events to be propagated when the current potential docking position is invalid.

Here you can see an animation of the solved problem.
Animation2

### DIFF
--- a/examples/example-nested-dockpanel-amd/README.md
+++ b/examples/example-nested-dockpanel-amd/README.md
@@ -1,0 +1,21 @@
+# example-dockpanel-amd
+
+_AMD_ version of [example-dockpanel](../example-dockpanel)
+
+## Prerequisites
+
+From the root lumino folder run:
+
+```
+yarn install
+yarn run build
+yarn run minimize
+```
+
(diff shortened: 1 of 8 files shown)


fcollonval (commented):
Thanks a lot @drcd1 

Would you mind adding your example in the `examples` folder as part of this PR (you may call it _example-nested-dockpanel_)

examples/example-nested-dockpanel-amd/README.md
```diff
@@ -0,0 +1,21 @@
+# example-dockpanel-amd
+
+_AMD_ version of [example-dockpanel](../example-dockpanel)
```
fcollonval, line 3:
```suggestion
# example-nested-dockpanel-amd

Example of nested dock panel.
```

fcollonval (changes requested):
Thanks for adding the example.

I have a tiny change request and we are good to go for me.

fcollonval (approved):
Thanks @drcd1

fcollonval:

PR #489, opened 2022-12-06
Collapse main menu options in a hamburger menu

steff456:
This PR fixes https://github.com/jupyterlab/jupyterlab/issues/13148

This PR makes the main menu bar responsive by collapsing the menus that don't fit in a hamburger menu,

responsive-menu

The resize event works for the resize action of the window and zoom in/out keyboard shortcuts, etc.


packages/widgets/src/menubar.ts
```diff
@@ -359,6 +366,7 @@ export class MenuBar extends Widget {
     this.node.addEventListener('mousemove', this);
     this.node.addEventListener('mouseleave', this);
     this.node.addEventListener('contextmenu', this);
+    window.addEventListener('resize', this);
```
fcollonval, line 369: You need to remove the listener in `onAfterDetach`.

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +390,74 @@ export class MenuBar extends Widget {
...
+      }
+      if (totalMenuSize > screenSize) {
+        index = i;
+        break;
+      }
+    }
+    if (first) {
+      first = false;
+    }
+    console.log(this._menuItemSizes);
```
fcollonval, line 424: Reminder to be removed

```suggestion
```

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +390,74 @@ export class MenuBar extends Widget {
...
+        this._hamburgerMenu.insertItem(0, {
+          type: 'submenu',
+          submenu: submenu
+        });
+        this.removeMenuAt(i);
+      }
+    } else if (this._hamburgerMenu !== null) {
+      let i = n - 1;
+      let hamburgerMenuItems = this._hamburgerMenu.items;
+      console.log(screenSize - totalMenuSize, this._menuItemSizes[i]);
```
fcollonval, line 448:
```suggestion
```

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +390,74 @@ export class MenuBar extends Widget {
...
+    }
+    if (first) {
+      first = false;
+    }
+    console.log(this._menuItemSizes);
+    if (index > -1) {
+      // Create hamburger menu
+      const commands = new CommandRegistry();
+      if (this._hamburgerMenu === null) {
+        this._hamburgerMenu = new Menu({ commands });
```
fcollonval, line 429: You can solve the code scan warning using this:

```suggestion
        this._hamburgerMenu = new Menu({ commands: new CommandRegistry() });
```

fcollonval (changes requested):
Thanks @steff456 

I have two questions:
- Should we forbid the menu label to expand on more than one line (as in your demo)?
- Should we cache the menu item sizes?

steff456:
Hi @fcollonval, thanks for reviewing this PR so fast!

> Should we forbid the menu label to expand on more than one line (as in your demo)?

I don't have a strong preference on this, and I think that this will not be the case in jupyterlab because all the menus only have one word. I'm not sure if there's another use case where the menu item has more than one word and if you prefer to give precedence to the actual menu item or the hamburger menu.

> Should we cache the menu item sizes?

I'm already saving in cache the menu item sizes to avoid recomputing them each time a `resize` event happens. This also helps me calculate the amount of free space on the right compared to the "original" size of the menu to take it out from the hamburger menu. If I don't save them I would not be able to do that size comparison.

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +391,71 @@ export class MenuBar extends Widget {
...
+        this._menuItemSizes.push(item.offsetWidth);
+      }
+      if (totalMenuSize > screenSize) {
+        index = i;
+        break;
+      }
+    }
+    if (first) {
+      first = false;
+    }
```
fcollonval, line 424: The code scan is flagging that because it is not used later. I guess you can simply drop it:

```suggestion
```

fcollonval (changes requested):
Thanks for the answer @steff456 

For the styling, your answer makes a lot of sense. And I agree that the best is to let downstream application to decide what they want to do.

For the caching, the code requires some additional changes. In addition to the code comment, could you also invalidate the cache in the method `onUpdateRequest` as the items list may change.

gabalafou (changes requested):
So happy to see this! 

I'm requesting changes because I think there's an error with the variable named `found` (please refer to the in-line code comments).

I also think there's some error in the logic that puts menu items into the hamburger and pulls them out because when I tested this locally, it was pretty easy for me to put the menu bar in the dock panel example into an invalid state, in which the "..." menu item appears somewhere in the middle of the menu bar instead of at the end, as shown in the following screen shot:

image

I think—but I'm not sure—that what's incorrect about this implementation is... well, I think I see a lack of symmetry between putting menus into the overflow/hamburger versus pulling them out. I realized that part of that asymmetry could be explained by the fact that the code might be relying on the fact that the resize event fires multiple times as the user drags the mouse to make the window smaller or bigger. However, I think it's bad practice to depend on that.

fcollonval:
> Thanks for the suggestions @fcollonval, please let me know how you see the implementation now and I wanted to ask you how I should create tests for this feature.

Thanks for the changes @steff456 
Regarding the tests, you can get inspiration from that test:

https://github.com/jupyterlab/lumino/blob/d51b0575c7797769af5a6aabed39500bd494a918/packages/widgets/tests/src/menubar.spec.ts#L894-L903

I guess that if you force the `width` style of `bar.node` and then requesting an update. You should be able to test whether all or some items are displayed depending on the width.

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +391,74 @@ export class MenuBar extends Widget {
     }
   }
 
+  /**
+   * A message handler invoked on an `'resize'` message.
+   */
+  protected _evtResize(event: Event): void {
+    let itemMenus = this.node.getElementsByClassName('lm-MenuBar-item');
```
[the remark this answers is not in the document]
steff456: I can add the comment, this instruction will get me the actual menu items that are displayed in the screen, so the menu items **inside** the overflow menu will not count.

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +391,74 @@ export class MenuBar extends Widget {
...
+
+    if (this._menuItemSizes.length == 0) {
+      // Check if it is the first resize and get info about menu items sizes
+      for (let i = 0; i < n; i++) {
+        let item = itemMenus[i] as HTMLLIElement;
+        let found = false;
+        // Add sizes to array
+        totalMenuSize += item.offsetWidth;
+        this._menuItemSizes.push(item.offsetWidth);
+        if (totalMenuSize > screenSize && !found) {
```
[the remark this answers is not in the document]
steff456: I think the assertion you suggested is also wrong, it will never enter the block code given that the index value is initialized in `-1` making that assertion `false` always.

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +391,74 @@ export class MenuBar extends Widget {
...
+        }
+      }
+    }
+
+    if (index > -1) {
+      // Create hamburger menu
+      if (this._hamburgerMenu === null) {
+        this._hamburgerMenu = new Menu({ commands: new CommandRegistry() });
+        this._hamburgerMenu.title.label = '...';
+        this._hamburgerMenu.title.mnemonic = 0;
```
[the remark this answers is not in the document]
steff456: I'm not sure exactly what `mnemonic` is but in other places in the code it is defined like this when creating a menu so I just copied it.

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +391,74 @@ export class MenuBar extends Widget {
     }
   }
 
+  /**
+   * A message handler invoked on an `'resize'` message.
+   */
+  protected _evtResize(event: Event): void {
```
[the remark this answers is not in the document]
steff456: No, there's not a resize event on the initial page load, so users will first have to resize their window. I'm not sure if there's a signal that exists when the page is loaded that I can connect this method to. If you know of it please let me know.

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +391,74 @@ export class MenuBar extends Widget {
...
+        this._hamburgerMenu.title.label = '...';
+        this._hamburgerMenu.title.mnemonic = 0;
+        this.addMenu(this._hamburgerMenu);
+      }
+
+      // Move menus
+      for (let i = index; i < n - 1; i++) {
+        let submenu = this.menus[i];
+        submenu.title.mnemonic = 0;
+        this._hamburgerMenu.insertItem(0, {
```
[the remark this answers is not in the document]
steff456: No, the functionality that I envisioned is that the order of the overflow menu is for the menu items that are left to right to gather top to bottom. Then, it is easier for me to preserve the order when unpacking the elements to the top menu again.

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +391,74 @@ export class MenuBar extends Widget {
...
+          type: 'submenu',
+          submenu: submenu
+        });
+        this.removeMenuAt(i);
+      }
+    } else if (this._hamburgerMenu !== null) {
```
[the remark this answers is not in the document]
steff456: It is an `else-if` block because I don't want to unpack all the elements of the overflow menu if there's no space for them, and obviously I don't want to try to do it if the overflow menu doesn't exist in the first place.

One of my first implementations tried to unpack everything and then just calculate which elements needed to go to the overflow menu but it had a lot of issues when testing it because the top menu was rendering faster, so sometimes menus will get lost or they will be inserted in the wrong order. That's why I decided to just unpack one element at the time per resize event. At the end even if the user drags the window continuously, that's firing multiple resize events.

packages/widgets/src/menubar.ts
```diff
@@ -382,6 +391,74 @@ export class MenuBar extends Widget {
...
+   * A message handler invoked on an `'resize'` message.
+   */
+  protected _evtResize(event: Event): void {
+    let itemMenus = this.node.getElementsByClassName('lm-MenuBar-item');
+    let screenSize = this.node.offsetWidth;
+    let totalMenuSize = 0;
+    let index = -1;
+    let n = itemMenus.length;
+
+    if (this._menuItemSizes.length == 0) {
```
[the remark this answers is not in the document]
steff456: I can move this block of code to another function but I would need to send almost all the variables defined in the function as a parameter, so I don't know what's best...

packages/widgets/src/menubar.ts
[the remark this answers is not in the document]
gabalafou: @afshin @fcollonval is there a good signal for initial load?

This seems bad from an end user's perspective, doesn't it? For example, if I've set my browser to a higher zoom and then load the JupyterLab UI, as an end user I don't want to have to jiggle the window resize handle to get the menu bar to render properly.

That said, just to be clear I'm not asking us to block on this, because it's something that can be added later.

packages/widgets/src/menubar.ts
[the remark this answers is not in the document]
gabalafou: It might just be me, but would it be better if the entire function were moved into a separate method? Like so:

```
_evtResize(event: Event): void {
  this._handleOverflow();
}
```

I'm not sure that's the best name for the method, but there are a few reasons why I think having such a method would be a good idea:

- I think it might be prudent to provide a way for clients to opt out of this behavior. I don't know all of the places that the Lumino menu bar class is used. I'm not 100% sure this behavior should be forced on downstream consumers always. Pulling the body of this function into its own method will make it easy to isolate that code and exit early if the client has passed some kind of "no overflow" option to the menu bar.
- Once we figure out how to hook into page load and where, we will be able to call this method (rather than `this._evtResize`) from there.
- Slightly simplified unit testing, since you don't have to trigger a resize event (also it makes it clear that you're not actually using the event object)
- Friendlier to future programmers if they want to add more stuff to the resize handler (that doesn't have to do with overflow)

packages/widgets/src/menubar.ts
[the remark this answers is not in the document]
gabalafou: Oh I see now why the code doesn't behave the way I thought it would. The resize event fires so many times when resizing the window that it's hard to actually have this loop handle more than one item at a time and when the user shrinks the window, it's like moving backwards through the menu items (because as the window shrinks the menu items get popped off from right to left).

gabalafou (changes requested):
Do we want to provide a "no resize" option?

Some of my comments here might be obsolete if you're currently working on trying to fix the issue with the bug I found in which the ellipsis appears somewhere in the middle of the menu bar rather than at the end, but I wanted to add them because they might impact the direction in which you go with the code.

Also, just a heads up, there are still some parts of this code (variables and comments) that refer to "hamburger" even though you renamed the class member. Could you update those please? 😊

steff456:
Hi @gabalafou, I had a meeting earlier this week with @afshin and I'm working in the complete refactoring of this PR to fix some of the issues that you flagged and others. I will ping you once that is ready for review 😊

steff456:
@gabalafou, @fcollonval, @afshin this PR is ready for another round of review 😬

packages/widgets/src/menubar.ts
```diff
@@ -403,8 +457,72 @@ export class MenuBar extends Widget {
...
+      // Calculate size of current menu
+      totalMenuSize += this._menuItemSizes[i];
+    }
+    // Render overflow menu if needed
+    if (this._overflowIndex > -1) {
+      // Create overflow menu
+      if (this._overflowMenu === null) {
+        this._overflowMenu = new Menu({ commands: new CommandRegistry() });
+        this._overflowMenu.title.label = '...';
+        this._overflowMenu.title.mnemonic = 0;
```
krassowski, line 469: Maybe worth adding a new optional method in `IRenderer`, e.g. `createOverflowMenu()`  that would allow to customise this? In particular downstream may want to customize label. You could provide a default implementation in `Renderer`.

packages/widgets/src/menubar.ts
```diff
@@ -403,8 +457,72 @@ export class MenuBar extends Widget {
...
+        let index = menus.length - 1 - i;
+        if (screenSize - totalMenuSize > this._menuItemSizes[index]) {
+          let menu = overflowMenuItems[0].submenu as Menu;
+          this._overflowMenu.removeItemAt(0);
+          this.insertMenu(length, menu, false);
+          let title = menu.title;
+          let active = false;
+          content[length] = renderer.renderItem({
+            title,
+            active,
```
krassowski, line 509: Let's fold it to reduce mental burden.

```suggestion
          content[length] = renderer.renderItem({
            title: menu.title,
            active: false,
```

packages/widgets/src/menubar.ts
```diff
@@ -403,8 +457,72 @@ export class MenuBar extends Widget {
...
+        this.removeMenu(submenu, false);
+      }
+      let title = this._overflowMenu.title;
+      let active = length === activeIndex;
+      if (active && menus[length].items.length == 0) {
+        active = false;
+      }
+      content[length] = renderer.renderItem({
+        title,
+        active,
```
krassowski, line 489:
```suggestion
      content[length] = renderer.renderItem({
        title: this._overflowMenu.title,
        active: length === activeIndex && menus[length].items.length !== 0
```

packages/widgets/src/menubar.ts
```diff
@@ -403,8 +457,72 @@ export class MenuBar extends Widget {
...
+      // Create overflow menu
+      if (this._overflowMenu === null) {
+        this._overflowMenu = new Menu({ commands: new CommandRegistry() });
+        this._overflowMenu.title.label = '...';
+        this._overflowMenu.title.mnemonic = 0;
+        this.addMenu(this._overflowMenu, false);
+      }
+      // Move menus to overflow menu
+      for (let i = menus.length - 2; i >= length; i--) {
+        let submenu = this.menus[i];
```
krassowski, line 474:
```suggestion
        const submenu = this.menus[i];
```

krassowski (changes requested):
This is awesome, thank you very much. This is also relatively complex logic with a lot of potential for off-by-one errors so we need good unit test coverage for this one.

packages/widgets/src/menubar.ts
```diff
@@ -479,9 +460,72 @@ export class MenuBar extends Widget {
...
+      if (this._overflowMenu.items.length === 0) {
+        this.removeMenu(this._overflowMenu, false);
+        content.pop();
+        this._overflowMenu = null;
+        this._overflowIndex = -1;
+      }
     }
-    this._menuItemSizes = [];
     VirtualDOM.render(content, this.contentNode);
+    this.updateOverflowIndex();
```
krassowski, line 525: It looks like this one will lead to the state being always delayed by one tick. It is not bad in itself if this is a conscious decision made as a performance-UX tradeoff. Could we document this?

One thing to consider is whether we should wrap it in an extra `requestAnimationFrame` to protect against layout trashing.

packages/widgets/src/menubar.ts
steff456: I get lint errors when I do it this way... not sure why

packages/widgets/src/menubar.ts
steff456: How do I wrap it in a `requestAnimationFrame`?

packages/widgets/src/menubar.ts
steff456: Same as above 😅

packages/widgets/src/menubar.ts
krassowski: What is the error?

steff456:
```
Argument of type '{ this: any; active: boolean; onfocus: () => void; }' is not assignable to parameter of type 'IRenderData'.
  Object literal may only specify known properties, and 'this' does not exist in type 'IRenderData'.ts(2345)
```

packages/widgets/src/menubar.ts
```diff
@@ -382,15 +390,61 @@ export class MenuBar extends Widget {
...
   }
 
+  /**
+   * A message handler invoked on a `'resize'` message.
+   */
+  protected onResize(msg: Widget.ResizeMessage): void {
+    this.update();
+  }
+
+  protected updateOverflowIndex(): void {
```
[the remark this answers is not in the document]
steff456: I tried to create the interface, but soon enough I found myself wondering that I needed to update the overflow index value in multiple spots, so I'm not sure if we need the interface. For now I just converted the function to private.

packages/widgets/src/menubar.ts
krassowski: I see that 4e8c6253f1df8b5b5cf7ca10d699af082673dd99 added `renderOverflowItem` which is a copy-paste of `renderItem` and is used to replace it; I think that we should . This might be useful (though I would avoid copy-pasting the code and instead just invoke the other method by default), but this is not what I had in mind.

I meant to suggest that we should wrap rendering of the overflow menu so that the `...`. and mnemonic can be customized downstream. The lines that I commented on would then become for example:

```suggestion
        this.renderer.decorateOverflowMenu(this._overflowMenu.title)
```

This is very minor, I don't think we have to do this. To limit the surface of API changes I would also suggest reverting `renderOverflowItem` for now - if this is needed in the future we can always add it (removing things is more difficult).

packages/widgets/src/menubar.ts
krassowski: I do not see this error locally. Should we see if it is reproducible on CI?

packages/widgets/src/menubar.ts
krassowski: It would be something like:

```js
requestAnimationFrame(() => this._updateOverflowIndex());
```

I have not tested if this helps. You would need to record a profile in developer tools when you resize the page to trigger collapsing before and after applying the change and compare. If you have the profiles it should be obvious from the time spend on forced style recalculation/layout whether it is worth waiting for the next frame or not.

packages/widgets/src/menubar.ts
steff456: let's try it!

afshin:
**This should be its own issue unless you have seen it and think there might be a fix you can apply here.**

It seems we have an outstanding bug in the `MenuBar` that does *not* come from this PR and happens with `mousemove` events. That was a red herring that I thought I could track down to this one before I tried just the upstream version and realized we have a problem (that does not need to be fixed here, but does need to be resolved before Lumino 2 final).

Chrome version of bug:
Screenshot 2023-01-27 at 19 07 27

Firefox version of bug:
Screenshot 2023-01-27 at 20 13 09

afshin:
Here is a weird artifact that *does* come from this PR, though. It happens when you first shrink and then re-expand the window. You can end up with two overflow menus. Here it is in Firefox:

https://user-images.githubusercontent.com/159529/215189022-72ec12fa-9bef-4e0a-b1bf-c2f366eb90e3.mov

afshin:
I think this PR is very close. The new logic flow is more in line with Lumino's widget lifecycle and easier to follow. Thanks for all the hard work @steff456! We can get this over the line.

steff456:
@afshin I already fixed the merging conflicts and added a test to this PR. Apparently the current failures involve another test, I'm not sure if re-running them will fix the issue. 

Please review this PR again and let me know if anything else is needed!

packages/widgets/src/menubar.ts
```diff
@@ -387,6 +413,13 @@ export class MenuBar extends Widget {
     }
   }
 
+  /**
+   * A message handler invoked on a `'resize'` message.
+   */
+  protected onResize(msg: Widget.ResizeMessage): void {
+    this.update();
```
fcollonval, line 420: You should propagate the message.

```suggestion
    this.update();
    super.onResize(msg);
```

fcollonval:
@steff456 the error seems legit but I don't know what is causing it as your changes from a quick glance seem unrelated.

steff456:
I'm not sure why that error is popping up either 😅. I'll give it another look because I'm not having that issue locally, just in the CI.

steff456:
As tests run in parallel, apparently having two animation frames in the same test I added was creating the weird failure. I split the test into two and now tests are working 🎉🎉🎉.

Let me know if there's anything else this PR need @fcollonval @afshin 

Thanks a lot!

fcollonval (approved):
Thanks @steff456 for the patience - let's merge this to be part of Lumino 2 RC.

krassowski:
@steff456 @gabalafou is there anything else needed on the JupyterLab side to make it work? When resizing the window currently I do not see the menu collapse as on the screencast nor is it seen in https://github.com/jupyterlab/jupyterlab/pull/14766. Was there a specific released version where it worked or was it not integration tested after merging this PR?

packages/widgets/src/menubar.ts
```diff
@@ -188,8 +208,8 @@ export class MenuBar extends Widget {
    * #### Notes
    * If the menu is already added to the menu bar, it will be moved.
    */
-  addMenu(menu: Menu): void {
-    this.insertMenu(this._menus.length, menu);
+  addMenu(menu: Menu, update: boolean = true): void {
```
steff456, line 211: Hi @krassowski, I remembered that we added this optional argument here in Lumino just in case other applications didn't want to have the collapse behavior. I can check if this argument is set to True in Jupyter, and that should be the solution of not seeing this feature active

Issue #496, opened 2022-12-22
Performance for consecutive tab switching improvements


krassowski:
### Problem

Some actions in the UI are expensive. When user navigates open tabs they may want to switch a few tabs ahead using a keyboard shortcut. Even with relatively low latency with windowed notebooks for any single tab, switch *n* tabs with large notebooks in a row may take too long for the user.

Related to https://github.com/jupyterlab/jupyterlab/issues/4292 and https://github.com/jupyterlab/benchmarks/issues/75

### Proposed Solution

We could:

1. aggregate consecutive switch tab requests if they appear within a debounce period and then switch to the final tab immediately
   - this means that for user who only switches one tab at a time the action will be slower (the time to switch will increase by the length of by the debounce period)
2. delay the expensive operation of rendering the tab beyond the debounce period; in particular we could "fake it" by displaying a static screenshot of the tab in the state as it was when it was most recently rendered
    - the screenshot of visible portion of each tab could be generated when it gets hidden (and possibly compressed in a webworker on idle)
    - this implies a memory usage increase (as we are storing an extra PNG screenshot for each hidden tab)
    - unfortunately making screenshots of DOM nodes is hard;
       - it is easy for datagrid as it is just a canvas and we can read it off the buffer
       - for non-canvas elements we basically need to render to canvas, for example:
          - [html2canvas](https://github.com/niklasvh/html2canvas) - MIT license, 47 kB, 4.5k forks
          - [rasterizeHTML](https://github.com/cburgmer/rasterizeHTML.js) - MIT license, 14.0 kB, ~200 forks
3. use `content-visibility` on tab-level to improve performance of tab switch. While previous work in jupyterlab-benchmarks repository investigated using content-visibility on cell level, this could be even more beneficial on widget level (in the same manner as https://github.com/jupyterlab/lumino/pull/231)
4. render a single main area widget in a panel in an iframe and use browser navigation for instantenious switching leveraging back-forward cache (https://web.dev/bfcache/)
5. <s>previously I [suggested](https://github.com/jupyterlab/jupyterlab/issues/4292#issuecomment-1270639448) using [Houdini](https://wiki.mozilla.org/CSS/Houdini) CSS Layout API but I no longer believe that this particular API exposes enough data to implement the caching mechanism that we need</s>

Additional comments:
- (1) I could not find any examples of applications using this approach
[...]

PR #497, opened 2022-12-23
Add content-visibility hide mode

krassowski:
See https://github.com/jupyterlab/lumino/issues/496#issuecomment-1363431726.

Once support for content-visibility is better across browsers we should make it the default as we have exactly the use case that it was intended for (see point two in [specification](https://www.w3.org/TR/css-contain-2/#using-cv-hidden)).

Performance-wise, I re-run the benchmark for three tabs (as in the issue: two test notebooks and launcher) after full integration and it works as expected:

Google Chrome 108

| hidden mode |IQM [Q1 - Q3] | mean | min |
|----|---|---|---|
| display | 3043.3 [2944.3 – 3193.1] ms | 3051.3 ms | 2390.9 ms
| content-visibility | 631.6 [606.4 – 694.6] ms | 687.7 ms | 566.5 ms
| scale | 548.6 [531 – 561.3] ms | 566.2 ms | 506.6 ms |

but differently to `scale`, it does not cause unintended issues in Chromium browsers where performance issues in presence of big notebooks are most pronounced, nor does it degrade in performance once there are too many tabs open.

Note: the benchmarks above were run with latest dev versions of JupyterLab (4.0alpha) and lumino from this PR, so these account for any potential changes to due windowed notebooks feature (full windowing mode).

Notes on current state in Firefox below. TLDR, we are good to merge IMO.

<details>

For anyone trying this out in Firefox: it may appear that it already "works" in Firefox 108 but it does not support `content-visibility` yet (although some basic support was already [merged](https://bugzilla.mozilla.org/show_bug.cgi?id=1758490) in 3 months ago, possibly pending bug fixes [like this one](https://bugzilla.mozilla.org/show_bug.cgi?id=1804761)). It "works" because in absence of `content-visibility` support it reduces to leaving the tab visible but hidden under the others due to z-index change. It means that we can confidently ship because it is in a way safer than existing _scale_ mode (if user chooses to enable it in a browser which does not support `content-visibility`, their workspace will still work; scale breaks things in Chromium browsers). The performance in this case is fine for few tabs (the three test notebooks scenario):

| hidden mode |IQM [Q1 - Q3] | mean | min |
|----|---|---|---|
| display | 1288.7 [1222.8 – 1351.8] ms | 1304.6 ms | 1132.1 ms |
| z-index | 960.6 [928.8 – 1004.9] ms | 1019.5 ms | 871.2 ms |
| scale | 925 [896.3 – 947.9] ms | 933.2 ms | 811.2 ms |

But it is worse when there are many notebooks (each 20 code cells):

| hidden mode | 10 notebooks | 30 notebooks |
|---|---|---|
| display | 1982.4 [1833.4 – 2145.6] |12061.3 [11591.2 – 12455.9] |
| z-index | 2517.6 [2435.2 – 2603.1] |21574.9 [21158.4 – 22010.4] |
| scale | 2007.3 [1912 – 2113.4] |17439.5 [16916 – 18083.8] |

Interestingly, scale in Firefox is only better for heavy notebooks, not for many small notebooks.

</details>

### DIFF
--- a/packages/widgets/src/docklayout.ts
+++ b/packages/widgets/src/docklayout.ts
@@ -839,18 +839,21 @@ export class DockLayout extends Layout {
 
     // Using transform create an additional layer in the pixel pipeline
     // to limit the number of layer, it is set only if there is more than one widget.
-    if (
-      this._hiddenMode === Widget.HiddenMode.Scale &&
-      refNode.tabBar.titles.length > 0
-    ) {
-      if (refNode.tabBar.titles.length == 1) {
+    if (this._hiddenMode === Widget.HiddenMode.Scale) {
+      if (refNode.tabBar.titles.length === 0) {
+        // Singular tab should use display mode to limit number of layers.
+        widget.hiddenMode = Widget.HiddenMode.Display;
+      } else if (refNode.tabBar.titles.length == 1) {
+        // If we are adding a second tab, switch the existing tab back to scale.
         const existingWidget = refNode.tabBar.titles[0].owner;
         existingWidget.hiddenMode = Widget.HiddenMode.Scale;
+      } else {
+        // For the third and subsequent tabs no special action is needed.
+        widget.hiddenMode = Widget.HiddenMode.Scale;
       }
-
-      widget.hiddenMode = Widget.HiddenMode.Scale;
     } else {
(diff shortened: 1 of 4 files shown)


fcollonval (approved):
Thanks a lot @krassowski

fcollonval:

PR #501, opened 2022-12-24
Only add aria-hidden for `scale` mode

### LINKED ISSUE #499: Widgets should not add/remove `aria-hidden` in `display` (and maybe `contentVisibility`) hidden mode
`aria-hidden` is not needed (hence should not be added) if element is hidden by standard mechanisms like `display: none`. It is only needed for non-standard means (such as `opacity`, `scale` or `translation` CSS).

MDN says:

> aria-hidden="true" should not be added when:
> 
> - The HTML [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute is present
> - The element or the element's ancestor is hidden with [display: none](https://developer.mozilla.org/en-US/docs/Web/CSS/display)
> - The element or the element's ancestor is hidden with [visibility: hidden](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility)

The behaviour for `content-visibility` is a bit more complex, see [Accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility#accessibility_concerns) section on MDN.

We should keep DOM changes to absolute minimum as every property changed forces style recalculation, affecting performance (and the less we change the more fast paths browsers can use).

krassowski:
Fixes #499. I would appreciate a review from someone working on accessibility.

### DIFF
--- a/packages/widgets/src/widget.ts
+++ b/packages/widgets/src/widget.ts
@@ -423,7 +423,6 @@ export class Widget implements IMessageHandler, IObservableDisposable {
       MessageLoop.sendMessage(this, Widget.Msg.BeforeShow);
     }
     this.clearFlag(Widget.Flag.IsHidden);
-    this.node.removeAttribute('aria-hidden');
     this._toggleHidden(false);
 
     if (this.isAttached && (!this.parent || this.parent.isVisible)) {
@@ -451,7 +450,6 @@ export class Widget implements IMessageHandler, IObservableDisposable {
       MessageLoop.sendMessage(this, Widget.Msg.BeforeHide);
     }
     this.setFlag(Widget.Flag.IsHidden);
-    this.node.setAttribute('aria-hidden', 'true');
     this._toggleHidden(true);
 
     if (this.isAttached && (!this.parent || this.parent.isVisible)) {
@@ -743,6 +741,7 @@ export class Widget implements IMessageHandler, IObservableDisposable {
           break;
         case Widget.HiddenMode.Scale:
           this.node.style.transform = 'scale(0)';
+          this.node.setAttribute('aria-hidden', 'true');
           break;
         case Widget.HiddenMode.ContentVisibility:
           this.node.style.contentVisibility = 'hidden';
@@ -756,6 +755,7 @@ export class Widget implements IMessageHandler, IObservableDisposable {
           break;
         case Widget.HiddenMode.Scale:
           this.node.style.transform = '';
+          this.node.removeAttribute('aria-hidden');
           break;
         case Widget.HiddenMode.ContentVisibility:
           this.node.style.contentVisibility = '';


fcollonval (commented):
Gentle ping to @gabalafou could you please review this one?

gabalafou (commented):
Hi @krassowski, thanks for working on this!

I'm not sure I entirely agree with MDN [that aria-hidden should not be added when an element has the `hidden` attribute, or `display: none`, or `visibility: hidden`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden). And the reason is precisely because of `content-visibility`. As far as I understand, if you style some offscreen node with `content-visibility: auto`, but inside that node you have a node that is hidden (via `display: none` or `visibility: hidden`), then the hidden node will still appear in the accessibility tree, which is most likely undesired. For example:

```html
<style>
.offscreen {
  content-visibility: auto;
}
.hidden {
  display: none;
}
</style>
<div>Above the fold</div>
<div class="offscreen">
  Below the fold.
  <div class="hidden">Hide this content.</div>
</div>
```

When the above page first loads and the user hasn't scrolled down, then "Hide this content" will appear in the accessibility tree, correct?

Why does the browser have to do a style recalculation when changing the aria-hidden value? Is it because the aria-hidden value can be targeted by CSS via the `[aria-hidden]` attribute selector? 

My other question is, since we're already changing properties on the widget's node when we call show/hide, why does changing aria-hidden at the same time cause additional work? Is it because the browser has to do the style recalculation at the exact moment that we call set/removeAttribute? Is there any way to batch together the node property changes?

One last question, how could I test this across browsers? Just so I understand what the desired behavior is, whenever we have a widget, no matter what `HiddenMode` it uses, when we call hide on that widget, its node should be removed from the accessibility tree, and when we call show, it should be added to the accessibility tree? Is that right?

Sorry if some of the above doesn't make sense. I had to type it up hastily. Again, thanks for working on this and please let me know how I can help. Happy to write tests or check this functionality on other browsers. Just let me know, thanks! 😄

krassowski:
We are not using `content-visibility: auto` though, only `content-visibility: hidden`.

> Why does the browser have to do a style recalculation when changing the aria-hidden value? Is it because the aria-hidden value can be targeted by CSS via the `[aria-hidden]` attribute selector?

Yes. Some browsers (see Firefox) have very good heuristics for avoiding doing so when not needed, others don't.

> My other question is, since we're already changing properties on the widget's node when we call show/hide, why does changing aria-hidden at the same time cause additional work?

Because browsers have fast paths (optimized code) to handle changes of some attributes or scenarios but not for others. We want to make as little DOM changes which invalidate the style cache as possible to trigger as many fast paths as possible.

>  Is it because the browser has to do the style recalculation at the exact moment that we call set/removeAttribute? Is there any way to batch together the node property changes?

No. Yes, browsers do batch the style recalculation and layout operations together automatically.

In general if you want to learn more see [this course](https://james-priest.github.io/udacity-nanodegree-mws/course-notes/browser-rendering-optimization.html).

> One last question, how could I test this across browsers? Just so I understand what the desired behavior is, whenever we have a widget, no matter what `HiddenMode` it uses, when we call hide on that widget, its node should be removed from the accessibility tree, and when we call show, it should be added to the accessibility tree? Is that right?

Yes, this is right. You would need to modify the DOM manually to test it at the moment (until we release).

fcollonval (approved):
Thanks @krassowski 

> I let you merge if you think it is ok.

Issue #504, opened 2022-12-27
Containers with managed size and position should use `contain:strict`


krassowski:
## Description

Containers in some layouts (e.g. panels) have fully managed size and position but are slow to work with.

## Reproduce

1. Open `gh-9757-reproducer.ipynb` and `all-html-elements.ipynb` notebooks as follows in a Chromium browser;
  ![Screenshot from 2022-12-27 12-46-53](https://user-images.githubusercontent.com/5832902/209668779-5973b906-f3ba-4a67-9b37-1ae1a7ab2aac.png)
3. Try to resize split panel in top-bottom
4. See jitter

| Without contain strict (current) | with `contain:strict` (proposed) |
|----|---|
| ![without-contain-strict](https://user-images.githubusercontent.com/5832902/209669330-9119d3f2-0d49-48e3-92b1-8225539e1087.gif) | ![with-contain-strict](https://user-images.githubusercontent.com/5832902/209669356-66073e83-19fd-4c5e-bb63-f90ea1642da3.gif) |

## Expected behavior

No jitter on resize. Widget containers with managed size use `contain: strict`.

PR #506, opened 2022-12-27
Add contain strict for elements with fully managed layout

krassowski:
Closes https://github.com/jupyterlab/lumino/issues/504

Uses strict CSS containment on elements with fully managed layout (absolutely positioned elements that specify all coordinates and dimensions, e.g. top + left + width + height, or top + left + bottom + right) to reduce layout cost (improves performance for resizing certain elements and other actions).

This requires that nothing within the widgets paints outside of it. Some modest changes may be needed downstream to remove painting outside of the widget box, as in https://github.com/jupyterlab/jupyterlab/pull/13671.

The added contain can be overridden in sub-classes of `LayoutItem`.

I would suggest that we merge this, cut a release, pull it in JupyterLab and run full test suite (once https://github.com/jupyterlab/jupyterlab/pull/13671 is in) to see if it causes any side-effects (I did not notice any in manual testing).

I would suggest that we do not backport this to 1.x and instead include a tailored solution in JupyterLab (I will prepare a PR).

### DIFF
--- a/packages/widgets/src/accordionlayout.ts
+++ b/packages/widgets/src/accordionlayout.ts
@@ -241,6 +241,7 @@ namespace Private {
   ): HTMLElement {
     const title = renderer.createSectionTitle(data);
     title.style.position = 'absolute';
+    title.style.contain = 'strict';
     title.setAttribute('aria-label', `${data.label} Section`);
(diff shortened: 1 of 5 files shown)


fcollonval (approved):
Thanks @krassowski 

Merging a cutting a v2 alpha release

PR #507, opened 2022-12-29
Accessibility: role attributes

brichet:
Related to https://github.com/jupyter/notebook/issues/6671.

Changes the roles of (1) accordion title and (2) toggleable items of command palette according to the [W3C recommendation ](https://www.w3.org/TR/wai-aria-1.1/#widget_roles).

### DIFF
--- a/packages/widgets/src/accordionpanel.ts
+++ b/packages/widgets/src/accordionpanel.ts
@@ -401,7 +401,7 @@ export namespace AccordionPanel {
      */
     createSectionTitle(data: Title<Widget>): HTMLElement {
       const handle = document.createElement('h3');
-      handle.setAttribute('role', 'button');
+      handle.setAttribute('role', 'tab');
       handle.setAttribute('tabindex', '0');
       handle.id = this.createTitleKey(data);
       handle.className = this.titleClassName;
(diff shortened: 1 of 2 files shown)


krassowski:
Regarding the command palette, the specs say:

> Authors MUST ensure that menu item checkboxes are [owned](https://www.w3.org/TR/wai-aria-1.1/#dfn-owned-element) by an element with role [menu](https://www.w3.org/TR/wai-aria-1.1/#menu) or [menubar](https://www.w3.org/TR/wai-aria-1.1/#menubar) in order to identify that they are related widgets. Authors MAY separate menu items into sets by use of a [separator](https://www.w3.org/TR/wai-aria-1.1/#separator) or an element with an equivalent role from the native markup language.

If I am reading this correctly, changing the role of checkbox to `menuitemcheckbox` may not suffice to fix the issue here. I would agree that command palette can be seen as a menu but then it currently has `role="region"` in JupyterLab (on `. lm-CommandPalette`). I think we should consider adding `role="menu"` on `.lm-CommandPalette-content` (as the searchbar is not a menu itself). Additionally we have `.lm-CommandPalette-header` (not sure if those count as separators).

brichet:
Thanks @krassowski for reviewing it.

You're right, we should add the `role="menu"` on `.lm-CommandPalette-content`, and probably also a `role="menuitem"` for all those without checkbox.

Not sure about the header, maybe we can leave it without role.

krassowski (approved):
Looks good, thanks!

PR #517, opened 2023-01-14
Enable creating widgets in shadow DOM

### LINKED ISSUE #434: Shadow DOM support for widgets
### Problem

Lumino can be used in situations where there are hundreds of thousands of nodes, and where stylesheets from different providers are attached at the same time. Limiting the stylesheet applicability to a specific subset of nodes is possible with two approaches:
- [CSS contain property](https://developer.mozilla.org/en-US/docs/Web/CSS/contain)
- [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM)

Lumino currently does not support shadow DOM in the sense that individual widgets cannot be moved to shadow DOM and there are no methods exposed allowing to attach stylesheets to specific widgets/shadow DOM roots.

### Proposed Solution

Note: If you have not worked with shadow DOM before please see the details below to understand why the solution (2) wraps the shadow root into another `<div>`.

<details>

DOM nodes can be moved to the shadow DOM by attaching them to a transient shadow DOM root which is attached to another DOM node. There can ever be only one shadow DOM root in each DOM element:

This is allowed:

```html
<div>
    # shadow root
        <div class="element-in-shadow-dom"></div>
        <div class="another-element-in-shadow-dom"></div>
</div>
<div>
    # shadow root
        <div class="element-in-another-shadow-dom"></div>
</div>
```

This is forbidden (multiple shadow roots were [removed from the specification in 2015](https://www.w3.org/wiki/Webapps/WebComponentsApril2015Meeting)):

```html
<div>
    # shadow root
        <div class="element-in-shadow-dom"></div>
    # shadow root
        <div class="another-in-another-shadow-dom"></div>
</div>
```

</details>

The two solutions which I considered are:
1. At widget attachment, check if node of the node of the parent widget hosts a shadow root and if it does, attaching to the shadow root instead of the parent node itself:
    ```diff
    -    this.parent!.node.insertBefore(widget.node, ref);
    +    (this.parent!.node.shadowRoot || this.parent!.node).insertBefore(widget.node, ref);
    ```
    in this scenario, the parent widget is in the shadow DOM:
    ```html
    <div class="lm-Widget">    <!-- this.parent.node -->
        # shadow root    <!-- this.parent.node.shadowRoot -->
            <div class="lm-Widget">    <!-- widget.node -->
            <div class="lm-Widget">    <!-- anotherWidget.node -->
    </div>
    ```
2. Separate `Widget.node` and `Widget.attachmentNode`; by default `Widget.attachmentNode` would be just an alias for `Widget.node` but when a widget is instructed to wrap itself into shadow DOM, it would point to the ShadowRoot (but wrapped into a transient div to avoid the issue of multiple-roots)
    ```diff
    -    this.parent!.node.insertBefore(widget.node, ref);
    +    this.parent!.node.insertBefore(widget.attachmentNode, ref);
    ```
    in this scenario, the child widgets are in the shadow DOM:
    ```html
    <div class="lm-Widget">    <!-- this.parent.node -->
        <div class="lm-transientNode">    <!-- widget.attachmentNode -->
            # shadow root    <!-- widget.attachmentNode.shadowRoot -->
                <div class="lm-Widget">    <!-- widget.node -->
        <div class="lm-transientNode">    <!-- anotherWidget.attachmentNode -->
            # shadow root    <!-- anotherWidget.attachmentNode.shadowRoot -->
                <div class="lm-Widget">    <!-- anotherWidget.node -->
    </div>
    ```
3. Using `Proxy` to create Frankenstein-kind hybrid of `ShadowRoot` and `HTMLElement` and keeping `node` without changes: this does not work because native code in `insertBefore` and friends does not accept non-native nodes (strict class checks are performed at runtime so even if the proxy quacks like a perfect node, it will be rejected).

Then we could either:
- a) allow each widget to be moved into shadow DOM via a new option in constructor, or
- b) provide a subclass of Widget say `ShadowWidget` which would put its node into shadow DOM.

Further we would need to be consider how to handle CSS stylesheets. Widgets could have a method like:

```ts
class Widget {
  /**
  * Adopt a constructed CSS stylesheet for the use in shadow DOM.
  * Is a no-op if the widget is not a shadow DOM root.
  * @returns whether the stylesheet was successfully adopted.
  */
  adoptStyleSheet(stylesheet: CSSStyleSheet): boolean {
    if (!this.options.shadowDOM) { return false; }   // not needed in (b)
    this.node.shadowRoot.adoptedStyleSheets.push(stylesheet);
    return true;
  }
}
```

I am slightly leaning towards 2a as that would allow us to enable shadow DOM downstream by changing the options without a need to duplicate class definitions. The separation of `node` and `attachmentNode` could be introduced in Lumino 2.0 whether we decide to proceed with the shadow DOM implementation or not.

### Additional context

I was able to set it up in JupyterLab and created a code snipped to copy all stylesheets for adoption in widgets, thus limiting the performance benefits to containing style changes to nodes within the shadow DOM but still using all styles; this is super preliminary and I will update if I get a chance to perform a proper benchmark:
- I saw very little performance benefits when moving `MainAreaWidget` to shadow DOM
- I saw more performance benefits when moving the entire `SplitPanel` to shadow DOM

My confidence in the results above is low. I expect that gains will differ depending on the browser.

krassowski:
Fixes #434.

Introduces a distinction between `attachmentNode` and a content `node`.

Adds new methods:
- `adoptStyleSheet()`
- `removeAdoptedStyleSheet()`

### DIFF
--- a/docs/source/migration.md
+++ b/docs/source/migration.md
@@ -89,6 +89,15 @@ while (!(it = it.next()).done) {
 }
 ```
 
+### Replace `widget.node` with `widget.attachmentNode`
+
+Lumino 2 distinguishes between the contents node (`node`) and attachment node
+(`attachmentNode`) of widgets to enable attaching widgets via shadow DOM root.
+By default attachment and contents node are the same, but for widgets with
+shadow DOM enabled, they differ. Downstream layouts need to update methods
+attaching and detaching widgets to use attachment node if they want to support
+moving the widgets to shadow DOM.
+
 ## Public API changes
 
 ### `@lumino/algorithm`
--- a/packages/widgets/src/docklayout.ts
+++ b/packages/widgets/src/docklayout.ts
@@ -533,7 +533,7 @@ export class DockLayout extends Layout {
    */
   protected attachWidget(widget: Widget): void {
     // Do nothing if the widget is already attached.
-    if (this.parent!.node === widget.node.parentNode) {
+    if (this.parent!.node === widget.attachmentNode.parentNode) {
       return;
     }
 
@@ -546,7 +546,7 @@ export class DockLayout extends Layout {
     }
 
     // Add the widget's node to the parent.
-    this.parent!.node.appendChild(widget.node);
+    this.parent!.node.appendChild(widget.attachmentNode);
 
     // Send an `'after-attach'` message if the parent is attached.
     if (this.parent!.isAttached) {
@@ -564,7 +564,7 @@ export class DockLayout extends Layout {
    */
   protected detachWidget(widget: Widget): void {
     // Do nothing if the widget is not attached.
-    if (this.parent!.node !== widget.node.parentNode) {
+    if (this.parent!.node !== widget.attachmentNode.parentNode) {
       return;
     }
 
@@ -574,7 +574,7 @@ export class DockLayout extends Layout {
     }
 
     // Remove the widget's node from the parent.
-    this.parent!.node.removeChild(widget.node);
+    this.parent!.node.removeChild(widget.attachmentNode);
 
     // Send an `'after-detach'` message if the parent is attached.
     if (this.parent!.isAttached) {
--- a/packages/widgets/src/panellayout.ts
+++ b/packages/widgets/src/panellayout.ts
@@ -212,7 +212,7 @@ export class PanelLayout extends Layout {
     }
 
     // Insert the widget's node before the sibling.
-    this.parent!.node.insertBefore(widget.node, ref);
+    this.parent!.node.insertBefore(widget.attachmentNode, ref);
 
     // Send an `'after-attach'` message if the parent is attached.
     if (this.parent!.isAttached) {
@@ -251,7 +251,7 @@ export class PanelLayout extends Layout {
     }
 
     // Remove the widget's node from the parent.
-    this.parent!.node.removeChild(widget.node);
+    this.parent!.node.removeChild(widget.attachmentNode);
 
     // Send an `'after-detach'` and  message if the parent is attached.
     if (this.parent!.isAttached) {
@@ -267,7 +267,7 @@ export class PanelLayout extends Layout {
     }
 
     // Insert the widget's node before the sibling.
-    this.parent!.node.insertBefore(widget.node, ref);
+    this.parent!.node.insertBefore(widget.attachmentNode, ref);
 
     // Send an `'after-attach'` message if the parent is attached.
     if (this.parent!.isAttached) {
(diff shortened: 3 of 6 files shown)


fcollonval (commented):
My general feeling is that it makes `Widget` even more complicated. And it will create confusion as the style sheet methods are not always available.

As nothing prevent a developer to set the widget node as a shadowRoot. I would prefer developer to go that path and knows explicitly that a particular Widget contains a shadowRoot.
To clarify, I would prefer developer to do:

```ts
const node = document.createElement('div');
const root = node.attachmentNode.attachShadow({mode: 'open'});
// populate `root` with the widget view
// ...
const w = new Widget({node});

// deal with local style sheet if needed
// ...
```

This is similar to wrapping a custom elements (= webcomponents).

```ts
const node = document.createElement('my-element');
const w = new Widget({node});
```

> `my-element` will contain a shadow DOM by definition.

krassowski:
There is an appeal of simplicity in what you propose, but there would be inconsistencies too, for example, `Widget.addClass()` and friends would effectively be a no-op because of the shadow DOM boundary, whereas the proposed code allows to inject the shadow root without changing the contract.

Also, we would end up duplicating the code for managing adopted styles. The alternative could be to use the composition approach that you suggest in a new subclass say `ShadowDOMWidget`:

```ts
class ShadowDOMWidget extends Widget {
   constructor(options: {node: Node}) {
      const wrapper = document.createElement('div');
      const root = wrapper.attachShadow({mode: 'open'});
      super({...options, node: wrapper});
      this._root = root;
   }
  adoptStyleSheet(sheet) {}
  removeAdoptedStyleSheet(sheet) {}
}
```

but then we cannot just make shadow DOM optional the same way as we do with `hiddenMode` (unless in the very leaf elements).

krassowski:
> but this PR is not meant to bring all three technologies of web components

but I am happy to consider whether we should go all in (if it simplifies things and enables other things thanks to slots, etc), but that would likely be Lumino 3 given the timelines.

fcollonval:
Reporting it some notes of the performance meeting:

Discussion around web components adoption in the future in lumino/lab ecosystem:
- in future maybe a new alternative implementation of Widget which would be web-component based
- places where react is used poorly (too many roots), like toolbar buttons
- icons may be a good target for gradual adoption, may give a boost
  - svg `<use>`

Mike also reported that the measured gains so far were not as impressive as with other performance improvements he did previously.

So I would prefer adding an new `ShadowDOMWidget` - maybe tagging it as experimental. And if a test case shows a important performance boost, we could use it at very specific places.

krassowski:
In that case I am removing this PR from Lumino 2.0 milestone as I believe we can add `ShadowDOMWidget` in 2.1. I will switch this one to draft for time being.

krassowski:
Even if going for `ShadowDOMWidget` we still need to add the distinction between `node` and `attachmentNode` in `Widget` class as otherwise `attachWidget()` methods would need to sepcial-case `ShadowDOMWidget`.

PR #531, opened 2023-02-05
Check for hidden items before re-rendering

alec-kr:
This PR fixes issue [#13677](https://github.com/jupyterlab/jupyterlab/issues/13677) from jupyterlab.

### Issue
Hidden items are still being re-rendered, despite the fact that they shouldn't be. 

### Solution
Added a check to [```onUpdateRequest```](https://github.com/jupyterlab/lumino/blob/fa048e1680cce138af24ee1aa964f2f02b5e61f6/packages/widgets/src/commandpalette.ts#L289) to determine if an item is hidden, before re-rendering it. This will prevent hidden items from being re-rendered, thus solving the issue.

### DIFF
--- a/packages/widgets/src/commandpalette.ts
+++ b/packages/widgets/src/commandpalette.ts
@@ -287,64 +287,65 @@ export class CommandPalette extends Widget {
    * A message handler invoked on an `'update-request'` message.
    */
   protected onUpdateRequest(msg: Message): void {
-    if (!this.isHidden) {
-        // Fetch the current query text and content node.
-        let query = this.inputNode.value;
-        let contentNode = this.contentNode;
+    if (this.isHidden) {
+      return;
+    }
+    // Fetch the current query text and content node.
+    let query = this.inputNode.value;
+    let contentNode = this.contentNode;
 
-        // Ensure the search results are generated.
-        let results = this._results;
-        if (!results) {
-        // Generate and store the new search results.
-        results = this._results = Private.search(this._items, query);
+    // Ensure the search results are generated.
+    let results = this._results;
+    if (!results) {
+      // Generate and store the new search results.
+      results = this._results = Private.search(this._items, query);
 
-        // Reset the active index.
-        this._activeIndex = query
-            ? ArrayExt.findFirstIndex(results, Private.canActivate)
-            : -1;
-        }
+    // Reset the active index.
+    this._activeIndex = query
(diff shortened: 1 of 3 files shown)


packages/widgets/src/commandpalette.ts
```diff
@@ -287,62 +287,64 @@ export class CommandPalette extends Widget {
...
-    if (!results) {
-      // Generate and store the new search results.
-      results = this._results = Private.search(this._items, query);
-
-      // Reset the active index.
-      this._activeIndex = query
-        ? ArrayExt.findFirstIndex(results, Private.canActivate)
-        : -1;
-    }
+    if (!this.isHidden) {
```
krassowski, line 290: Could this condition be reversed to limit the number of lines changed (/indented), like

```
if (this.isHidden) {
   return;
}
```

?

aleckr: Sure! No idea why I didn't think of this :man_facepalming:

alec-kr: Yes! No idea why I didn't think of that at first 🤦🏽‍♂️

alec-kr: The indentations have now been reverted

fcollonval (changes requested):
Thanks @alec-kr 

Could you trigger an update when the command palette is visible? This will ensure we apply the command palette updates shunted when it was hidden.

```typescript
protected onAfterShow(msg: Message): {
  this.update();
  super.onAfterShow(msg);
}
```
https://github.com/jupyterlab/lumino/blob/38757e291d84837b8dfdb43f5098951feaec3d40/packages/widgets/src/widget.ts#L670

alec-kr:
Sure! I've updated the method. Is this correct?

fcollonval:
> Sure! I've updated the method. Is this correct?
> 
> Edit: I definitely didn't do this right. Where exactly should this change be made? Sorry about the confusion

Sorry my fault I mislead you by referring to the inherited method. You need to add my snippet within the `CommandPalette` class; for example there: 

https://github.com/jupyterlab/lumino/blob/e1e1b11f21ecf2892ebceb7038a4eb3b0311308c/packages/widgets/src/commandpalette.ts#L274

fcollonval (approved):
Thanks @alec-kr

alec-kr:
Thank you for your patience 😄

PR #534, opened 2023-02-07
Remove cursor style

### LINKED ISSUE #533: Lumino should not set cursor (css issue)
👋 
In ipywidgets, we have the cursor set to 'default' due to lumino. This causes links to not have a "hand" cursor but a normal pointer instead. See https://github.com/jupyter-widgets/ipywidgets/issues/3692

I think this CSS property should not be set by lumino on lm-Widget at https://github.com/jupyterlab/lumino/blob/dbb4f6aae04996abfbc5ee17d5fa8e693f3eb024/packages/widgets/style/widget.css#L19

cc @SylvainCorlay @ibdafna 

See also https://github.com/jupyter-widgets/ipywidgets/pull/3696

fcollonval:
Fixes #533 

Remove CSS cursor rule on Widget.

### DIFF
--- a/packages/widgets/style/widget.css
+++ b/packages/widgets/style/widget.css
@@ -16,7 +16,6 @@
   box-sizing: border-box;
   position: relative;
   overflow: hidden;
-  cursor: default;
 }
 
 .lm-Widget.lm-mod-hidden {


krassowski (approved):
Thanks!

PR #537, opened 2023-02-13
Accessibility in accordion panel

brichet:
This PR fixes an accessibility issue about accordion panel:
- removes the role `tab` to the title, which is not in a `tablist` element.
- adds an ID to the widgets if necessary, to correctly link the `aria-controls` attribute of the corresponding title.

### DIFF
--- a/packages/widgets/tests/src/accordionpanel.spec.ts
+++ b/packages/widgets/tests/src/accordionpanel.spec.ts
@@ -77,6 +77,37 @@ describe('@lumino/widgets', () => {
       });
     });
 
+    describe('#accessibility()', () => {
+      it('should create a widget ID if it does not exist', () => {
+        let panel = new LogAccordionPanel();
+        let w1 = new Widget();
+        let w2 = new Widget();
+
+        // Expects a widget ID to be created.
+        expect(!w1.id);
+        panel.addWidget(w1);
+        expect(w1.id);
+
+        // Expects the widget ID to be unchanged.
+        w2.id = 'test_id';
+        panel.addWidget(w2);
+        expect(w2.id === 'test_id');
+      });
+
+      it('should link the widget to its title', () => {
+        let layout = new AccordionLayout({ renderer });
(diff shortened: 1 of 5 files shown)


brichet:
> I guess you use the `Renderer` to reuse the `uuid`. But could you please prefer not changing any public API and add the `id` if it does not exist in `AccordionLayout.insertWidget` - inherited from https://github.com/jupyterlab/lumino/blob/main/packages/widgets/src/panellayout.ts#L87

Thanks for review @fcollonval

packages/widgets/tests/src/accordionpanel.spec.ts
```diff
@@ -77,6 +77,37 @@ describe('@lumino/widgets', () => {
...
     });
 
+    describe('#accessibility()', () => {
+      it('should create a widget ID if it does not exist', () => {
+        let panel = new LogAccordionPanel();
+        let w1 = new Widget();
+        let w2 = new Widget();
+
+        // Expects a widget ID to be created.
+        expect(!w1.id);
```
fcollonval, line 87: I think you cannot just call `expect`, you need to add an assertion like

`expect(w1.id).to.be.undefined;`

For test equality, you should use `expect(a).to.equal('value');`

You can look at the [chai library](https://www.chaijs.com/api/bdd/) to see what is available.

> Could you fix the test at line 123 that probably misguided you as it was missing an assertion.

fcollonval (changes requested):
Thanks @brichet for the changes.

Could you add assertions to the tests and run `yarn run api` to update the API files (the associated CI job is failing).

packages/widgets/tests/src/accordionpanel.spec.ts
brichet: I wanted to make sure the tests pass :smile: 
Thanks for noticing.

fcollonval (approved):
Thanks @brichet

Issue #535, opened 2023-02-09
[DataGrid] Add ability to resize a column's width to fit its content


vthemelis:
This is related to the conversation under https://github.com/bloomberg/ipydatagrid/issues/302

I would like to be able to resize a column in the grid so that it fits the content of every row just like double tapping on the edge between two columns in Excel does.

Currently, there is the ability to resize the grid to fit the content of the column headers but the same is not available for the content of the rows.

vthemelis:
Done

PR #546, opened 2023-02-24
Resize column to fit text on double click

vthemelis:
Implements: https://github.com/jupyterlab/lumino/issues/535 and https://github.com/bloomberg/ipydatagrid/issues/302

Automatically resizes the width of columns by double clicking on the barriers between column headers.

### DIFF
--- a/packages/datagrid/src/datagrid.ts
+++ b/packages/datagrid/src/datagrid.ts
@@ -1407,7 +1407,7 @@ export class DataGrid extends Widget {
   resizeColumn(
     region: DataModel.ColumnRegion,
     index: number,
-    size: number
+    size: number | null
   ): void {
     let msg = new Private.ColumnResizeRequest(region, index, size);
     MessageLoop.postMessage(this._viewport, msg);
@@ -1462,7 +1462,7 @@ export class DataGrid extends Widget {
   }
 
   /**
-   * Auto sizes column widths based on their text content.
+   * Auto sizes column-header widths based on their text content.
    * @param area which area to resize: 'body', 'row-header' or 'all'.
    * @param padding padding added to resized columns (pixels).
    * @param numCols specify cap on the number of column resizes (optional).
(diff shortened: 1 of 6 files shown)


packages/datagrid/src/datagrid.ts
```diff
@@ -2096,6 +2096,54 @@ export class DataGrid extends Widget {
...
+    let maxWidth = 0;
+    for (let j = 0; j < numRows; j++) {
+      const textWidth = this._getCellTextWidth(dataModel, index, j, 'body');
+      maxWidth = Math.max(maxWidth, textWidth);
+    }
+
+    return maxWidth;
+  }
+
+  private _getCellTextWidth(
```
vthemelis, line 2117: This is reusing existing code to find the length of a cell in pixels.

This doesn't seem accurate. Do you have a recommended way for getting the length of any random Renderer? I'm not very familiar with the abstractions.

packages/datagrid/src/datagrid.ts
```diff
@@ -4085,33 +4119,17 @@ export class DataGrid extends Widget {
...
-        };
-
-        // Get the renderer for the given cell.
-        const renderer = this.cellRenderers.get(config) as TextRenderer;
-
-        // Use the canvas context to measure the cell's text width
-        const gc = this.canvasGC;
-        gc.font = CellRenderer.resolveOption(renderer.font, config);
-        const textWidth = gc.measureText(cellValue).width;
+        const textWidth = this._getCellTextWidth(
```
vthemelis, line 4127: NB that this does not translate to the same code as the original.

Initially, the `cellValue` was procured from the `corner-header`, however, the `config` would refer to the `column-header` area. I guess that's a bug?

packages/datagrid/src/datagrid.ts
```diff
@@ -6908,8 +6926,13 @@ namespace Private {
      * @param index - The index of column in the region.
      *
      * @param size - The target size of the section.
+     *               If null, then infer the size to fit.
      */
-    constructor(region: DataModel.ColumnRegion, index: number, size: number) {
+    constructor(
+      region: DataModel.ColumnRegion,
+      index: number,
+      size: number | null
```
vthemelis, line 6934: This is not very null-safe at the moment but will be by the time I convert this draft into a mergeable PR.

packages/datagrid/src/datagrid.ts
krassowski: It may need setting more context options for higher fidelity. For reference eliding uses this:

https://github.com/jupyterlab/lumino/blob/129221f9959d0f92e0c69159b464bd23cbc1f0a3/packages/datagrid/src/textrenderer.ts#L210-L217

krassowski:
> It's slow --It will kill your browser if you try it on a large column.

We could have a dedicated code path for very large columns (say > 1000 rows) with a heuristic. It could find e.g. 100 cells with longest text (by character count) and then only `measureText` width of these and take the maximum.

packages/datagrid/src/datagrid.ts
```diff
@@ -2096,6 +2096,77 @@ export class DataGrid extends Widget {
...
+    if (!dataModel) {
+      return null;
+    }
+
+    const numRows = dataModel.rowCount('body');
+    const configs = Array.from({ length: numRows }, (_val, idx) =>
+      DataGrid._getConfig(dataModel, idx, index, 'body')
+    );
+
+    // Heuristic: Sort by the length of the text to render and only fully calculate the text width
```
vthemelis, line 2111: I can decrease the number of selected rows or try using a heap/quickselect for getting the top-N elements instead if this is not adequate.

packages/datagrid/src/datagrid.ts
```diff
@@ -2096,6 +2096,77 @@ export class DataGrid extends Widget {
...
+      metadata: DataGrid._getCellMetadata(dataModel, location, row, col)
+    };
+  }
+
+  private _getTextToRender(config: CellRenderer.CellConfig): string {
+    const renderer = this.cellRenderers.get(config) as TextRenderer;
+    return renderer.getText(config);
+  }
+
+  private _getCellTextWidth(config: CellRenderer.CellConfig): number {
```
vthemelis, line 2151: This feels a bit like something that should b implemented in each TextRendered instead.

packages/datagrid/src/hyperlinkrenderer.ts
```diff
@@ -41,6 +41,25 @@ export class HyperlinkRenderer extends TextRenderer {
    */
   readonly urlName: CellRenderer.ConfigOption<string> | undefined;
 
+  /**
```
vthemelis, line 44:
```suggestion
```

packages/datagrid/src/datagrid.ts
vthemelis: Thanks for the suggestion! This seems to help!

vthemelis:
Issues I've looked into:

* ~~It's inaccurate --I'm reusing existing code to find the length of the text of a cell but it doesn't seem correct.~~
* ~~It's slow --It will kill your browser if you try it on a large column.~~
* ~~The code is a bit untidy but will fix that once we establish a direction.~~

vthemelis:
An issue that I'm seeing on the side is that `dblclick` event seem to get dropped sometimes completely.

I don't think that this is an issue with my code as I observe the same behaviour in `main`. This prohibits the smooth function of the existing edit functionality that also relies on double-clicking cells.

packages/datagrid/src/datagrid.ts
vthemelis: Update: MinHeap won't really help with performance. The main issue seems to be that getting the values of the cells is extremely slow. This is possibly accounted by the fact that the data is stored in a row-major form:

https://github.com/jupyterlab/lumino/blob/9c4933fb2db746346882550fd230238c420f2501/packages/datagrid/src/jsonmodel.ts#L83

I think that what I'm doing here is probably good-enough for now but can change it if you think that this is likely not adequate.

vthemelis:
Hi @krassowski, 

Would you be able to have another look at this PR?

vthemelis:
The CI seems a bit flakey at the moment. Do you know how I can rerun one of the builds?

krassowski:
I kicked the CI, will look at all the PRs by Monday.

packages/datagrid/src/datagrid.ts
```diff
@@ -2096,6 +2096,102 @@ export class DataGrid extends Widget {
...
+  }
+
+  private _getMaxWidthInArea(
+    dataModel: DataModel,
+    index: number,
+    region: DataModel.CellRegion,
+    rowRegion: DataModel.RowRegion
+  ): number {
+    const numRows = dataModel.rowCount(rowRegion);
+    // Will only allocate upt o 1_000_000 elements otherwise performance can tank.
```
krassowski, line 2130:
```suggestion
    // Will only allocate up to 1_000_000 elements otherwise performance can tank.
```

krassowski (commented):
Heads-up: we would like to gradually add tests for any APIs changed in datagrid (which will make reviewing easier and merging PR faster). I had started this in https://github.com/jupyterlab/lumino/pull/585 and once it is in we could add a simple test for the APIs modified here too - I can help with this.

packages/datagrid/src/textrenderer.ts
```diff
@@ -32,6 +32,11 @@ export class TextRenderer extends CellRenderer {
     this.wrapText = options.wrapText || false;
   }
 
+  /**
+   * Horizontal padding for the text
+   */
+  static H_PADDING = 8;
```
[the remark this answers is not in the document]
vthemelis: Sure, do you want me to also add a member on the TextRendrer class to get the padding from?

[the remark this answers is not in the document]
krassowski: Yes, I would suggest to follow the pattern of all other properties (such as `font` or `wrapText`); in particular given the explicit naming of `horizontalAlignment` maybe it should be `horizontalPadding`?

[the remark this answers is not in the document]
krassowski: Or we could just lift this variable to global namespace so that it is not exported and does not introduce a new public API member. I do not have a strong opinion one way or the other.

alvinsga:
Hi, can I get an update on this?

fcollonval:
@vthemelis friendly ping - do you have time to finish this one up?

vthemelis:
> @vthemelis friendly ping - do you have time to finish this one up?

Yes, will look into it this weekend

vthemelis:
> Heads-up: we would like to gradually add tests for any APIs changed in datagrid (which will make reviewing easier and merging PR faster). I had started this in #585 and once it is in we could add a simple test for the APIs modified here too - I can help with this.

@krassowski, I made the changes suggested by your comments but I'm not sure how to best test my changes. The setup seems a bit too involved and I'm not very familiar with UI testing.

Would you mind merging without tests for now?

krassowski:
Let's move forward and add test in a follow up PR

PR #549, opened 2023-03-03
Fix requiring position for `MenuBar` undetermined index

fcollonval:
In https://github.com/jupyter/notebook/pull/6757, the following error was seen:

image

This error has probably appeared due to the conjunction of trouble with the new overflow menu feature and single item menu bar used in notebook v7.

### DIFF
--- a/packages/widgets/src/menubar.ts
+++ b/packages/widgets/src/menubar.ts
@@ -699,7 +699,8 @@ export class MenuBar extends Widget {
     }
 
     // Get position for the new menu >before< updating active index.
-    const position = this._positionForMenu(index);
+    const position =
+      index >= 0 && this._childMenu ? this._positionForMenu(index) : null;
 
     // Before any modification, update window data.
     Menu.saveWindowData();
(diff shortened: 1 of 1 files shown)


fcollonval:
@krassowski I would love your feedback on this as it touches some of the optimization you have been adding to the menubar.

packages/widgets/src/menubar.ts
```diff
@@ -699,7 +699,8 @@ export class MenuBar extends Widget {
     }
 
     // Get position for the new menu >before< updating active index.
-    const position = this._positionForMenu(index);
+    const position =
+      index >= 0 && this._childMenu ? this._positionForMenu(index) : null;
```
krassowski, line 703: I am not sure if `index >= 0` ever gets `false`: the case of `-1` is handled in line 697 and `ArrayExt.findFirstIndex` should always return a non-negative number or `-1`.

But then I am also not sure why `itemNode` would ever be undefined (what would mutate `this.contentNode.children` between line 685 and 702?). Conditioning on `this._childMenu` seems fine but I don't have the full picture right now.

krassowski: Oh, I see  line 697 only short-circuits if index is -1 and child_menu is already open. Ok so this makes sense.

krassowski (approved):
Thank you!

PR #550, opened 2023-03-04
Make `IOverflowMenuOptions.title` optional

jtpio:
Right now it looks like it's possible to disable the overflow menu for a menu bar with:

```typescript
const overflowOptions = {
  overflowMenuOptions: { isVisible: false, title: '' }
};
```

However `title` still needs to be provided even if `isVisible` is set to `false`.

This change makes `title` optional so this does not have to be the case anymore:

```typescript
const overflowOptions = {
  overflowMenuOptions: { isVisible: false }
};
```

---

`overflowMenuOptions.overflowMenuVisible` has been renamed `overflowMenuOptions.isVisible`

### DIFF
--- a/packages/widgets/src/menubar.ts
+++ b/packages/widgets/src/menubar.ts
@@ -1117,7 +1117,7 @@ export interface IOverflowMenuOptions {
    *
    * Defaults to `true`.
    */
-  overflowMenuVisible: boolean;
+  isVisible: boolean;
   /**
    * Determines the title of the overflow menu.
    *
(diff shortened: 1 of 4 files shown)


packages/widgets/src/menubar.ts
```diff
@@ -1110,14 +1114,16 @@ export namespace MenuBar {
 export interface IOverflowMenuOptions {
   /**
    * Determines if a overflow menu appears when the menu items overflow.
+   *
+   * Defaults to `true`.
    */
   overflowMenuVisible: boolean;
```
afshin, line 1120:
```suggestion
  isVisible: boolean;
```

Since `IOverflowMenuOptions` already has "overflow menu" in the name, I propose using the same `isVisible` flag that is used elsewhere in the API.

jtpio: Sounds good :+1:

PR #560, opened 2023-03-18
Disable size containment for split handle

krassowski:
Fixes https://github.com/jupyterlab/jupyterlab/issues/14219

### DIFF
--- a/packages/widgets/style/splitpanel.css
+++ b/packages/widgets/style/splitpanel.css
@@ -18,8 +18,6 @@
 
 .lm-SplitPanel-handle {
   z-index: 1;
-  /* disable size containment as the handle overhangs another div */
-  contain: style!important;
 }
 
 .lm-SplitPanel-handle.lm-mod-hidden {
--- a/packages/widgets/style/splitpanel.css
+++ b/packages/widgets/style/splitpanel.css
@@ -18,6 +18,8 @@
 
 .lm-SplitPanel-handle {
   z-index: 1;
+  /* disable size containment as the handle overhangs another div */
+  contain: style!important;
 }
 
 .lm-SplitPanel-handle.lm-mod-hidden {
--- a/packages/widgets/src/splitlayout.ts
+++ b/packages/widgets/src/splitlayout.ts
@@ -829,7 +829,8 @@ namespace Private {
   ): HTMLDivElement {
     let handle = renderer.createHandle();
     handle.style.position = 'absolute';
-    handle.style.contain = 'strict';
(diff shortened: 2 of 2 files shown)


packages/widgets/style/splitpanel.css
```diff
@@ -18,6 +18,8 @@
 
 .lm-SplitPanel-handle {
   z-index: 1;
+  /* disable size containment as the handle overhangs another div */
+  contain: style!important;
```
JasonWeill, line 22:
```suggestion
  contain: style !important;
```

Unless it's meaningful to have `style!important` as one contiguous token, adding a space would make this more readable.

JasonWeill (approved):

packages/widgets/style/splitpanel.css
andrii-i: Is using !important here OK because essentially we never want the style to be overwritten by anything?

krassowski: This is a good question. In JupyterLab it would be ok with the logic "we are overriding lumino and changing widget setting to not use strict containment in JS is not worth the trouble". But here we are modifying core lumino so maybe another fix by not adding `contain: strict` would be better. I will need to check if we can do so without breaking API.

krassowski: So instead of adding this style we should just remove this line:

https://github.com/jupyterlab/lumino/blob/133d8725f4dd2c3a56b10d917f1a7b116ff17ce8/packages/widgets/src/splitlayout.ts#L832

Thanks for questioning me - the identification of the problem was right but solution was wrong, another reason not to make PRs at 3am even if those are as simple as 2 lines :)

krassowski: Or rather we should change `strict` to `style` here.

andrii-i: No problems! I've searched for '!important' in Lumino before asking and it is used in couple of places in the codebase. Thanks for fix and explanations

andrii-i (approved):
Fixes the problem, side panel "split handle" now behaves in line with previous versions. Below is comparison of 3.2.9 where problem is not yet present (JupyterLab Light theme) vs current head-of-master with proposed fix applied directly to split handle css (JupyterLab Dark theme)

drag_zone

andrii-i:
@krassowski wanted to ask what needs to be done to put lumino with #560 merged into jupyterlab? Is it something like lumino new release --> update jupyterlab dependencies?

krassowski:
Yes. We have a few PRs on lumino which need review that could be included in a new release.

Issue #547, opened 2023-02-25
[DataGrid] double-clicking cells to edit them is flakey


vthemelis:
## Description

Double-clicking editable cells is meant to actuate an editor to change their values (see https://github.com/jupyterlab/lumino/pull/14).

From my experience with the example DataGrids ([here](https://github.com/jupyterlab/lumino/tree/main/examples/example-datagrid)), the `dblclick` event does not reliably fire and it's not always possible to see the editor.

I also noticed the same while working on https://github.com/jupyterlab/lumino/pull/546 which also relies on `dblclick` events. 

## Reproduce

1. Start the example DataGrid page according to the steps in the main README file.
2. Double click on a few cells on the Editable Grid in the top left.
3. You will notice that you cannot consistently trigger the edit mode.

You can always trigger the edit mode by selecting a cell and starting to type though.

## Expected behavior

`dblclick` events should be consistently fired and trigger the editor.

## Context

- Operating System and version: macOS Monterey (12.0.1)
- Browser and version: Chrome Version 110.0.5481.177 (Official Build) (arm64) and Safari Version 15.1 (17612.2.9.1.20)

vthemelis:
CC: @mbektas as the author of https://github.com/jupyterlab/lumino/pull/14

jjrv:
Another clue is that if I comment out this line:
https://github.com/jupyterlab/lumino/blob/d577cc9ebc98401dbc605029d31d5f782329a6c7/packages/datagrid/src/basicmousehandler.ts#L247

A failed double click will select cells between (0,0) and the clicked cell. While that selection is visible, double clicking to edit is seemingly randomly broken.

vthemelis:
Interesting, looks like this `div` was only recently added in https://github.com/jupyterlab/lumino/pull/502 by @krassowski .

Would you like to open a PR to fix it?

jjrv:
The rabbit hole continues in the PR and is deeper than expected.

PR #564, opened 2023-03-22
Fix dblclick events with Drag.overrideCursor active (#547)

jjrv:
The `.lm-cursor-backdrop` div introduced in #502 was preventing dblclick events from reaching datagrid, breaking cell edit functionality.

### DIFF
--- a/packages/dragdrop/src/index.ts
+++ b/packages/dragdrop/src/index.ts
@@ -1232,6 +1232,9 @@ namespace Private {
         ? doc.body
         : (doc.firstElementChild as HTMLElement);
     if (!cursorBackdrop.isConnected) {
+      // Hide the backdrop until the pointer moves to avoid issues with
+      // native double click detection, used in e.g. datagrid editing.
+      cursorBackdrop.style.transform = 'scale(0)';
       body.appendChild(cursorBackdrop);
       document.addEventListener('pointermove', alignBackdrop, {
         capture: true,


jjrv:
Ping @krassowski I don't think making the backdrop div transparent to pointer events would interfere with the performance reasons for which it was created?

krassowski:
This is not immediately clear to me and will require manual benchmarking

jjrv:
@krassowski Hmm! This sounds strange but I tested the Lumino datagrid example with only grid6 (the editable one) and 100000 divs and spans added to the document. Clicks on the datagrid are equally slow regardless of whether this fix is present or not.

However if I do this:
```
.lm-cursor-backdrop {
display:none;
}
```

All the clicks are handled instantly and also bug #547 is fixed. Opening and closing the cell editor a few times crashes the browser tab though, but that sounds like a different issue.

krassowski:
Great to hear! We ideally want to keep the cursor override but of course we can adjust the implementation. I am away from keyboard right now so cannot test but what I am worried about is whether cursor override works when `pointer-events: none` is set. The alternative solution would be to add event listener on the backdrop div and pass the click events down transparently.

jjrv:
I'm saying in the case of this benchmark, the backdrop div is hurting performance and `display: none` certainly turns off the cursor override or any other effects it might have, performance or otherwise. I'd like another minimal benchmark (with Lumino and not the rest of JupyterLab) where the backdrop div has a positive effect, to figure out when it should be present or not, because it seems to have a negative performance effect in this case.

jjrv:
There's also another performance issue involved here because if I only add 2000 divs and spans, open and close the cell editor, then the browser starts choking on every click. With 20000 divs and spans, it works fine but open and close the editor and the tab just crashes. Something about the editor triggers a performance problem unrelated to the backdrop.

jjrv:
OK ignore that benchmark, the performance issues were because I added the test divs directly under body and the backdrop caused a layout slowdown when added as their sibling. That's an unrealistic situation in practice.

Also, the `pointer-events: none` does break the override. Back to the drawing board...

jjrv:
New attempt! The prior functionality was already that the backdrop div is only correctly placed after the mouse moves with the button pressed:

```
      document.addEventListener('pointermove', alignBackdrop, {
        capture: true,
        passive: true
      });
```

So before that `pointermove` event, we might as well hide the div. Then double click events to the underlying element still get detected, and datagrid editing is fixed.

packages/dragdrop/src/index.ts
```diff
@@ -1232,6 +1232,7 @@ namespace Private {
         ? doc.body
         : (doc.firstElementChild as HTMLElement);
     if (!cursorBackdrop.isConnected) {
+      cursorBackdrop.style.transform = 'scale(0)';
```
krassowski, line 1235: This is a clever reuse of transform which will get overriden on mousemove. Can you add a comment explaining what you described in the last post here?

krassowski (commented):
Conceptually looks good.

We may still want to propagate click events through backdrop manually by creating synthetic events but this can be done in a separate PR.

packages/dragdrop/src/index.ts
jjrv: Good point! Comment added.

vthemelis:
Hi, would it be possible to merge this now? It would be great if I could get the edit functionality to work robustly.

packages/dragdrop/src/index.ts
vthemelis: The only thing I'd add is that it's still not obvious to me what the mechanism by which this transform is (if it is) overwritten on mouse move by this comment.

krassowski (approved):
Thank you @jjrv!

PR #578, opened 2023-04-26
Invalid unicode characters removed from datagrid (#456)

### LINKED ISSUE #456: DataGrid eliding does not respect unicode character integrity
## Description

For example https://www.compart.com/en/unicode/U+1F4E6

Screenshot from 2022-11-01 17-19-46

## Reproduce

https://mybinder.org/v2/gh/jupyterlab/jupyterlab-plugin-playground/main?urlpath=lab with:

<details>

```ts
import { JupyterFrontEnd, JupyterFrontEndPlugin } from '@jupyterlab/application';
import { MainAreaWidget } from '@jupyterlab/apputils';
import { DataGrid, DataModel, BasicKeyHandler, BasicMouseHandler, BasicSelectionModel } from '@lumino/datagrid';

class LargeDataModel extends DataModel {
  rowCount(region: DataModel.RowRegion): number {
    return region === 'body' ? 100 : 1;
  }

  columnCount(region: DataModel.ColumnRegion): number {
    return region === 'body' ? 10 : 1;
  }

  data(region: DataModel.CellRegion, row: number, column: number): any {
    if (region !== 'body') {
      return `${row}, ${column}`;
    }
    return '📦'.repeat(20);
  }
}

const extension: JupyterFrontEndPlugin<void> = {
  id: 'datagrid',
  autoStart: true,
  activate: (
    app: JupyterFrontEnd
  ) => {
    const model = new LargeDataModel();
    const content = new DataGrid();
  
    content.dataModel = model;
    // (handlers are not required for reproduction)
    content.keyHandler = new BasicKeyHandler();
    content.mouseHandler = new BasicMouseHandler();
    content.selectionModel = new BasicSelectionModel({
      dataModel: model
    });
    const widget = new MainAreaWidget({content});
    widget.id = 'datagrid-example';
    widget.title.label = 'Unicode Datagrid Example';
    app.shell.add(widget, 'main');
  },
};

export default extension;

```

</details>

1. Rune the example provided above
2. Try resizing one of the columns
3. See unknown character show up

## Expected behavior

Unicode-aware slice is used, see https://stackoverflow.com/a/62341816

nicojapas:
Fixes #456

When dealing with astral symbols and ellipsing, datagrid generates invalid Unicode characters because of the use of `substring()`.

before

With the regular expression `/[\u{D800}-\u{DFFF}]/gu` we match any character falling within the range of surrogate code points. This includes both high surrogates (0xD800 to 0xDBFF) and low surrogates (0xDC00 to 0xDFFF). So any invalid Unicode character resulting from splitting a surrogate pair is removed with `replace()`.

after

### DIFF
--- a/.github/workflows/check-api-changes.yml
+++ b/.github/workflows/check-api-changes.yml
@@ -52,7 +52,7 @@ jobs:
 
       - name: Get API changes
         id: api-changed
-        uses: tj-actions/changed-files@v35.8.0
+        uses: tj-actions/changed-files@v35.9.0
         with:
           base_sha: 'HEAD~1'
           sha: 'HEAD'
--- a/packages/datagrid/src/textrenderer.ts
+++ b/packages/datagrid/src/textrenderer.ts
@@ -325,6 +325,10 @@ export class TextRenderer extends CellRenderer {
           // Otherwise incrementally remove the last character
           text = elide + text.substring(2);
         }
+
+        // Remove invalid unicode characters if any
+        text = text.replace(/[\u{D800}-\u{DFFF}]/gu, "");
+
         textWidth = gc.measureText(text).width;
       }
     }
--- a/packages/datagrid/src/textrenderer.ts
+++ b/packages/datagrid/src/textrenderer.ts
@@ -306,28 +306,31 @@ export class TextRenderer extends CellRenderer {
 
     // Compute elided text
     if (elideDirection === 'right') {
-      while (textWidth > boxWidth && text.length > 1) {
-        if (text.length > 4 && textWidth >= 2 * boxWidth) {
+      while (textWidth > boxWidth && [...text].length > 1) {
+        // Create array to deal with astral symbols 
+        let textArr = [...text]
+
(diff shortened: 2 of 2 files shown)


krassowski:
Thank you for opening the PR! On conceptual level, what if someone has a table with all Unicode code points? Or if those are mapped to something else in a font. Would it be better to rewrite eliding to use Unicode-aware slice by first converting the string to an array as in https://stackoverflow.com/questions/62341685/javascript-unicode-aware-string-slice/62341816#62341816 ?

nicojapas:
> Thank you for opening the PR! On conceptual level, what if someone has a table with all Unicode code points? Or if those are mapped to something else in a font. Would it be better to rewrite eliding to use Unicode-aware slice by first converting the string to an array as in https://stackoverflow.com/questions/62341685/javascript-unicode-aware-string-slice/62341816#62341816 ?

Hi! Yes, that is a better approach I think. I just commited a new solution.

krassowski (approved):
Thank you! I will open a follow-up PR with unit tests.

PR #581, opened 2023-05-03
added tabindex to tabbar

s596757:


packages/widgets/src/tabbar.ts
```diff
@@ -1830,7 +1830,8 @@ namespace Private {
     let node = document.createElement('div');
     let content = document.createElement('ul');
     content.setAttribute('role', 'tablist');
-    content.className = 'lm-TabBar-content';
+    content.setAttribute('tabindex', '0');
+    content.className = 'lm-TabBar-content-test';
```
krassowski, line 1834: Please restore the previous class.

packages/widgets/src/tabbar.ts
```diff
@@ -1830,7 +1830,8 @@ namespace Private {
     let node = document.createElement('div');
     let content = document.createElement('ul');
     content.setAttribute('role', 'tablist');
-    content.className = 'lm-TabBar-content';
+    content.setAttribute('tabindex', '0');
```
krassowski, line 1833: Please provide a rationale for setting the tabindex in the PR description explaining why it is needed and linking any relevant issues (in this or other repositories).

yarn.lock
krassowski, line 1: Please undo all the changes to `yarn.lock` file. If you wish to upgrade any dependencies please do this in a separate dedicated PR unless the PR requires it (this one does not as far as I can see).

krassowski (changes requested):

PR #583, opened 2023-05-09
Added "tabindex=0" for sidebar accessibility

j264415:
References
[#9686 ](https://github.com/jupyterlab/jupyterlab/issues/9686) and [#1440](https://github.com/jupyterlab/jupyterlab/pull/14400)

Code changes
Easy part of #9686: added labels and tabindex="0" for sidebar elements so screen-readers can access them.

@krassowski Please review and comment to see if these are the right changes needed.

### DIFF
--- a/packages/virtualdom/src/index.ts
+++ b/packages/virtualdom/src/index.ts
@@ -175,7 +175,8 @@ export type ARIAAttrNames =
   | 'aria-valuemin'
   | 'aria-valuenow'
   | 'aria-valuetext'
-  | 'role';
+  | 'role'
+  | 'tabIndex';
 
 /**
  * The names of the supported HTML5 CSS property names.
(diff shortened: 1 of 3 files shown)


packages/virtualdom/src/index.ts
```diff
@@ -175,7 +175,8 @@ export type ARIAAttrNames =
   | 'aria-valuemin'
   | 'aria-valuenow'
   | 'aria-valuetext'
-  | 'role';
+  | 'role'
+  | 'tabIndex';
```
krassowski, line 179: `tabindex` (not no capitalisation) is already defined in `ElementAttrNames` 

https://github.com/jupyterlab/lumino/blob/579425109f937b7e1842eb9a6efacbc585dfd0e3/packages/virtualdom/src/index.ts#L112-L114

I wonder if it would be better to instead modify `createTabARIA` signature so that it returns `ElementARIAAttrs & ElementAttrNames` (and keep the capitalisation the same).

j264415: Hi, the requested changes have just been made in recent commit.

krassowski (approved):
Thank you @j264415! I pushed a commit to fix the failing API changes test. In future, to fix it you can run: `yarn run api` and commit the changes.

CC @gabalafou for awareness

Issue #328, opened 2022-08-04
Add tests for data grid package


afshin:
The `DataGrid` class is widely used and reliable. But it does not have any tests. Without tests, we risk regressions and unintended side-effects going forward.

gabalafou:
I could take an attempt at this later this week. Of all the test files in the repo, which might be the most useful for me to study to help me get started on this one?

afshin:
I think looking at the tests in the `@lumino/widgets` package is probably the best way to approach this.

gabalafou:
I assigned myself this task but just you so know, I won't be able to start working on it until next week at the earliest.

gabalafou:
I think it would be interesting to explore this code some more, but I need to focus on https://github.com/Quansight-Labs/jupyter-a11y-testing, so I'm unassigning myself.

PR #585, opened 2023-05-11
Seed tests for datagrid, test `TextRenderer/drawText`

krassowski:
Follow-up to https://github.com/jupyterlab/lumino/pull/578, first step towards https://github.com/jupyterlab/lumino/issues/328 - this is not much but having the test scaffold should make it easier to add tests as we go.


afshin (approved):
This is great, thank you @krassowski!

PR #588, opened 2023-05-26
Add a 'defer' option to the autoStart argument

brichet:
~~This PR adds an optional `deferred` flag to the plugin options.~~
**EDIT** This PR allows to set the value 'defer' to the `autoStart` argument of a plugin.

This is related to https://github.com/jupyterlab/jupyterlab/issues/14576.

It also adds a function for activating all deferred plugins, preferably after application startup.

It does not modify the behavior of `autoStart`, which means that if `autoStart: false`, the plugin will not be activated using the dedicated function.

### DIFF
--- a/packages/application/src/index.ts
+++ b/packages/application/src/index.ts
@@ -52,20 +52,14 @@ export interface IPlugin<T extends Application, U> {
   description?: string;
 
   /**
-   * Whether the plugin should be activated on application start.
+   * Whether the plugin should be activated on application start or waiting for being
+   * required. If the value is 'defer' then the plugin should be activated only after
+   * the application is started.
    *
    * #### Notes
    * The default is `false`.
    */
-  autoStart?: boolean;
-
-  /**
-   * Whether the plugin should be activated after the application start.
-   *
-   * #### Notes
-   * The default is `false`.
-   */
-  deferred?: boolean;
+  autoStart?: boolean | 'defer';
 
   /**
    * The types of required services for the plugin, if any.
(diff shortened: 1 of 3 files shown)


packages/application/src/index.ts
```diff
@@ -59,6 +59,14 @@ export interface IPlugin<T extends Application, U> {
    */
   autoStart?: boolean;
```
fcollonval, line 60: Could we use `autoStart?: boolean | 'defer'` instead of introducing another flag? This will be consistent with the actual three available state.

packages/application/src/index.ts
```diff
@@ -525,6 +533,31 @@ export class Application<T extends Widget = Widget> {
     return this._delegate.promise;
   }
 
+  /**
+   * Collect the deferred plugins.
+   *
+   * @returns The list of all the deferred plugins.
+   */
+  deferredPlugins(): string[] {
```
fcollonval, line 541: Let's use a getter here:

```suggestion
   * The list of all the deferred plugins.
   */
  get deferredPlugins(): string[] {
```

packages/application/src/index.ts
```diff
@@ -525,6 +533,31 @@ export class Application<T extends Widget = Widget> {
...
+    return Array.from(this._plugins.keys()).filter(
+      id => this._plugins.get(id)!.deferred
+    );
```
fcollonval, line 544: FYI you could use the map directly in this case to avoid the use of `!`:

```suggestion
    return Array.from(this._plugins).filter(
      ([id, plugin]) => plugin.deferred
    ).map(([id, plugin]) => id);
```

fcollonval (changes requested):
Thanks @brichet 

I have some suggestions / questions

packages/application/src/index.ts
brichet: Sure, it should work without breaking the current API.

brichet:
Thanks @fcollonval, I updated the PR to include your suggestions.

brichet:
This is backward-compatible, so an extension with `autoStart: 'defer'` will still work with a previous version of *lumino*. 
This is seen as `autoStart: true`, the extension will not be deferred but will still be loaded.

fcollonval (approved):
Thanks @brichet 

i took the liberty to change slightly the `activateDeferredPlugins` to not return an array of void and some technical debts maintenance removing a Map that is used for a Set.

brichet:
> removing a Map that is used for a Set.

Thanks @fcollonval, I can't remember why I first used a Map for this use case.

fcollonval:
> > removing a Map that is used for a Set.
> 
> Thanks @fcollonval, I can't remember why I first used a Map for this use case.

It is not on you it was there from the start of the project :wink: . I think this was used to ensure id uniqueness back in the day where `Set` did not exist.

PR #590, opened 2023-05-30
elements respond to enter or spacebar

g547315:
References
https://github.com/jupyterlab/jupyterlab/issues/9686

Code changes
Hard part of #9686: added functionality so elements now respond to "enters" or "spaces"on a keyboard like a link or button .

@krassowski Please review and comment

### DIFF
--- a/packages/widgets/src/tabbar.ts
+++ b/packages/widgets/src/tabbar.ts
@@ -575,6 +575,56 @@ export class TabBar<T> extends Widget {
     this._releaseMouse();
   }
 
+    /**
+   * Get all adjecent child nodes from the parent of the focused Element by Id
+   *
+   * #### Notes
+   * This currently only get the id of all children
+   * but can be extended to get other element data
+   * 
+   * findIndex method item array will need to be updated 
+   */
+    getElementData(element: { children: any }) {
+      let elementData = [];
+      if (element.children) {
+        for (let child of element.children) {
+          if (
+            child.classList.contains('lm-TabBar-tab') ||
+            child.classList('lm-TabBar-tab lm-mod-current')
+          ) {
+            elementData.push({
+              id: child.id
+            });
+          }
+        }
+      } else {
+      }
+  
+      return elementData;
+    }
+  
+    /**
+     * Get the index of a child element within its parent using its id 
+     * 
+     * #### Notes
+     * will return index = -1 if an index is not found 
+     */
+    findIndex(
+      items: Array<{
+        id: string;
+      }>,
+      id: string
+    ): number {
+      let index = -1;
+      for (let i = 0; i < items.length; i++) {
+        if (items[i].id === id) {
+          index = i;
+          break;
+        }
+      }
+      return index;
+    }
+
   /**
    * Handle the DOM events for the tab bar.
    *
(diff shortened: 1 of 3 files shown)


g547315:
Addressing the failing test: test:firefox-headless

- The changes do not affect this test as it fails with or without the changes
- Error Message: `30 05 2023 14:27:53.287:ERROR [launcher]: Cannot start FirefoxHeadless
        XPCOMGlueLoad error for file /home/xxx-xxxx/.cache/ms-playwright/firefox-1369/firefox/libmozgtk.so:
libgtk-3.so.0: cannot open shared object file: No such file or directory
Couldn't load XPCOM.`

As for Enforce PR label / enforce-label

- Need a  maintainer to label the PR (don't have permissions )

packages/widgets/src/tabbar.ts
```diff
@@ -711,6 +712,44 @@ export class TabBar<T> extends Widget {
...
+
+          // send simulated request with genatated coordinates.
+          let simPointerClick = new PointerEvent('', {
+            button: 0,
+            clientX: clientX,
+            clientY: clientY
+          });
+
+          this._evtPointerDown(simPointerClick);
+          this._evtPointerUp(simPointerClick);
```
krassowski, line 749: We discussed this on accessibility meeting today. From implementation side we would prefer this to avoid synthetic events and instead copy (or extract to granular private method) any relevant code, which should be pretty minimal for the purpose of switching tabs - you just need to find the index of the `focusedElement` among tab nodes, update `.currentIndex` and call activate as in [here](https://github.com/jupyterlab/jupyterlab/blob/d98ef15bfa916309c82d105a5d442d915d5dd1df/packages/application/src/shell.ts#L782-L793).

packages/widgets/src/tabbar.ts
```diff
@@ -711,6 +712,44 @@ export class TabBar<T> extends Widget {
...
+      event.key === ' '
+    ) {
+      event.preventDefault();
+      event.stopPropagation();
+
+      // Get focus element that is in focus by the tab key
+      let focusedElement = document.activeElement;
+
+      if (focusedElement) {
+        if (focusedElement.className === 'lm-TabBar-addButton') {
```
krassowski, line 733: This should check if the class name is on the list of classes as there can be more than one class (in particular downstream can add other classes).

packages/widgets/src/tabbar.ts
fcollonval: @g547315 do you need help addressing the above comment?

g547315: Code changes
Removed synthetic events onclick event creation
Added helper method to find all adjecent child nodes 
Added helper method to find index of the focusedElement from adjecent child nodes
Added functionality to update currentIndex


User-facing changes
Elements on left and right TabBar and DockPanel TabBar now respond to enter or spacebar

Backwards-incompatible changes
None

g547315:
To fix the "Does PR have API changes" test I ran the yarn run api and got this error

- ERROR: Error parsing ...lumino/packages/widgets/api-extractor.json: The "mainEntryPointFilePath" path does not exist: ...lumino/packages/widgets/types/index.d.ts

- I manually added the changes the command should have but the test still fails. I think it just needs someone to run the command as I am unable to 

Addressing the failing test: Tests / JS (macos-latest, webkit-headless)

- The Error: WebkitHeadless stderr: /home/ec2-user/.cache/ms-playwright/webkit-1751/minibrowser-wpe/bin/MiniBrowser: error while loading shared libraries: libgstreamer-1.0.so.0: cannot open shared object file: No such file or directory

- The changes do not affect this test as it fails with or without the changes

fcollonval:
@g547315 thanks for pushing this. I took the liberty to push some simplification to push this PR through the finish line. Would you mind having a look to my changes?

fcollonval (approved):
@brichet has time to improving further keyboard navigation on tabs. So I'll merge this to allow him to built on top of it.

PR #592, opened 2023-06-12
Use Actual Return character for Mac shortcuts.

Carreau:
The current character is "Leftwards Arrow with Hook" U+21A9 https://www.compart.com/en/unicode/U+21A9  from the "Arrows" block which often is rendered as colored emoji (CF code where U+FE0F Variation Selector-16) was added.

Replace with "Return Symbol" U+23CE from the "Miscellaneous Technical", which is more consistent as many of the other characters in this list are also from the "Miscellaneous Technical" section (Backspace, Alt, Shift..)

This should not need the Variation Selector and will have a similar look to Shift/Maj... in most fonts.

### DIFF
--- a/packages/commands/src/index.ts
+++ b/packages/commands/src/index.ts
@@ -1461,8 +1461,7 @@ namespace Private {
   const MAC_DISPLAY: { [key: string]: string } = {
     Backspace: '⌫',
     Tab: '⇥',
-    // U+FE0E prevents the "enter" character from rendering in color as an emoji
-    Enter: '↩\uFE0E',
+    Enter: '⏎',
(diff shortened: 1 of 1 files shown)


Carreau:
@JasonWeill You were the one who added the Variation Selector 16 recently, maybe you will like this change :-)

JasonWeill (approved):
Nice work!

Carreau:
> Nice work!

I wouldn't have though of the VS16 ! But I guess we are both triggered by emojis in menus :-)

fcollonval (approved):
Thanks @Carreau

PR #595, opened 2023-06-13
Fix position of drag image

### LINKED ISSUE #593: Drag image is rendered off-screen
## Description

Since 2.0.0, the "image" that moves with your mouse when you drag & drop is rendered off-screen instead of under your cursor.

## Reproduce

1. Create a DockPanel with a widget in it
2. Drag the widget tab around

## Expected behavior

An image of the widget's tab moves around with your cursor.

## Context

- Operating System and version: Windows 10
- Browser and version: Chrome 114.0.5735.110
- JupyterLab version: N/A
- `@lumino` version: 2.0.0

I think this was introduced in #502 which replaced
https://github.com/jupyterlab/lumino/blob/578b0a3dfc40ff585dc8ce559c8fb75214a8d4ab/packages/dragdrop/src/index.ts#L234-L235

with
https://github.com/jupyterlab/lumino/blob/60060576f0ded96fb2f6d9d3c46cc9f9ca5dd080/packages/dragdrop/src/index.ts#L234

and now it seems the fixed-position offset is incorrect. Setting
```css
top: 0;
left: 0;
```
sort of fixes it but the top left of the image is anchored to the cursor rather than where you dragged from.

MetRonnie:
Closes #593 

I tried using transform-origin but that didn't work.

Note I've only tested this in my web app using Lumino but not a minimal example.

### DIFF
--- a/packages/dragdrop/src/index.ts
+++ b/packages/dragdrop/src/index.ts
@@ -420,9 +420,6 @@ export class Drag implements IDisposable {
         ? this.document.body
         : (this.document.firstElementChild as HTMLElement);
     body.appendChild(this.dragImage);
-    // Ensure center of drag image will be at cursor:
-    style.top = `-${this.dragImage.clientHeight / 2}px`;
-    style.left = `-${this.dragImage.clientWidth / 2}px`;
   }
 
   /**
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -1024,7 +1024,7 @@ export class DockPanel extends Widget {
     sender.releaseMouse();
 
     // Extract the data from the args.
-    let { title, tab, clientX, clientY } = args;
+    let { title, tab, clientX, clientY, offset } = args;
 
     // Setup the mime data for the drag operation.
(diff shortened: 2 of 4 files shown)


packages/default-theme/style/tabbar.css
```diff
@@ -93,5 +93,4 @@
   min-width: 125px;
   border: none;
   box-shadow: 1px 1px 2px rgba(0, 0, 0, 0.3);
-  transform: translateX(-40%) translateY(-58%);
```
MetRonnie, line 96: I removed this because it's always overridden by

https://github.com/jupyterlab/lumino/blob/3f0be5cd32c1ff892ed887e60409c1a3b37a9238/packages/dragdrop/src/index.ts#L229-L234

krassowski:
I think the proposed solution would be fine performance-wise as it only modifies position once when creating the drag node. I would recommend merging and once a new release is out I can run benchmarks to ensure there is no regressions and investigate manually if needed.

MetRonnie:
@fcollonval Nice 👍

fcollonval:
Thanks both - so I'm merging as is.

PR #599, opened 2023-06-22
Upgrade dev dependencies

fcollonval:
Bump:
- Typescript 4.9 -> 5.1
- Yarn 3.4 -> 3.6.
- Lerna 6.3 -> 7.0
- typedoc 0.23 -> 0.24


fcollonval:
I think it is safe to bump TypeScript from 4.9 t 5.1. But as we had issue before while bumping TypeScript, I would appreciate thoughts from other devs.

cc @jupyterlab/jupyterlab-council

echarles:
I guess yarn and lerna deserve a bunch of attention also. My feedbacks

- CI should ideally be green and stable (now failing on Tests/build)
- I have tested on my local env with `yarn`, `yarn build` and `yarn test` and it works fine
- Most important, JupyterLab should be aligned with those changes. Any PR for that?

jtpio:
> I think it is safe to bump TypeScript from 4.9 t 5.1. But as we had issue before while bumping TypeScript, I would appreciate thoughts from other devs.

This PR may be relevant for the TypeScript 5.1 update.

fcollonval:
> CI should ideally be green and stable (now failing on Tests/build)

:+1: CI is greener (failing test are due to flakiness)

> Most important, JupyterLab should be aligned with those changes. Any PR for that?

@echarles would you mind giving more details about your thinking here? I don't see the reasons for dev deps alignement between Lab and Lumino (it has almost never been true in the past).

---

FYI I computed the diff for the types of `@lumino/widgets` between `main` and this PR:

```diff
diff --color packages/widgets/types/index.d.ts packages/widgets/types.bak/index.d.ts
0a1,4
> /**
>  * @packageDocumentation
>  * @module widgets
>  */
Only in packages/widgets/types.bak: tsdoc-metadata.json
```

The only change is the documentation addition by this PR. So it seems we will avoid the [trouble hit in the past](https://github.com/jupyterlab/lumino/issues/170).

PR #607, opened 2023-07-06
MenuBar: do not focus on hover

gabalafou:
When the MenuBar class was created, it was designed so that whenever a user hovers over an item in the menu with a mouse, the active index would change to the index of the item being hovered. The setter for the active index also had a side effect that would update the DOM in order to apply the active CSS class to the item being hovered so that clients consuming this class could style the item (just like the `:hover` pseudo-class).

Later, in an attempt to make the Lumino menu bar more accessible, tabindex=0 values were added to the menubar items via #187. With the menu bar items keyboard-focusable (but with no CSS styling to indicate it), it made sense to try to synchronize the active index with the item focussed.

However, this had one unfortunate side effect, which was that whenever a user moved their mouse over a menu bar item, the app would move browser focus from anywhere else in the app onto that item. 

This PR is an attempt to fix that issue and to put the menu bar's focus management on better footing.

Originally I refactored the menu bar class to simply not change `activeIndex` on `mousemove`, but I realized that that would probably break a lot of downstream code. For years, since the beginning it seems, the convention of the Lumino menu bar has been that if an item in the menu bar is hovered, then Lumino will apply the `lm-mod-active` class. In the interest of not breaking people's code, I decided to keep that convention.


packages/widgets/src/menubar.ts
```diff
@@ -737,11 +749,11 @@ export class MenuBar extends Widget {
   }
 
   /**
-   * Handle the `'mouseleave'` event for the menu bar.
+   * Handle the `'focusout'` event for the menu bar.
    */
-  private _evtMouseLeave(event: MouseEvent): void {
-    // Reset the active index if there is no open menu.
-    if (!this._childMenu) {
+  private _evtFocusOut(event: FocusEvent): void {
```
gabalafou, line 754: The MenuBar class was originally written to handle mouse and keyboard interaction; however, it really wasn't concerned with browser focus. For example, in the original version, you could use the left and right arrow keys to "move" from one item to another in the menu bar, but browser focus did not move when you pressed those keys. 

There was no way to even put browser focus on individual item nodes in the menu bar until https://github.com/jupyterlab/lumino/pull/187. As a result, it made sense to set the active index to -1 when the mouse left the menu bar. But now that the active index is closely tied to focus, to be consistent we shouldn't reset the active index just because the mouse leaves the menu bar. 

However, we can safely reset the active index if the menu bar loses focus.

gabalafou (commented):
Left some inline comments

marthacryan:
Thank you so much for working on this - and for the super thorough descriptions, research, and codepens! It really helps to have all the context with this stuff cause it can be sort of subtle. I haven't tested this locally but as the person that added a lot of the changes that you're addressing in this PR, I approve! I also might recommend someone that's worked a bit more with Lumino checks this out too (@afshin) cause it's been a while since I made changes to Lumino.

fcollonval:
Hey @gabalafou would you mind running:

```ts
yarn api
yarn lint
```

To fix the CI?

gabalafou:
@fcollonval done! But running `yarn api` changed a bunch of API files that have nothing to do with the changes in this PR, not sure what's going on...

packages/widgets/src/menubar.ts
```diff
@@ -947,6 +973,15 @@ export namespace MenuBar {
...
     readonly tabbable: boolean;
 
+    /**
+     * Whether the item is disabled.
+     *
+     * #### Notes
+     * A disabled item cannot be active.
+     * A disabled item cannot be focussed.
+     */
+    readonly disabled: boolean;
```
fcollonval, line 983: As this is new, we must make it optional to keep the API backward compatible:

```suggestion
    readonly disabled?: boolean;
```

fcollonval (changes requested):
Hey @gabalafou 

Thanks for pushing on this. I rebased (this was the reason of so many api doc changes) and added the menubar to the doc to ease testing.

i only have one comment otherwise it looks good to me.

packages/widgets/src/menubar.ts
gabalafou: good catch, thank you!

fcollonval (approved):
Thanks @gabalafou

PR #615, opened 2023-07-27
Added role of button to addButton to make itkeyboard accessible when using screenreader

s596757:
Issue number: 14814

https://github.com/jupyterlab/jupyterlab/issues/14814 

In order to make the the addLauncher button keyboard accessible when using screen reader, added the role of button to the div element. 

Tested using windows narrator. The add button is now keyboard accessible with screen reader and without it. In order to launch new launcher via keyboard: use Enter key or Spacebar Key.

### DIFF
--- a/packages/widgets/src/tabbar.ts
+++ b/packages/widgets/src/tabbar.ts
@@ -2007,6 +2007,7 @@ namespace Private {
     let add = document.createElement('div');
     add.className = 'lm-TabBar-addButton lm-mod-hidden';
     add.setAttribute('tabindex', '-1');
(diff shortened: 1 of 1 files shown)


fcollonval:
Thanks @s596757 

As #612 added some feature on the tab bar, I rebased this PR. Would you mind giving it a try to confirm it is still fixing the issue?

s596757:
> Thanks @s596757
> 
> As #612 added some feature on the tab bar, I rebased this PR. Would you mind giving it a try to confirm it is still fixing the issue?

Yes this is still working on my end

PR #630, opened 2023-09-06
Datagrid: Introduce AsyncCellRenderer and ImageRenderer

martinRenou:
This is a step towards fixing https://github.com/bloomberg/ipydatagrid/issues/253

## Introducing `AsyncCellRenderer`

Prior to this change, the Lumino datagrid would not allow rendering cells asynchronously. This means that rendering images was not really possible (unless the cell renderer had a reference to the datagrid and requested a redraw once the asynchronous task was done).

This PR introduces the `AsyncCellRenderer` abstract class, which the datagrid uses as following (more or less pseudo-code):
```typescript
if (renderer instanceof AsyncCellRenderer) {
  if (!renderer.isReady(config)) {
    // Paint a placeholder, waiting for all asynchronous tasks to be performed
    renderer.paintPlaceholder(gc, config);

    // Perform all asynchronous tasks (without waiting) then redraw the region of the cell
    renderer.load(config).then(() => {
      this.repaintRegion(cell.position);
    }); 
  } else {
    // The async renderer is ready, we can paint synchronously
    renderer.paint(gc, config);
  }
} else {
  // Classic synchronous painting
  renderer.paint(gc, config);
}
```

This allows supporting asynchronous cell rendering without slowing down the synchronous way.

## Introducing the `ImageRenderer`

This implements the `AsyncCellRenderer` and allows to load images in the datagrid:

Screenshot from 2023-09-06 13-59-48

### DIFF
--- a/packages/datagrid/src/imagerenderer.ts
+++ b/packages/datagrid/src/imagerenderer.ts
@@ -37,13 +37,25 @@ export class ImageRenderer extends AsyncCellRenderer {
     super();
 
     this.backgroundColor = options.backgroundColor || '';
+    this.textColor = options.textColor || '#000000';
+    this.placeholder = options.placeholder || '...';
     this.sizingMode = options.sizingMode || 'fit-height';
   }
+
+  /**
+   * The CSS color for drawing the placeholder text.
+   */
+  readonly textColor: CellRenderer.ConfigOption<string>;
 
   /**
    * The CSS color for the cell background.
    */
   readonly backgroundColor: CellRenderer.ConfigOption<string>;
+
+  /**
+   * The placeholder text.
+   */
+  readonly placeholder: CellRenderer.ConfigOption<string>;
 
   /**
    * Sizing mode. Can be 'original', 'fit', or 'fill'.
@@ -83,7 +95,6 @@ export class ImageRenderer extends AsyncCellRenderer {
 
     const img = new Image();
     img.onload = () => {
-      // Load image
       ImageRenderer.dataCache.set(value, img);
 
       loadedPromise.resolve();
@@ -126,7 +137,7 @@ export class ImageRenderer extends AsyncCellRenderer {
    */
   drawBackground(gc: GraphicsContext, config: CellRenderer.CellConfig): void {
     // Resolve the background color for the cell.
-    let color = CellRenderer.resolveOption(this.backgroundColor, config);
+    const color = CellRenderer.resolveOption(this.backgroundColor, config);
 
     // Bail if there is no background color to draw.
     if (!color) {
@@ -146,13 +157,15 @@ export class ImageRenderer extends AsyncCellRenderer {
    * @param config - The configuration data for the cell.
    */
   drawPlaceholder(gc: GraphicsContext, config: CellRenderer.CellConfig): void {
-    // TODO Consider inheriting from/using a TextRenderer for rendering the placeholder
+    const placeholder = CellRenderer.resolveOption(this.placeholder, config);
+    const color = CellRenderer.resolveOption(this.textColor, config);
+
     const textX = config.x + config.width / 2;
     const textY = config.y + config.height / 2;
 
     // Draw the placeholder.
-    gc.fillStyle = 'black'; // TODO Make this part of the cell config
-    gc.fillText('loading...', textX, textY); // TODO Make this "loading..." part of the cell config
+    gc.fillStyle = color;
+    gc.fillText(placeholder, textX, textY);
   }
 
   /**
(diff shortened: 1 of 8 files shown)


martinRenou:
Ready for review!! CI failure seems unrelated, I don't have the right to restart it

martinRenou:
All green! Thanks to the person who restarted it :D

packages/datagrid/src/imagerenderer.ts
```diff
@@ -0,0 +1,252 @@
...
+      case 'fill':
+        gc.drawImage(img, config.x, config.y, config.width, config.height);
+        break;
+      case 'original':
+        gc.drawImage(img, config.x, config.y);
+        break;
+    }
+  }
+
+  static dataCache = new Map<string, HTMLImageElement | undefined>();
```
krassowski, line 209: Do elements in this cache need to be disposed some way to avoid memory leaks?

krassowski (commented):
Looks great! I kicked the failing test. I left two questions on the API design.

Also, we now have a minimal scaffold for datagrid tests [here](https://github.com/jupyterlab/lumino/tree/f37259e6638eafa243bc0308401c8c674dbb585c/packages/datagrid/tests/src), it would be greatly appreciated if the critical subset of the new API was unit-tested.

packages/datagrid/src/imagerenderer.ts
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * Sizing mode. Can be 'original', 'fit-height', 'fit-width', or 'fill'.
+     * 'fit-height' will make the image fit the available height in the cell, respecting the image size ratio.
+     * 'fit-width' will make the image fit the available width in the cell, respecting the image size ratio.
+     * 'fill' will make the image fill the available space in the cell, NOT respecting the image size ratio.
+     * 'original' will make respect the size of the original image.
+     *
+     * The default is 'fit-height'
+     */
+    sizingMode?: CellRenderer.ConfigOption<SizingMode>;
```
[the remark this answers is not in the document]
martinRenou: That works with me. Note that we cannot apply CSS rules though, so we'll need to parse the percentage and compute it manually. But it's not impossible :D

packages/datagrid/src/imagerenderer.ts
martinRenou: As it's a static property, it will live as long as the page is running, is that right? So I guess there will be no memory leak happening?

martinRenou: I thought we may want to clean the cache periodically, could be done in a separate PR if needed.

krassowski: In context of outputs in a notebook, if user opens and then closes 100 notebooks with different in images in data grids, will the memory stay constant or increase with each notebook? If this was tied to a renderer instance, I imagine we would not need to worry as the reference would not linger, but then we would need to fetch the resource for each output rendering which is not ideal either.

krassowski: Culling the cache periodically might be a solution indeed, but we would want to do this selectively (not touch any currently rendered data grids). So we may want to change the cache implementation in the future to do some reference counting or time-based expiration. To make it easier on future ourselves, what if that was a non-exported global variable so that it is not a part of public API? Any downside to this idea?

martinRenou: Memory will increase if those images are always different.

> To make it easier on future ourselves, what if that was a non-exported global variable so that it is not a part of public API

Right!! We should definitely make that cache property private.

I'm fine with the idea to periodically clean the cache. The browser will have cached the image somewhere anyway, so next time we reload the same image it should not take much time to download again.

krassowski: Yes, absolutely!

packages/datagrid/src/imagerenderer.ts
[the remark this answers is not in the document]
krassowski: Yes, parsing only `%` values for now should be managable. I would encourage doing this in this PR. In the future we might swap the implementation for `CSSStyleValue.parse` once Firefox supports it.

martinRenou:
> Also, we now have a minimal scaffold for datagrid tests [here](https://github.com/jupyterlab/lumino/tree/f37259e6638eafa243bc0308401c8c674dbb585c/packages/datagrid/tests/src), it would be greatly appreciated if the critical subset of the new API was unit-tested

Oops forgot to tackle this, will do

martinRenou:
@krassowski I think it's good for another round of review if you have time!

krassowski (approved):
Thank you @martinRenou!

PR #635, opened 2023-09-21
Keydown event at bubbling phase

brichet:
This PR adds an experimental parameter when starting application to handle the keydown event at bubbling phase.

Currently the keydown events are handled at capture phase. With this PR it with possible to switch from capture to bubbling at startup.

See https://github.com/jupyterlab/jupyterlab/pull/14115#issuecomment-1727686379 for context.

cc. @fcollonval @krassowski @gabalafou

### DIFF
--- a/packages/application/src/index.ts
+++ b/packages/application/src/index.ts
@@ -189,6 +189,20 @@ export class Application<T extends Widget = Widget> {
   }
 
   /**
+   * Getter and setter for the bubblingKeydown experimental flag.
+   *
+   * @experimental
+   */
+  get bubblingKeydown(): boolean {
+    return this._bubblingKeydown;
+  }
+  set bubblingKeydown(value: boolean) {
+    document.removeEventListener('keydown', this, !this._bubblingKeydown);
+    this._bubblingKeydown = value;
+    document.addEventListener('keydown', this, !this._bubblingKeydown);
+  }
+
+  /**
    * Get a plugin description.
    *
    * @param id - The ID of the plugin of interest.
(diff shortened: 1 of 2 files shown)


packages/application/src/index.ts
```diff
@@ -715,6 +721,8 @@ export namespace Application {
      * This will override `startPlugins` and any `autoStart` plugins.
      */
     ignorePlugins?: string[];
+
+    bubblingKeydown?: boolean;
```
krassowski, line 725: Let's document that this is experimental and what is the default value.

packages/application/src/index.ts
```diff
@@ -715,6 +721,8 @@ export namespace Application {
      * This will override `startPlugins` and any `autoStart` plugins.
      */
     ignorePlugins?: string[];
+
```
fcollonval, line 724:
```suggestion
/**
 * Whether to capture keydown event at bubbling or capturing (default) phase for keyboard
 * shortcuts.
 *
 * @experimental
 */
```

packages/application/src/index.ts
```diff
@@ -506,6 +506,8 @@ export class Application<T extends Widget = Widget> {
     // Mark the application as started;
     this._started = true;
 
+    this._bubblingKeydown = options.bubblingKeydown || false;
```
fcollonval, line 509: Could you add a public getter/setter so that it can be change dynamically for testing? Marking those as @experimental or even @ignore to keeping then secret.

fcollonval (changes requested):
Thanks @brichet here are some suggestions

fcollonval (approved):
Thanks @brichet 

@krassowski I would like to get a pre-release of lumino with this on Monday, is it ok for you?

krassowski:
Sounds great, thank you!

fcollonval:
Merging this and doing an alpha release

PR #637, opened 2023-09-25
Accept individual modifier keys as valid keybindings

g547315:
## References
[#15090](https://github.com/jupyterlab/jupyterlab/issues/15090)
- [X] Trigger functionalities on key down and key up of Alt or Alt + Shift
## Code changes
Allow the key down event listener to return a mod key or combination of mods to be processed and matched to its keybinding and the keyboard shortcut associated with it 

Tests have been changed to reflect the new functionality of mod keys. As mod keys now have additional functionality, pressing a mod key will break the sequence when entering a key sequence i.e. [Shift K, Shift L]. The test 'should ignore modifier keys pressed in the middle of key sequence' has been changed to reflect this. A similar change has been made to 'should process key sequences that use different modifier keys' to remove unnecessary keystrokes.  

'should return nothing for keys that are marked as modifier in keyboard layout' has been modified as mod keys now return keyboard events. 

To see the full interned functionality please combine this PR with:
- [X] Added and styled overlay UI element [#633](https://github.com/jupyterlab/lumino/pull/633) 
## User-facing changes
None
## Backwards-incompatible changes

None


brichet (commented):
Thanks for working on this @g547315.

I have some comments below.

Would it be possible to wait for a delay before returning the modifier keys combination (when there are only modifier keys pressed) ?
That could avoid triggering the event linked to that modifier key, while only using the modifier in a shortcut.

tonyfast:
should these modify combinations be configurable?

packages/commands/src/index.ts
```diff
@@ -1250,10 +1250,8 @@ export namespace CommandRegistry {
   export function keystrokeForKeydownEvent(event: KeyboardEvent): string {
     let layout = getKeyboardLayout();
     let key = layout.keyForKeydownEvent(event);
-    if (!key || layout.isModifierKey(key)) {
```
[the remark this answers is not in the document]
g547315: This would break our intended functionality as we need to handle modifier keys alone.
As we need to display an overlay element when the corresponding key Alt or Alt Shift is pressed.

For more context: [Add overlay UI element to display Jupyter tab bar keyboard shortcut number](https://github.com/jupyterlab/jupyterlab/issues/15090)

g547315:
> should these modify combinations be configurable?

could you elaborate please
@tonyfast

tonyfast:
> > should these modify combinations be configurable?

modifier keys for assistive technology vary with locale, operating system, and vendor. there might be edge cases where this specific modifier combination is challenge for certain users. the most assistive solution would allow for configurable modifier key codes with <kbd>Alt + Shift</kbd> as the default.

packages/commands/src/index.ts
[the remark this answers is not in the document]
brichet: Yes, I know the idea is to handle the modifier keys alone.
My comment was about a way to disable this functionality if not necessary, e.g. people who don't want to display the overlay. But this condition should probably be handled in jupyterlab.

packages/commands/src/index.ts
```diff
@@ -1266,7 +1264,21 @@ export namespace CommandRegistry {
...
+    // Handle the edge cases
+    // Handle the edge case for triggering the Alt Shift key binding
+    if (mods.length === 2 && (mods[0] === 'Alt' && mods[1] === 'Shift')) {
+        return "Alt" + ' ' + "Shift" + ' '
+    } 
+
+    // Handle the edge case for triggering the Alt key binding
+    if (mods.length === 1 && mods[0] === 'Alt') {
+      return "Alt" + ' '
+      }
```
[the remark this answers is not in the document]
brichet: It should probably be handled  when formatting the shortcut, to remove the last space if there is only modifiers keys: https://github.com/jupyterlab/lumino/blob/b9d2db30fd0804c2856651ac7aa6e1f889a47fdc/packages/commands/src/index.ts#L1158

Maybe we could add something like the following at the end:
```ts
if (!parts.key) {
  return mods.trim();
}
```

packages/commands/src/index.ts
```diff
@@ -1266,7 +1264,21 @@ export namespace CommandRegistry {
     if (event.metaKey && Platform.IS_MAC) {
       mods.push('Cmd');
     }
-    mods.push(key);
+    if (!mods.includes(event.key)) {
```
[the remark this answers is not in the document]
brichet: The fix breaks the combination including `Ctrl`, I think it should be reverted.

Maybe the double modifiers keys is not a problem finally, according to this comment: https://github.com/jupyterlab/lumino/blob/b9d2db30fd0804c2856651ac7aa6e1f889a47fdc/packages/commands/src/index.ts#L1111

I haven't tested it, but it's worth making sure.

packages/commands/src/index.ts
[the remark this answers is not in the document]
g547315: From what i can tell Jupiter does not currently have a way to detect if a user is using accessibility tooling, or which specific tools they need access to (e.g., keyboard only) making it out of reach to disable the functionality on a per user basis if not needed.

We could implement an opt-in option for users, but the challenge would be making these options discoverable. One way to do this would be to present the options to all users on app load.

I am happy to discuss this option further and possibly implement it in another PR.

g547315:
> > > should these modify combinations be configurable?
> 
> modifier keys for assistive technology vary with locale, operating system, and vendor. there might be edge cases where this specific modifier combination is challenge for certain users. the most assistive solution would allow for configurable modifier key codes with Alt + Shift as the default.

The combinations are configurable by the user through settings

g547315:
@brichet please re-review with the added changes and updated PR description

m158261:
Tests / JS

g547315:
> Thanks for working on this @g547315.
> 
> I have some comments below.
> 
> Would it be possible to wait for a delay before returning the modifier keys combination (when there are only modifier keys pressed) ? That could avoid triggering the event linked to that modifier key, while only using the modifier in a shortcut.

This is now done @brichet

brichet:
I'll take a look at it in the week.
At first glance, I think we should add a delay when there are only mod key.
Otherwise, the related action will be triggered every time we want to do a shortcut using mod keys.

packages/commands/src/index.ts
```diff
@@ -537,14 +537,36 @@ export class CommandRegistry {
       event
     );
 
-    // If there is no exact match and no partial match, replay
-    // any suppressed events and clear the pending state.
-    if (!exact && !partial) {
-      this._replayKeydownEvents();
```
brichet, line 543: This should not be removed, it seems that we cannot use the keyboard keys (other than the Lumino shortcut) anymore.
On my side I can't write in cell, I can't use web browser shortcuts...)

packages/commands/src/index.ts
```diff
@@ -537,14 +537,36 @@ export class CommandRegistry {
...
+
+
+    // If there is an exact match that is not excluded and no partial match, the exact match
+    // can be dispatched immediately. The pending state is cleared so
+    // the next key press starts from the default state.
+    // add new modifier only key commands that should be excluded to the end of the if statement
+    if (
+      exact &&
+      !partial &&
+      !exact?.command.includes('application:activate-sidebar-overlays')
```
brichet, line 561: This is Jupyterlab specific, and should not be used in Lumino.

packages/commands/src/index.ts
```diff
@@ -1266,7 +1289,12 @@ export namespace CommandRegistry {
     if (event.metaKey && Platform.IS_MAC) {
       mods.push('Cmd');
     }
-    mods.push(key);
+    if (!layout.isModifierKey(key)) {
+      mods.push(key);
+      // for  modifier and character key strings
+      return mods.join(' ');
```
brichet, line 1295:
```suggestion
```

Unnecessary, this is the same statement as below

brichet (commented):
Thanks @g547315 
I don't really understand where a delay has been added for modifier keys only...

Is Jupyterlab working with this modification included ? On my side I'm not able to write in a cell for example.

packages/commands/src/index.ts
g547315: I have added it back

packages/commands/src/index.ts
g547315: I have removed it

packages/commands/src/index.ts
g547315: I have generalised it to mod key(s)

g547315:
> Thanks @g547315 I don't really understand where a delay has been added for modifier keys only...
> 
> Is Jupyterlab working with this modification included ? On my side I'm not able to write in a cell for example.

Cell issue should now be fixed.

I terms of how the mod keys are delayed:

Here we don't want to fire the exact matches that are triggered by only modifier key(s)

```javascript
    // If there is an exact match but no partial match, the exact match
    // can be dispatched immediately. The pending state is cleared so
    // the next key press starts from the default state.
    if (
      exact &&
      !partial &&
      !(
        exact.keys.toString() === 'Alt' ||
        exact.keys.toString() === 'Ctrl' ||
        exact.keys.toString() === 'Shift' ||
        exact.keys.toString() === 'Alt Shift'
      )
    ) {
      this._executeKeyBinding(exact);
      this._clearPendingState();
      return;
    }
```
The exact match is then stored for future dispatch (delayed)
```javascript
// If there is both an exact match and a partial match, the exact
    // match is stored for future dispatch in case the timer expires
    // before a more specific match is triggered.
    // which is the case fro 'Alt' and 'Alt Shift'.
    if (exact) {
      this._exactKeyMatch = exact;
    }

    // Store the event for possible playback in the future.
    this._keydownEvents.push(event);
```
A timer that counts down for 1000 milliseconds (1 second) is started. This timer can be stopped by other exact or partial matches with the 1000 milliseconds.
```javascript

    // (Re)start the timer to dispatch the most recent exact match
    // in case the partial match fails to result in an exact match.
    this._startTimer();
  }

  /**
   * Start or restart the pending timeout.
   */
  private _startTimer(): void {
    this._clearTimer();
    this._timerID = window.setTimeout(() => {
      this._onPendingTimeout();
    }, Private.CHORD_TIMEOUT);
  }
```

brichet:
> In terms of how the mod keys are delayed:

thanks for the explanation

packages/commands/src/index.ts
```diff
@@ -545,6 +545,19 @@ export class CommandRegistry {
...
+    // Remove modifier key only keystrokes from the current key sequence.
+    // keystrokes that are modifier key(s) plus another key are not affected
+    // intended functionality: Alt then Alt 1 should result to: ['Alt'] then ['Alt 1']
+    // Removing this code results in mod key duplication: ['Alt', 'Alt 1']
+    if (
+      // Check that only a mod key has been pressed
+      (event.altKey && event.key === 'Alt') ||
+      (event.ctrlKey && event.key === 'Ctrl') ||
+      (event.shiftKey && event.key === 'Shift')
+    ) {
```
brichet, line 557: Should we use `CommandRegistry.isModifierKeyPressed(event))` ?
Isn't it supposed to check whether the key pressed is a modifier ?

packages/commands/src/index.ts
```diff
@@ -553,7 +566,16 @@ export class CommandRegistry {
...
+    if (
+      exact &&
+      !partial &&
+      !(
+        exact.keys.toString() === 'Alt' ||
+        exact.keys.toString() === 'Ctrl' ||
+        exact.keys.toString() === 'Shift' ||
+        exact.keys.toString() === 'Alt Shift'
+      )
+    ) {
```
brichet, line 578:
```suggestion
    const layout = getKeyboardLayout();
    const exactKeys = exact?.keys.toString().split(' ');
    if (
      exact &&
      !partial &&
      !exactKeys.every(key => layout.isModifierKey(key))
    ) {
```
It could be more generic and check all modification key combinations (there is probably a better implementation than this).

packages/commands/src/index.ts
```diff
@@ -613,15 +635,19 @@ export class CommandRegistry {
    */
   private _executeKeyBinding(binding: CommandRegistry.IKeyBinding): void {
     let { command, args } = binding;
-    if (!this.hasCommand(command) || !this.isEnabled(command, args)) {
+    let newArgs: ReadonlyPartialJSONObject = {
+      _luminoEvent: { type: 'keybinding', keys: binding.keys },
+      ...args
+    };
```
brichet, line 641: Not sure about modifying the args of the command.
Can you explain what is the objective ?

packages/commands/tests/src/index.spec.ts
```diff
@@ -1153,7 +1153,7 @@ describe('@lumino/commands', () => {
         document.body.removeEventListener('keydown', keydown);
       });
 
-      it('should ignore modifier keys pressed in the middle of key sequence', () => {
+      it('should register a key sequence', () => {
```
brichet, line 1156: This test and the following should probably not been changed.
IIUC, these modifications will break the key sequence containing modifiers.

brichet (changes requested):

packages/commands/src/index.ts
g547315: Part of a manual re base I did, The intended objective is covered here:
https://github.com/jupyterlab/lumino/commit/3aaa56f7dc4a3ddc48f4fa985fcbedcfd716a947

packages/commands/tests/src/index.spec.ts
g547315: As mod keys now have additional functionality, pressing a mod key will break the sequence when entering a key sequence i.e. [Shift K, Shift L]. 

The test 'should ignore modifier keys pressed in the middle of key sequence' has been changed to reflect this. A similar change has been made to 'should process key sequences that use different modifier keys' to remove unnecessary keystrokes.

'should return nothing for keys that are marked as modifier in keyboard layout' has been modified as mod keys now return keyboard events.

packages/commands/src/index.ts
brichet: OK, that shouldn't be in this PR. Can you fix it ?

brichet:
> In terms of how the mod keys are delayed:

Thinking again about it, I don't think the delay should use the partial/exact mechanism.
I think that we can assume that the "mod keys only" will never be used for regular shortcuts, but only to display helpers.
In my mind, the related command should be triggered if the key(s) is(are) held for some delay only. The way it is implemented now, if you press and release the 'Alt' key, the overlay will be displayed after one second, which would be a strange behavior in my opinion.

The timer should probably be dedicated to this feature, and deleted on key up, if the key(s) is(are) not held.
This way, the "mod keys only" would not be added to `_keydownEvents`, and it would fix the sequences with combinations (see https://github.com/jupyterlab/lumino/pull/637#discussion_r1392881819).

g547315:
> > In terms of how the mod keys are delayed:
> 
> Thinking again about it, I don't think the delay should use the partial/exact mechanism. I think that we can assume that the "mod keys only" will never be used for regular shortcuts, but only to display helpers. In my mind, the related command should be triggered if the key(s) is(are) held for some delay only. The way it is implemented now, if you press and release the 'Alt' key, the overlay will be displayed after one second, which would be a strange behavior in my opinion.
> 
> The timer should probably be dedicated to this feature, and deleted on key up, if the key(s) is(are) not held. This way, the "mod keys only" would not be added to `_keydownEvents`, and it would fix the sequences with combinations (see [#637 (comment)](https://github.com/jupyterlab/lumino/pull/637#discussion_r1392881819)).

Thanks for the review the suggested changes have now been implemented can you please re-review

brichet:
@g547315 thanks again for working on this, adding this feature without breaking anything is not easy.

I opened https://github.com/g547315/lumino/pull/1 on this branch. It is based on what you've done with the timer, I mostly moved some call to avoid breaking sequences.

packages/commands/src/index.ts
```diff
@@ -579,6 +594,36 @@ export class CommandRegistry {
     this._startTimer();
   }
 
+  /**
+   * Process a ``keyup`` event to clear the timer on the modifier id exists.
```
krassowski, line 598: what is `id` here?

krassowski:
The failure is relevant:

```
Run for file in review/api/application.api.md review/api/commands.api.md; do
review/api/application.api.md
review/api/commands.api.md
```

Needs running `yarn run api` and committing changes

packages/commands/src/index.ts
brichet: typo:

`Process a ``keyup`` event to clear the timer on the modifier, if it exists`

brichet, line 598:
```suggestion
   * Process a ``keyup`` event to clear the timer on the modifier, if it exists.
```

g547315:
@brichet Thanks for the second pair of eyes and the contributions to the PR.
Having tested the code change via my dev environment and this binder link [Link](https://gist.github.com/g547315/23c8c4aa917578ab817b9fa20be05516) I can't seem to consistently see the intended functionality.

Mod keys bindings are being detected and matched, however they are not executed consistently or at all

brichet:
> @brichet Thanks for the second pair of eyes and the contributions to the PR. Having tested the code change via my dev environment and this binder link [Link](https://gist.github.com/g547315/23c8c4aa917578ab817b9fa20be05516) (Please click the launch binder button) I can't seem to consistently see the intended functionality.
> 
> Mod keys bindings are being detected and matched, however they are not executed consistently or at all
> 
> * Please note the binder instance has styled overlay UI element that should be displayed when Mod keys bindings are executed

This is nice for testing @g547315.

It works well on my side, when holding the key for 1s (it currently uses the `CHORD_TIMEOUT` variable).

Peek 2023-12-14 18-56


Let's create a new variable with a shorter delay for this feature.

g547315:
> > @brichet Thanks for the second pair of eyes and the contributions to the PR. Having tested the code change via my dev environment and this binder link [Link](https://gist.github.com/g547315/23c8c4aa917578ab817b9fa20be05516) (Please click the launch binder button) I can't seem to consistently see the intended functionality.
> > Mod keys bindings are being detected and matched, however they are not executed consistently or at all
> > 
> > * Please note the binder instance has styled overlay UI element that should be displayed when Mod keys bindings are executed
> 
> This is nice for testing @g547315.
> 
> It works well on my side, when holding the key for 1s (it currently uses the `CHORD_TIMEOUT` variable).
> 
> Peek 2023-12-14 18-56 [ Peek 2023-12-14 18-56 ](https://private-user-images.githubusercontent.com/32258950/290613234-09e2f415-f0f8-4987-a073-941b679bad89.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTEiLCJleHAiOjE3MDI2NTIxNDIsIm5iZiI6MTcwMjY1MTg0MiwicGF0aCI6Ii8zMjI1ODk1MC8yOTA2MTMyMzQtMDllMmY0MTUtZjBmOC00OTg3LWEwNzMtOTQxYjY3OWJhZDg5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFJV05KWUFYNENTVkVINTNBJTJGMjAyMzEyMTUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjMxMjE1VDE0NTA0MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTk0ZmVjMzU2NjRlODRiNmFhMWM5ZjY2OTM3ODNkYThhZjUyYzY5MDY1NmMwNjRmNDY0ZjljOWM4Y2UzZGQyYzAmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.h5XXzHss7YfJGGxNkD_leUaxCbPIRD9U4XLFiO_UfYw) [ ](https://private-user-images.githubusercontent.com/32258950/290613234-09e2f415-f0f8-4987-a073-941b679bad89.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTEiLCJleHAiOjE3MDI2NTIxNDIsIm5iZiI6MTcwMjY1MTg0MiwicGF0aCI6Ii8zMjI1ODk1MC8yOTA2MTMyMzQtMDllMmY0MTUtZjBmOC00OTg3LWEwNzMtOTQxYjY3OWJhZDg5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFJV05KWUFYNENTVkVINTNBJTJGMjAyMzEyMTUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjMxMjE1VDE0NTA0MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTk0ZmVjMzU2NjRlODRiNmFhMWM5ZjY2OTM3ODNkYThhZjUyYzY5MDY1NmMwNjRmNDY0ZjljOWM4Y2UzZGQyYzAmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.h5XXzHss7YfJGGxNkD_leUaxCbPIRD9U4XLFiO_UfYw)
> 
> Let's create a new variable with a shorter delay for this feature.

New variable added and set to 500ms

review/api/widgets.api.md
```diff
@@ -1187,6 +1187,7 @@ export namespace TabBar {
         renderCloseIcon(data: IRenderData<any>): VirtualElement;
         renderIcon(data: IRenderData<any>): VirtualElement;
         renderLabel(data: IRenderData<any>): VirtualElement;
+        renderOverlay(dataset: ElementDataset): VirtualElement;
```
brichet, line 1190:
```suggestion
```
This change does not come from this PR

brichet (approved):

g547315:
This pull request should to be ready for merging. However, I am unable to continue working on it after March 28, 2024. If further refinements are identified, I welcome any future contributors to pick up and progress the work

brichet:
Thanks again @g547315 for working on this.

Since I worked on it, it would be nice to have another opinion before merging.
@krassowski do you think we can merge this PR ?

packages/commands/src/index.ts
```diff
@@ -532,6 +527,26 @@ export class CommandRegistry {
...
+        event
+      );
+      if (exact) {
+        // If the mod keys match an exact shortcut, start a dedicated timer.
+        event.preventDefault();
+        event.stopPropagation();
+        this._startModifierTimer(exact);
+      } else {
+        // Otherwise stop potential existing timer.
+        this._clearModifierTimer();
```
krassowski, line 545: Thanks for the ping @brichet!

Should the modifier timer also be stopped if an exact match for a chord is found?

For example if the sequence is `Ctrl Z` (not `Ctrl + Z` but `Ctrl` followed by `Z`) we would get two events (one for `Ctrl`, one for `Z`). My impression from just reading the code (I could be wrong) is that currently it would trigger both the handler for `Ctrl` and for the `Ctrl` followed by `Z` chord.

In other words, should `_clearModifierTimer` be called in `_clearPendingState` too?

brichet: You are right, but this behavior seems OK if this feature is only used to display overlays.
You may want to keep the modifier key(s) pressed (and the overlays visible) to trigger another shortcut.

If we keep it that way, we may want to document it somewhere to avoid using this modifier key for unexpected action.

krassowski: > You may want to keep the modifier key(s) pressed (and the overlays visible) to trigger another shortcut.

Yes, but I also do not want to have the overlay visible ("sticky") after I already pressed the combo I was interested in. I think if user is pressing two shortcuts and they need overlays, say Ctrl + C and Ctrl + V it is fine it they do that in sequence:
- Ctrl: shows overlay
- Ctrl + C: overlay hides immediately
- Ctrl: shows overlay again
- Ctrl + V: overlay hides immediately again

brichet: Actually, the overlay visible or not depends on the command that will be triggered.
Here it is only about the timeout before triggering the command, so you're right we better clear it an other command is triggered. BTW the timeout is also cleared on key up event with this PR.
I updated it.

krassowski (approved):
Thank you!

Issue #570, opened 2023-03-31
Allow a custom label for menu items, overriding the command name


JasonWeill:
### Problem

Found while working on a fix for https://github.com/jupyter/notebook/issues/6806.

When I create a menu item based on a command, the command's `label` is the only possible label for the menu item:

https://github.com/jupyterlab/lumino/blob/f3ca270d7c1173cc52a8715d1fe7c4e1ea3ab2a5/packages/widgets/src/menu.ts#L1770

I would like to override this `label` to, for example, use a shorter one for a compact menu.

### Proposed Solution

Add an optional property called `label` to `Menu.IItemOptions`. If present, the `label` in `IItemOptions` takes precedence over the `label` for the command.

krassowski:
Thank you for opening this. It looks like a reasonable additive change, I do not see any downsides.

> I would like to override this `label` to, for example, use a shorter one for a compact menu.

This is currently accomplished by functional labels, e.g. for command palette we use longer label by checking for `args['isPalette']` as in the [`terminal:create-new` command](https://github.com/jupyterlab/jupyterlab/blob/ee06f972d274fc0d7b3422af937b6e7a7d84894a/packages/terminal-extension/src/index.ts#L347-L350). There are two problems with this solution:
- command has to know about its uses
- we do not receive `isSomething` from most emitters. Once we rework commands to allow them to [receive the event that triggered them](https://github.com/jupyterlab/lumino/issues/554) it would no longer be a problem as potentially it could enable something like:
  ```ts
  {
    label: (args) => (
      args.event & findMenuForEvent(args.event).?id == 'myCompactMenu'
        ? 'shortLabel'
        : 'longLabel'
    )
  }
  ```

The first problem is a design question - should menus be:
- a) simple with commands being omniscient, or
- b) allowed to take responsibility for labels, reducing the responsibilities of commands.

(b) would imply small changes to where localisation of label would need to happen (in package with command or package with menu, only of practical significance when reusing command in third-party package), but I think it is reasonable.

fcollonval:
Some thoughts, if we want to improve application performance, we should move towards having a no-code description of entry points such as commands in menus or toolbars. This means that we indeed have two paths for customization, allow the command consumer (i.e. menus or toolbars) to override the default command attribute (low code approach) or we go with a code approach like the one suggested by Mike above. But this definitely converge to design questions raised above (and not only for menus):

- Omniscient commands
- Customization by consumer

[...]

PR #644, opened 2023-10-04
adds isKeybinding arg to _executeKeyBinding

andrewfulton9:
Necessary for [jupyterlab issue #15046](https://github.com/jupyterlab/jupyterlab/issues/15046) .

relevant discussion captured in #570.

Simply appends an `isKeybinding` argument to `args` in `_executeKeyBinding`.

### DIFF
--- a/packages/commands/src/index.ts
+++ b/packages/commands/src/index.ts
@@ -613,7 +613,7 @@ export class CommandRegistry {
    */
   private _executeKeyBinding(binding: CommandRegistry.IKeyBinding): void {
     let { command, args } = binding;
-    let newArgs = { ...args, isKeybinding: true };
+    let newArgs: ReadonlyPartialJSONObject = {_luminoEvent: { type: "keybinding", keys: binding.keys }, ...args};
     if (!this.hasCommand(command) || !this.isEnabled(command, newArgs)) {
       let word = this.hasCommand(command) ? 'enabled' : 'registered';
       let keys = binding.keys.join(', ');
--- a/packages/commands/src/index.ts
+++ b/packages/commands/src/index.ts
@@ -711,6 +711,8 @@ export namespace CommandRegistry {
      * the command.
      *
      * This may be invoked even when `isEnabled` returns `false`.
+     * 
+     * If called via a keybinding the passed args will include a `_luminoEvent` that specify the event type and keys pressed for customization.
      */
     execute: CommandFunc<any | Promise<any>>;
 
@@ -866,6 +868,8 @@ export namespace CommandRegistry {
      * command as grayed-out or in some other non-interactive fashion.
      *
      * The default value is `() => true`.
+     * 
+     * If called via a keybinding the passed args will include a `_luminoEvent` that specify the event type and keys pressed for customization
      */
     isEnabled?: CommandFunc<boolean>;
 
@@ -991,7 +995,7 @@ export namespace CommandRegistry {
     /**
      * The arguments for the command, if necessary.
      *
-     * The default value is an empty object.
+     * The default value is an empty object. If a command is activated by _executeKeyBinding, `args = {_luminoEvent: {type: <string>, keys: <string[]>}}`
      */
     args?: ReadonlyPartialJSONObject;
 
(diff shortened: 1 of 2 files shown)


krassowski:
What would be a good place(s) to document this new argument?

andrewfulton9:
That's a good question. It looks like the Lumino docs aren't much more than the API reference, but the function that is directly changed here is private and isn't in the docs. I could add something to the docstring for `IKeyBindingOptions.args` and maybe `ICommandOptions.isEnabled` and `ICommandOptions.execute` referencing the added argument in the case that the command is called from a keybinding. How does that sound?

packages/commands/src/index.ts
```diff
@@ -613,15 +613,16 @@ export class CommandRegistry {
    */
   private _executeKeyBinding(binding: CommandRegistry.IKeyBinding): void {
     let { command, args } = binding;
-    if (!this.hasCommand(command) || !this.isEnabled(command, args)) {
+    let newArgs = { ...args, isKeybinding: true };
```
fcollonval, line 616: Thanks @andrewfulton9 

To move further in the direction of passing the event information.

Should we set isKeybinding with the `binding.keys` instead of a boolean?

krassowski: Great question. For context we discussed exposing a **subset** of the event itself for menus, for an unrelated reason: https://github.com/jupyterlab/jupyterlab/pull/14173#issuecomment-1465292805

I imagine we could go in the direction of having a `ILuminoEvent` in which is (a) fully JSON-serializable, (b) includes context specific to Lumino applications (higher level concepts as `type: "menu"` or `type: "keybinding"`).

andrewfulton9: @fcollonval, do you mean something like `let newArgs = { ...args, isKeybinding: binding.keys };`? That would then allow for more customization in downstream actions. Am I understanding that correctly?

krassowski: I was personally thinking:

```js
newArgs = { ...args, event: { type: "keybinding", keys: binding.keys } }
```

or if we want to "hide" it:

```js
newArgs = { ...args, _luminoEvent: { type: "keybinding", keys: binding.keys } }
```

or something in between. Thoughts?

andrewfulton9: That seems reasonable to me. Would it always be `type` and `keys` for the `_luminoEvent`? Would you just keep `keys` blank if the event type was `menu`?

andrewfulton9: It looks like `ReadonlyPartialJSONObject` is `readonly [key: string]`, so I don't think I can do that without changing `ReadonlyPartialJsonObject` or updating  `CommandRegistry.execute` to include another args type

krassowski: I do not see a problem here. Can you expand? In general the proposed event would be a union of types following a base interface for shared fields like "type". Such an interface/type w should indeed have read-only fields. It should conform to read-only JSON type.

andrewfulton9: I may be misunderstanding an error. when I try to access `newArgs._luminoEvent?.type` I am getting this error:

```
Property 'type' does not exist on type 'string | number | boolean | ReadonlyPartialJSONObject | ReadonlyPartialJSONArray'.
  Property 'type' does not exist on type 'string'.ts(2339)
```

Any idea why it thinks `_luminoEvent` is a string?

krassowski: You will need to add a cast. It should work since I've if the shower values is ReadonlyPartialJSONObject.

fcollonval (commented):
Thanks @andrewfulton9 it looks really nice. I like the idea of prefixing by `_` added arguments.

Would you mind extending the following test to check it receives the new attribute with the correct value: https://github.com/jupyterlab/lumino/blob/42b30f7f1b4b5ee1d1cec75583adc659b57f02bf/packages/commands/tests/src/index.spec.ts#L634

andrewfulton9:
> Thanks @andrewfulton9 it looks really nice. I like the idea of prefixing by `_` added arguments.
> 
> Would you mind extending the following test to check it receives the new attribute with the correct value:
> 
> https://github.com/jupyterlab/lumino/blob/42b30f7f1b4b5ee1d1cec75583adc659b57f02bf/packages/commands/tests/src/index.spec.ts#L634

Thanks! I just added the test.

fcollonval (approved):
Thanks a lot @andrewfulton9

PR #652, opened 2023-10-28
Propagate scroll from backdrop

krassowski:
### References

Fixes https://github.com/jupyterlab/jupyterlab/issues/14881

### Code changes

- makes the `.lm-cursor-backdrop` scrollable using a pseudo-element and stlyes, but hiding the scrollbar
- propagates intercepted scroll events to relevant element with `data-lm-dragscroll` property
- adds `top`/`left` to `.lm-cursor-backdrop` as in https://github.com/jupyterlab/lumino/pull/651

### User-facing changes

Restoring scrolling in browser when dragging cells:

fixed-scrolling-notebook

Once https://github.com/jupyterlab/jupyterlab/pull/15318 is in, also enables scrolling in file browser with selection

after-file-browser

### Backwards-incompatible changes

None

### DIFF
--- a/packages/dragdrop/src/index.ts
+++ b/packages/dragdrop/src/index.ts
@@ -1260,8 +1260,8 @@ namespace Private {
       // Hide the backdrop until the pointer moves to avoid issues with
       // native double click detection, used in e.g. datagrid editing.
       cursorBackdrop.style.transform = 'scale(0)';
-      resetBackdropScroll();
       body.appendChild(cursorBackdrop);
+      resetBackdropScroll();
       document.addEventListener('pointermove', alignBackdrop, {
         capture: true,
         passive: true
--- a/packages/dragdrop/src/index.ts
+++ b/packages/dragdrop/src/index.ts
@@ -1315,38 +1315,25 @@ namespace Private {
       return;
     }
     // Apply the scroll delta to the correct target.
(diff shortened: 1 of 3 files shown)


packages/dragdrop/src/index.ts
```diff
@@ -837,30 +837,51 @@ namespace Private {
...
   /**
-   * Find the event target using pointer position.
+   * Find the event target using pointer position if given, or otherwise
+   * the central position of the backdrop.
    */
   export function findElementBehidBackdrop(
-    event: PointerEvent,
+    event?: PointerEvent,
     root: Document | ShadowRoot = document
   ) {
```
fcollonval, line 846: should we also add 

```ts
export function findElementBehindBackdrop
```

without typo in the name and deprecate this name?

fcollonval (approved):
Thanks @krassowski 

I only have a small suggestion.

packages/dragdrop/src/index.ts
krassowski: Good catch. This is in the `Private` namespace, so I think we can skip deprecation, is my reasoning correct?

fcollonval: indeed I miss that - less deprecation :100:

PR #653, opened 2023-10-30
Remove dynamic keyboard event mode

brichet:
This PR is a follow up of https://github.com/jupyterlab/lumino/pull/635, to remove an experimental implementation.

The keyboard event mode cannot be dynamically set up anymore, only when the application starts.

### DIFF
--- a/packages/application/src/index.ts
+++ b/packages/application/src/index.ts
@@ -608,9 +608,6 @@ export class Application<T extends Widget = Widget> {
    * A subclass may reimplement this method as needed.
    */
   protected addEventListeners(): void {
-    if (this._bubblingKeydown) {
-      console.debug('The keydown events are handled during bubbling phase');
-    }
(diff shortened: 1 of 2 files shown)


packages/application/src/index.ts
```diff
@@ -623,7 +609,7 @@ export class Application<T extends Widget = Widget> {
    */
   protected addEventListeners(): void {
     if (this._bubblingKeydown) {
-      console.log('The keydown events are handled during bubbling phase');
+      console.debug('The keydown events are handled during bubbling phase');
```
fcollonval, line 612: do we want to keep this log?

brichet: Probably not

krassowski (approved):
Thanks!

fcollonval (approved):
Thanks

PR #655, opened 2023-11-02
Close launcher keyboard navigation/shortcut

e218736:
## References

#15303 - Issue no. 3 - Arrow key navigation between launcher tabs not as expected

When screen reader is active, pressing right arrow keys steps through the launcher title (each individual letter) rather than moving to the next launcher tab.

## Code changes

adding close launcher buttons as focusable elements to be navigated to with right arrow keys
enabling focused tab to be closed using "Delete" on keyboard

### DIFF
--- a/packages/widgets/src/tabbar.ts
+++ b/packages/widgets/src/tabbar.ts
@@ -7,17 +7,17 @@
 |
 | The full license is in the file LICENSE, distributed with this software.
 |----------------------------------------------------------------------------*/
-import { ArrayExt } from '@lumino/algorithm';
+import { ArrayExt } from "@lumino/algorithm";
 
-import { IDisposable } from '@lumino/disposable';
+import { IDisposable } from "@lumino/disposable";
 
-import { ElementExt } from '@lumino/domutils';
+import { ElementExt } from "@lumino/domutils";
 
-import { Drag } from '@lumino/dragdrop';
+import { Drag } from "@lumino/dragdrop";
 
-import { Message, MessageLoop } from '@lumino/messaging';
+import { Message, MessageLoop } from "@lumino/messaging";
 
-import { ISignal, Signal } from '@lumino/signaling';
+import { ISignal, Signal } from "@lumino/signaling";
 
 import {
   ElementARIAAttrs,
@@ -26,20 +26,20 @@ import {
   ElementInlineStyle,
   h,
   VirtualDOM,
-  VirtualElement
-} from '@lumino/virtualdom';
+  VirtualElement,
+} from "@lumino/virtualdom";
 
-import { Title } from './title';
+import { Title } from "./title";
 
-import { Widget } from './widget';
+import { Widget } from "./widget";
 
 const ARROW_KEYS = [
-  'ArrowLeft',
-  'ArrowUp',
-  'ArrowRight',
-  'ArrowDown',
-  'Home',
-  'End'
+  "ArrowLeft",
+  "ArrowUp",
+  "ArrowRight",
+  "ArrowDown",
+  "Home",
+  "End",
 ];
 
 /**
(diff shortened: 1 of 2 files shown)


packages/widgets/src/tabbar.ts
```diff
@@ -807,7 +832,15 @@ export class TabBar<T> extends Widget {
...
     } else if (ARROW_KEYS.includes(event.key)) {
       // Create a list of all focusable elements in the tab bar.
-      const focusable: Element[] = [...this.contentNode.children];
+      const focusable: Element[] = [];
+      console.log(focusable);
+
+      if (this.closeLauncherNode) {
+        for (var i = 0; i < this.closeLauncherNode.length; i++) {
+          focusable.push(this.closeLauncherNode[i].parentElement!);
+          focusable.push(this.closeLauncherNode[i]);
```
gabalafou, line 841: I think it may break UX conventions around tabs to make the close button focusable with the left and right arrows. Do you have any examples of established apps that follow this UX? I noticed that VS Code uses the arrow keys to move between tabs, and once you are focussed on a particular tab, you can press the tab key to focus the close button for that tab (a second press of the tab key takes you away from the tab bar).

What do you think of that behavior?

gabalafou (changes requested):
I added some review comments but before we can move forward with this PR we have to remove all the code-formatting changes that were introduced.

gabalafou:
Also, please update the PR description. It's currently a (near) duplicate of https://github.com/jupyterlab/jupyterlab/pull/15347

packages/widgets/src/tabbar.ts
```diff
@@ -777,15 +777,28 @@ export class TabBar<T> extends Widget {
       return;
     }
 
+    // Get focus element that is in focus by the tab key
+    const focusedElement = document.activeElement;
+    // Check if Delete key has been pressed and Delete that tab
+    if (event.key === 'Delete') {
```
krassowski, line 783: Why delete? I tested multiple applications (Chrome, Edge, Firefox, VSCode, Gnome Text Editor) and none of them uses Delete to close a focused tab. In my opinion if there was to be a shortcut added:
- it should be added in JupyterLab not in lumino
- it should be explained why the specific default is proposed

krassowski (changes requested):

packages/widgets/src/tabbar.ts
e218736: Hi @krassowski 
From an accessibility point of view documentation states that this functionality is an optional feature for tab elements: [here](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)  
Happy to remove this and create a new PR in Jupyter, thoughts?

packages/widgets/src/tabbar.ts
```diff
@@ -1829,6 +1856,7 @@ export namespace TabBar {
     createTabARIA(data: IRenderData<any>): ElementARIAAttrs | ElementBaseAttrs {
       return {
         role: 'tab',
+        'aria-label': 'Launcher',
```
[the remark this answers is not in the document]
e218736: Thank you, will remove.

packages/widgets/src/tabbar.ts
krassowski: Since this is a part of a standard I think it is ok to use. Can you the link and rationale to the PR description?

e218736: Yes, have added in

krassowski:
@gabalafou I see that you mentioned this PR in https://github.com/jupyterlab/jupyterlab/pull/15347. Once this is ready I am happy to merge and release but currently your review is blocking here:

image

gabalafou (approved):
I still have some concerns and questions. Will get back to this asap.

gabalafou (changes requested):
oops meant to mark it request changes

PR #662, opened 2023-11-30
Skip processing prevented default key event

fcollonval:
This removes the processing of keydown event by the commands registry if it is prevented (aka already treated).

### DIFF
--- a/packages/commands/src/index.ts
+++ b/packages/commands/src/index.ts
@@ -511,8 +511,13 @@ export class CommandRegistry {
    * events will not invoke commands.
    */
   processKeydownEvent(event: KeyboardEvent): void {
-    // Bail immediately if playing back keystrokes.
-    if (this._replaying || CommandRegistry.isModifierKeyPressed(event)) {
+    // Bail immediately if playing back keystrokes
+    // or if the event has been processed
+    if (
+      event.defaultPrevented ||
+      this._replaying ||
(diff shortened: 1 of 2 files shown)


brichet:
Thanks @fcollonval.

The code looks good, but I did not manage to test it in Jupyterlab.

I understood this PR is primarily to fix the codemirror search VS jupyterlab search, according to https://github.com/jupyterlab/jupyterlab/pull/15459.

I tried to enable the codemirror search without success.
I can also see that the codemirror search is enabled by default for [inline editor](https://github.com/jupyterlab/jupyterlab/blob/746ce2d75652a207fa1cdac07dd05854ac15eafb/packages/codemirror/src/factory.ts#L29). Does it mean that the codemirror search should replace the jupyterlab search in Notebook, for instance ?

fcollonval:
Thanks for the review @brichet 

It will not be easy to test on lab directly.

The main ground is that a keyboard shortcut event handled by codemirror can not be stopped, only marked as default prevented.

But in the generic sense, event that are default prevented should not be treated (except to do some stuff related to focused element changes).

Regarding the search, the search from code mirror has always be enabled in simple editor (easy way for us to provide a feature). But it must and it is deactivated in document (the cell editor required a special treatment as set in the lab pr).

brichet (approved):
Looks good, thanks

PR #665, opened 2023-12-13
Update README.md

novrain:
Add an example which using lumino in Vue3.


fcollonval (approved):
Thanks for sharing @novrain

novrain:
> Thanks for sharing @novrain

Thanks for lumino, it's great.

PR #666, opened 2023-12-14
Keyboard shortcut enhanced accessibility for screen readers

e218736:
References
[Screen reader doesn't read all keyboard shortcut punctuation](https://github.com/jupyterlab/jupyterlab/issues/15497)

Code changes
- Function added to render aria labels for top menu item keyboard shortcuts which include full description rather than punctuation symbol. 
- Test added with key binding command to include punctuation shortcut

User-facing changes
none visible. 
For screen reader users, keyboard shortcuts that include punctuation will be announced by default when focused.

### DIFF
--- a/packages/widgets/src/menu.ts
+++ b/packages/widgets/src/menu.ts
@@ -27,7 +27,7 @@ import {
   ElementDataset,
   h,
   VirtualDOM,
-  VirtualElement
+  VirtualElement,
 } from '@lumino/virtualdom';
 
 import { Widget } from './widget';
@@ -556,7 +556,7 @@ export class Menu extends Widget {
         collapsed,
         onfocus: () => {
           this.activeIndex = i;
-        }
+        },
       });
     }
     VirtualDOM.render(content, this.contentNode);
@@ -708,7 +708,7 @@ export class Menu extends Widget {
    */
   private _evtMouseMove(event: MouseEvent): void {
     // Hit test the item nodes for the item under the mouse.
-    let index = ArrayExt.findFirstIndex(this.contentNode.children, node => {
+    let index = ArrayExt.findFirstIndex(this.contentNode.children, (node) => {
       return ElementExt.hitTest(node, event.clientX, event.clientY);
     });
 
@@ -1177,7 +1177,7 @@ export namespace Menu {
           dataset,
           tabindex: '0',
           onfocus: data.onfocus,
-          ...aria
+          ...aria,
         },
         this.renderIcon(data),
         this.renderLabel(data),
@@ -1221,10 +1221,11 @@ export namespace Menu {
      */
     renderShortcut(data: IRenderData): VirtualElement {
       let content = this.formatShortcut(data);
+      let ariaContent = this.formatShortcutText(data);
       return h.div(
         {
           className: 'lm-Menu-itemShortcut',
-          'aria-keyshortcuts': `${content}`
+          'aria-label': `${ariaContent}`,
         },
         content
       );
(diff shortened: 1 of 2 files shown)


packages/widgets/src/commandpalette.ts
```diff
@@ -973,6 +980,41 @@ export namespace CommandPalette {
...
+     */
+    formatItemAria(data: IItemRenderData): h.Child {
+      const keyToText: { [key: string]: string } = {
+        ']': 'Closing bracket',
+        '[': 'Opening bracket',
+        ',': 'Comma',
+        '.': 'Full stop',
+        "'": 'Single quote',
+        '-': 'Hyphen-minus'
+      };
```
krassowski, line 994: I think that we would want to make it easy to override in subclasses of the renderer so that they could reuse the logic below, but change the values on the right-hand side. In particular, JupyterLab will want to substituted the values with translated versions.

One solution would be making `keyToText` an option that can be passed to the renderer constructor, something like:

```ts
constructor(options: IRenderer.IOptions = {}) {
  this._keyToText = options.keyToText ?? DEFAULT_KEY_TO_TEXT;
}
```

But please read the next comment which tackles the deeper question on where the logic should be placed before implementing the suggestion above.

packages/widgets/src/commandpalette.ts
```diff
@@ -973,6 +980,41 @@ export namespace CommandPalette {
...
+          const individualKeys = result.split('+');
+          let index = individualKeys.indexOf(punctuation);
+          if (index != -1) {
+            individualKeys[index] = keyToText[punctuation];
+          }
+          const textShortcut = individualKeys.join('+');
+          return textShortcut;
+        }
+      }
+      return kbText ? CommandRegistry.formatKeystroke(kbText.keys) : null;
```
krassowski, line 1015: Should this logic be moved to `formatKeystroke`, or maybe even better should `CommandRegistry` have an extra function called `CommandRegistry.formatKeystrokeAriaLabel`? All the renderer would do would be calling such method? Maybe translated `keyToText` should be passed to `CommandRegistry` options to avoid redefining/passing it around in all places we will need it?

packages/widgets/src/menu.ts
```diff
@@ -1371,6 +1378,41 @@ export namespace Menu {
...
+          const individualKeys = result.split('+');
+          let index = individualKeys.indexOf(punctuation);
+          if (index != -1) {
+            individualKeys[index] = keyToText[punctuation];
+          }
+          const textShortcut = individualKeys.join('+');
+          return textShortcut;
+        }
+      }
+      return kbText ? CommandRegistry.formatKeystroke(kbText.keys) : null;
```
krassowski, line 1414: All right, this makes it pretty clear that we need to have this implemented at `CommandRegistry` level (see earlier suggestion on creating `CommandRegistry.formatKeystrokeAriaLabel`)

krassowski (changes requested):
Thank you for working on this!

Did you consider using [aria-keyshortcuts](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-keyshortcuts) instead?

packages/widgets/src/menu.ts
e218736: Hi @krassowski, 

All makes sense, thank you for your comments. I will revert back to draft for now whist I work on these and let you know once ready for review again.

packages/widgets/src/commandpalette.ts
e218736: @krassowski I had tried to use aria-keyshortcuts in place of aria-label however, my screen reader (Windows Narrator) still omits some punctuation. With aria-label it is read out as expected.

I have implemented your other suggestions which should now allow translations to be substituted in JupyterLab

PR #669, opened 2023-12-15
:hammer: make title/tabbar's close icon/class customable.

novrain:
The lumino default theme require font-awsome, but in some project, such as Vue project, they donot use it.
Then the close icon will not show correctly. Import font-awsome to make the SINGLE icon show correctly does't make sences.

changes:

- provide the same api like title.icon, make the close icon customable

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
   "singleQuote": true,
   "trailingComma": "none",
-  "arrowParens": "avoid"
+  "arrowParens": "avoid",
+  "endOfLine": "crlf"
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,5 @@
 {
   "singleQuote": true,
   "trailingComma": "none",
-  "arrowParens": "avoid",
-  "endOfLine": "crlf"
+  "arrowParens": "avoid"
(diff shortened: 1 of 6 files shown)


novrain:
anything that I can do to make this request be accept?

krassowski:
Could you instead override `renderCloseIcon` in a subclass? This way we would not need to extend the API. This is what JupyterLab does [here](https://github.com/jupyterlab/jupyterlab/blob/0a5e152dfb7e9d1dc6f8ba994c3f7fe2bc481420/packages/ui-components/src/icon/widgets/tabbarsvg.ts#L48-L68).

novrain:
> Could you instead override `renderCloseIcon` in a subclass? This way we would not need to extend the API. This is what JupyterLab does [here](https://github.com/jupyterlab/jupyterlab/blob/0a5e152dfb7e9d1dc6f8ba994c3f7fe2bc481420/packages/ui-components/src/icon/widgets/tabbarsvg.ts#L48-L68).

Thank you! That's a great option. I will try it.

fcollonval:
Hey @novrain I'm closing this as answered with alternative. Let me know if you still need some changes for your use case.

PR #683, opened 2024-02-07
Support selectors with shadow-piercing descendant combinator (`/deep/`)

krassowski:
A potential solution for https://github.com/jupyterlab/jupyterlab/issues/15757

By allowing commands to use `/deep/` selector we could filter out shortcuts use on inputs withing shadow DOM.


krassowski:
Closing as we went a different route in https://github.com/jupyterlab/jupyterlab/pull/15774

Issue #688, opened 2024-03-05
New keybinding attribute to indicate shortcuts which prevent default actions


krassowski:
### Problem

- Currently lumino shortcuts can prevent user from typing characters on certain keyboards (because it always calls `preventDefault()` and `stopPropagation()` on matched keyboard events)
- JupyterLab 4.1.3+ solves it by delaying command execution until it knows if a character would have been typed (https://github.com/jupyterlab/jupyterlab/pull/15790)
- However, some shortcuts need to be executed immediately to prevent default browser action
- Currently there is no way to distinguish the situation in which the current keyboard event would invoke a shortcut which needs such immediately handling 

### Proposed Solution

1. Add new attribute `preventDefault: bool = true` to [`IKeyBindingOptions`](https://github.com/jupyterlab/lumino/blob/f1014291ac83e4524bbf57afb2cbdd55ea4ee546/packages/commands/src/index.ts#L962)
2. Add new public method on commands to check if the default action on the event should be prevented, say:
    ```ts
        willPreventDefault(event) {
            // Bail immediately if playing back keystrokes.
            if (this._replaying || CommandRegistry.isModifierKeyPressed(event)) {
                return;
            }
            // Get the normalized keystroke for the event.
            let keystroke = CommandRegistry.keystrokeForKeydownEvent(event);
            // If the keystroke is not valid for the keyboard layout, replay
            // any suppressed events and clear the pending state.
            if (!keystroke) {
                this._replayKeydownEvents();
                this._clearPendingState();
                return;
            }
            // Add the keystroke to the current key sequence.
            const keystrokes = [*this._keystrokes, keystroke];
            // Find the exact and partial matches for the key sequence.
            let { exact } = Private.matchKeyBinding(this._keyBindings, keystrokes, event);
    
            if (exact.preventDefault || partial.preventDefault) {
              return true;
            }
            return false;
        }
    ```
3. Change `processKeydownEvent` to only `preventDefault` when needed, this is change:
    https://github.com/jupyterlab/lumino/blob/f1014291ac83e4524bbf57afb2cbdd55ea4ee546/packages/commands/src/index.ts#L553-L557
    
    to:
    
    ```ts
    if (exact.preventDefault || partial.preventDefault) {
       event.preventDefault(); 
       event.stopPropagation(); 
    }
    ```
4. In JupyterLab:
   a) mark all keybindings as preventDefault=false by default, and flick the few ones which should prevent default manually to true
[...]

PR #689, opened 2024-03-06
Optional prevent default and asynchronous hold for keybinding execution

krassowski:
Closes https://github.com/jupyterlab/lumino/issues/688

- Adds `preventDefault` option to `IKeyBinding` and `IKeyBindingOptions` to enable downstreams to opt out of preventing default on the processed keyboard events
- Adds `CommandRegistry.holdKeyBindingExecution(event: KeyboardEvent, permission: Promise<boolean>): void;` method which enables downstream to hold the execution of a keybinding until the permission to execute resolves to `true`, or to prevent it if it resolves with `false`

### DIFF
--- a/packages/commands/src/index.ts
+++ b/packages/commands/src/index.ts
@@ -644,12 +644,17 @@ export class CommandRegistry {
     if (this._holdKeyBindingPromises.size !== 0) {
       // Wait until all hold requests on execution are lifted.
       const executionAllowed = (
-        await Promise.all(
-          this._keydownEvents.map(
-            async event =>
-              this._holdKeyBindingPromises.get(event) ?? Promise.resolve(true)
-          )
-        )
+        await Promise.race([
+          Promise.all(
+            this._keydownEvents.map(
+              async event =>
+                this._holdKeyBindingPromises.get(event) ?? Promise.resolve(true)
+            )
+          ),
+          new Promise<boolean[]>(resolve => {
+            setTimeout(() => resolve([false]), Private.KEYBINDING_HOLD_TIMEOUT);
+          })
+        ])
       ).every(Boolean);
       // Clear the hold requests.
       this._holdKeyBindingPromises.clear();
@@ -1349,6 +1354,11 @@ namespace Private {
    * The timeout in ms for triggering a key binding chord.
    */
   export const CHORD_TIMEOUT = 1000;
+
+  /**
+   * The timeout in ms for stopping the hold on keybinding execution.
+   */
+  export const KEYBINDING_HOLD_TIMEOUT = 1000;
 
   /**
    * A convenience type alias for a command func.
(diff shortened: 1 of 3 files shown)


packages/commands/src/index.ts
```diff
@@ -579,6 +583,21 @@ export class CommandRegistry {
...
+   * @param event - The event object for a `'keydown'` event.
+   * @param permission - The promise with value indicating whether to proceed with the execution.
+   *
+   * ### Note
+   * This enables the caller of `processKeydownEvent` to asynchronously prevent the
+   * execution of the command based on external events.
+   */
+  holdKeyBindingExecution(event: KeyboardEvent, permission: Promise<boolean>) {
+    this._holdKeyBindingPromises.set(event, permission);
+  }
```
fcollonval, line 599: could you elaborate the need for that API? Is it to deal with composed shortcut requiring multiples shortcut to be pressed in sequence?

packages/commands/src/index.ts
```diff
@@ -615,8 +634,31 @@ export class CommandRegistry {
...
+  ): Promise<void> {
+    if (this._holdKeyBindingPromises.size !== 0) {
+      // Wait until all hold requests on execution are lifted.
+      const executionAllowed = (
+        await Promise.all(
+          this._keydownEvents.map(
+            async event =>
+              this._holdKeyBindingPromises.get(event) ?? Promise.resolve(true)
+          )
+        )
```
fcollonval, line 652: Should we add a race condition with a timeout to avoid a dead lock?

fcollonval (commented):
Thanks a lot @krassowski 

This looks great. I'm only a bit confused about the need for the advanced `holdKeyBindingExecution`.

packages/commands/src/index.ts
krassowski: No, this is strictly to enable https://github.com/jupyterlab/jupyterlab/pull/15927.

The intent is to block execution of a shortcut if its keystroke would lead to a character being typed in (as detected when handling `onbeforeinput` event).

The problem is that we do not know if a keypress would result in a character being produced at the time when `keydown` is handled, because only after `keydown` is handled the `onbeforeinput` event is emitted.

Therefore https://github.com/jupyterlab/jupyterlab/pull/15927 passes a promise to `holdKeyBindingExecution` which resolves to `true` if the subsequent `onbeforeinput` event does not produce a character, and `false` if it does.

packages/commands/src/index.ts
krassowski: There is already a timeout in https://github.com/jupyterlab/jupyterlab/pull/15927. I will add a timeout here too.

krassowski: Done in c5c0a9c7084fd4ecf876053bfc982d6cdd3d76e8.

packages/commands/src/index.ts
fcollonval: Thanks for the details

fcollonval (approved):
Thanks @krassowski

PR #690, opened 2024-03-12
Update Release Scripts

blink1073:
See https://github.com/jupyter-server/jupyter_releaser/pull/557 for context.

### DIFF
--- a/.github/workflows/prep-release.yml
+++ b/.github/workflows/prep-release.yml
@@ -12,6 +12,10 @@ on:
       post_version_spec:
         description: "Post Version Specifier"
         required: false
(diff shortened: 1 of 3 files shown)


jtpio (approved):
Thanks!

krassowski:
Can we delete this branch?

I noticed that I cannot remove it as it is protected along some other branches which I think should not be protected (see https://github.com/jupyterlab/lumino/rules/512358?ref=refs%2Fheads%2Fupdate-release-scripts)

blink1073:
I deleted it and adjusted the protection rules

Issue #699, opened 2024-04-26
Support host and ref options in Menu.open


ianthomas23:
I would like to add `host` and `ref` `HTMLElement` options to `Menu.open`.

Currently `Menu.open` always adds the menu `div` to the DOM as the last child of the `document.body`. This has caused a problem in `ipydatagrid` as although it does give the correct behaviour in JupyterLab, it is not correct for NbClassic, Voila, etc, as their different DOM structure directly under the `body` means the menus have been appearing in the wrong location (full analysis is in [this ipydatagrid comment](https://github.com/bloomberg/ipydatagrid/issues/422#issuecomment-2075292900)).

`Menu.open` doesn't give us any control over where the menu is inserted in the DOM, but `Widget.attach` (which is what is called by `Menu.open`) does via the arguments `host` (the parent element to insert under) and `ref` (the element to insert before). The proposal here therefore is to expose these two arguments via `Menu.IOptions` in `Menu.open`.

The workaround in `ipydatagrid` PR bloomberg/ipydatagrid#509 is to let the menu be attached as normal to `body` by calling `Menu.open`, then immediately use `Widget.detach` and `Widget.attach` to move it to the desired location in the DOM. By exposing the `host` and `ref` in `Menu.open` we will be able to change `ipydatagrid` to attach it directly to the correct location and avoid having to detach and reattach it.

Strictly speaking the `ipydatagrid` use case only needs the `ref` option and not the `host` one. But I think it best to add all rather than part of the API to control exactly where in the DOM the menu is added.

I will submit a PR for this shortly.

ianthomas23:
This was fixed by #700.

PR #700, opened 2024-04-26
Support host and ref options in Menu.open

ianthomas23:
Implements enhancement #699.

This exposes the `Widget.attach` optional arguments `host` and `ref` in `Menu.IOptions` to provide greater control over exactly where in the DOM the menu `div` is attached.

I have kept the names and descriptions of the options the same as in `Widget.attach`.

### DIFF
--- a/packages/widgets/src/menu.ts
+++ b/packages/widgets/src/menu.ts
@@ -454,12 +454,14 @@ export class Menu extends Widget {
       return;
     }
 
-    // Extract the position options.
+    // Extract the menu options.
     let forceX = options.forceX || false;
     let forceY = options.forceY || false;
+    const host = options.host || null;
+    const ref = options.ref || null;
 
     // Open the menu as a root menu.
-    Private.openRootMenu(this, x, y, forceX, forceY);
+    Private.openRootMenu(this, x, y, forceX, forceY, host, ref);
 
     // Activate the menu to accept keyboard input.
     this.activate();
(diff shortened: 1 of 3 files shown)


packages/widgets/src/menu.ts
```diff
@@ -1535,7 +1552,9 @@ namespace Private {
     x: number,
     y: number,
     forceX: boolean,
-    forceY: boolean
+    forceY: boolean,
+    host: HTMLElement | null,
+    ref: HTMLElement | null
   ): void {
     // Get the current position and size of the main viewport.
     const windowData = getWindowData();
```
krassowski, line 1560: With the proposed changes the menu could end up beyond the boundaries of the element. This will be potentially problematic if the parent element (host) has clips the menu off. Is my thinking correct?

I think this is fine to merge as-is, I just wonder if this should be documented.

krassowski:
You will need to run `yarn run api` and commit changes.

fcollonval (approved):
Thanks @ianthomas23

FYI (not requiring code changes) it is better to used the non nullish operator `??` rather than the or operator.
The reason or is used a lot in lumino is that the code predates the introduction of that operator.

The goal is to avoid the case of a falsy value being overridden by the default like `0 || 1` != `0 ?? 1` (the first equals one when the second equals 0).

fcollonval:
> in addition to the API update, you will need to lint the code with `yarn lint`. One of the tests is failing due to

[warn] packages/widgets/tests/src/menu.spec.ts

ianthomas23:
Thanks for the guidance @krassowski and @fcollonval. I've added the extra documentation to `Menu.open`, although it could possibly be added to `IOpenOptions` instead, as that was my entry point into the docs. I've also switched the `||` to `??` in the new code whilst I was there.

ianthomas23:
I cannot tell if the failing `macos-latest` `webkit-headless` CI run is something that I have caused or if it is unrelated.

fcollonval:
> I cannot tell if the failing `macos-latest` `webkit-headless` CI run is something that I have caused or if it is unrelated.

This is due to flakiness - I let @krassowski have a second look.

fcollonval:
gentle ping to @krassowski

krassowski (approved):
Thanks!

Issue #586, opened 2023-05-14
Reuse the plugin system without depending on the Lumino Widgets


jtpio:
### Problem

The Lumino plugin system is simple and powerful. It could also be used for other purposes that don't necessarily provide UI elements to the interface.

However the Lumino `Application` currently requires a `T extends Widget = Widget` as the generic type for a shell:

https://github.com/jupyterlab/lumino/blob/613dc9df19652316a8e32c704d4937fcd43983cc/packages/application/src/index.ts#L138

### Proposed Solution

Make it possible to use the plugin and dependency injection system without a shell.

### Additional context

This was discussed with @echarles during JupyterCon 2023.

For reference JupyterLite uses a Lumino application without a `shell` by providing `never` as the generic type: https://github.com/jupyterlite/jupyterlite/blob/f98c5c0f90f25466aea6beeb4f2bd2ce8549aa68/packages/server/src/app.ts#L17

```ts
export class JupyterLiteServer extends Application<never>
```

jtpio:
@echarles if you have more examples or use cases feel free to add them here, thanks!

echarles:
Thx @jtpio for logging this. I can share for now https://github.com/datalayer-examples/jupyterlab-extensions-example/blob/main/test/lumino-inject.spec.ts which aims to showcase generic dependency injection based on Lumino. I will feed more there and document what could be done if we would abstract away from Widget.

fcollonval:
We could provide an raw `ABCApplication` than encapsulated only the plugin logic without command registry, context menu and shell.
Or simply template without restrictions the context menu and the shell.

PR #703, opened 2024-05-19
Extract plugin handling in a `PluginRegistry` class independent of the `Application`

### LINKED ISSUE #697: Refactor the plugins handling out of the `Application`
### Problem

The plugins are handled directly within the `Application`. This prevents the usage of plugins to initialize application attribute. The symptomatic case is the need to customize the service manager in JupyterLab.

Xref: 
- https://github.com/jupyterlab/jupyterlab/issues/15329
- https://github.com/jupyterlab/jupyterlab/pull/16141#issuecomment-2049120215

### Proposed Solution

Extracting the plugin resolver logic in its own class (because why should such great feat be hard linked to the Application). Once we get that the logic in index.js will be:

- Create the plugin resolver with all the plugins
- Provide that resolver to the lumino Application

> :warning: we will need to expose the resolver method in the Application to ensure backward compatibility

fcollonval:
Fixes #697 
Fixes https://github.com/jupyterlab/lumino/issues/586

Refactor the plugins handling outside of the `Application` class to be able to instantiate and use it independently.
Add a allowed/blocked list feature to the new class.

The new `PluginRegistry` class is defined in coreutils to avoid bringing the widgets package in the stack dependency of the plugin system consumer. That implies adding the algorithm package as dependency to coreutils.


packages/coreutils/src/plugins.ts
```diff
@@ -0,0 +1,867 @@
...
+   * plugins activation and deactivation functions.
+   *
+   * It can only be set once.
+   *
+   * By default, it is `null`.
+   */
+  get application(): T {
+    return this._application;
+  }
+  set application(v: any) {
```
brichet, line 145: Shouldn't we type the parameter here ?
```suggestion
  set application(v: T) {
```

packages/coreutils/src/typing.d.ts
brichet, line 1: Out of curiosity, why do we need this ?
Should it be defined in other packages using `console` ?

brichet (commented):
Thanks @fcollonval, LGTM, at least everything work in a dev installation.

I just have a few comments.

packages/coreutils/src/typing.d.ts
fcollonval: This package is compatible for browser and nodejs environment.

The automatic way to make TypeScript happy is to mention in the tsconfig that the target is `DOM`. However node js does not implement all the DOM API. But it does implement the console. So this pleases TypeScript but prevent future change to reference for example `window` or `document` (not available in node js).

packages/coreutils/src/plugins.ts
fcollonval: Indeed it makes sense

fcollonval:
Thanks for the review @brichet

brichet (approved):

PR #705, opened 2024-05-27
Make the `Application` fully composable

fcollonval:
This follows the extraction of the plugin logic #703 to make the `Application` fully composable.

It also reduce the constrain on the shell to be a raw html element.


packages/application/src/index.ts
```diff
@@ -39,7 +39,7 @@ export { type IPlugin };
  * UI applications with the ability to be safely extended by third
  * party code via plugins.
  */
-export class Application<T extends Widget = Widget> {
+export class Application<T extends Widget | HTMLElement = Widget> {
```
jtpio, line 42: Maybe it could be useful to detail the benefit for providing an `HTMLElement` directly, instead of creating a `new Widget()` with its `node` option?

jtpio: Thanks!

jtpio, line 42: Just to have the full picture here, and to provide more context for the future when looking back at that PR.

The main purpose for allowing an `HTMLElement` directly would be for apps reusing the Lumino plugin system to not have to depend on `@lumino/widgets` to instantiate a placeholder shell widget, for example like in the following example?

https://github.com/jupyterlab/lumino/blob/0fb17ae08476cb6c03c1dced1454391c8ea38355/packages/application/tests/src/index.spec.ts#L21-L24

fcollonval: Yes, although that the `ContextMenu` API is preventing to make `@lumino/widgets` an optional dependency;

https://github.com/jupyterlab/lumino/blob/0fb17ae08476cb6c03c1dced1454391c8ea38355/packages/application/src/index.ts#L70

PR #706, opened 2024-05-29
Interleaves search results across categories

JasonWeill:
Fixes https://github.com/jupyterlab/jupyterlab/issues/9659.

Allows search results in the command palette to be interleaved across groups.

Before the change, low-scoring search results are still grouped together, leaving higher-scoring search results in other groups lower:

Screenshot 2024-05-29 at 4 46 48 PM

After the change, search results are strictly sorted in descending score order. Headers for the same group may occur more than once. If two successive search results have the same header, then we don't display a new header.

Screenshot 2024-05-29 at 4 35 20 PM

### DIFF
--- a/packages/widgets/src/commandpalette.ts
+++ b/packages/widgets/src/commandpalette.ts
@@ -1386,7 +1386,7 @@ namespace Private {
       let category = item.category;
 
       // Is this the same category as the preceding result?
-      if (i === 0 || category !== scores[i-1].item.category) {
+      if (i === 0 || category !== scores[i - 1].item.category) {
         // Add the header result for the category.
         results.push({ type: 'header', category, indices: categoryIndices });
       }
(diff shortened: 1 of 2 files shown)


fcollonval:
Thanks @JasonWeill would you mind adding a test for this? And the linter is not happy :wink:

krassowski:
Could it be an option to have either one or the other behaviour?

Od note the headings are still shown in JupyterLab when using command palette in the sidebar and the old behaviour makes more sense there.

The PR title also suggest that this is an addition of an option rather than a change in behaviour

JasonWeill:
I've added a unit test specifically about `search`, using a similar command palette setup to one of the example files.

I don't think that this should be an option. Searching for commands by a keyword should return the highest-scoring results first. What is the value of displaying many low-score results in a large category, pushing down a high-scoring result from another, smaller, category?

JasonWeill:
@fcollonval Added a unit test for the search functionality. I'd appreciate if you could review when you get the time. Thanks!

fcollonval (approved):
Thanks @JasonWeill

PR #732, opened 2024-12-17
Add `horizontalAlignment` to `Menu.open()` options

### LINKED ISSUE #709: `Menu.open()` should have an option to align `x` and `y` to the top right of the menu.
### Problem

I am trying to add a button to an extension that opens a menu. The button is on the far right side of the window, so the menu is automatically shifting to fit into the window. If I use the `forceX` option a portion of the menu will be out of the window and not viewable.  

[See the thread here for more context.](https://github.com/nebari-dev/jupyterlab-new-launcher/pull/52#discussion_r1636112719)

### Proposed Solution

I propose adding a parameter to `Menu.IOpenOptions` to specify how the menu would relate to `x` and `y`. It could be something like `align` with the possible options `"right"` and `"left"`. `"left"` would be default and it would align the top-left of the menu to `x` and `y`. `"right"` would align the top-right of the menu to `x` and `y`.

andrewfulton9:
Closes #709 

Adds an "align" option to Menu.IOpenOptions that can be set to "left" or "right" and adjusts the menu position so the top left or right corner is matched to the given x and y coordinates based on the given align value. It defaults to left or right depending on the document `dir` attribute to account for the given language direction.

### DIFF
--- a/packages/widgets/src/menu.ts
+++ b/packages/widgets/src/menu.ts
@@ -465,9 +465,10 @@ export class Menu extends Widget {
     let forceY = options.forceY || false;
     const host = options.host ?? null;
     const ref = options.ref ?? null;
+    const align = options.align ?? 'left';
 
     // Open the menu as a root menu.
-    Private.openRootMenu(this, x, y, forceX, forceY, host, ref);
+    Private.openRootMenu(this, x, y, forceX, forceY, align, host, ref);
 
     // Activate the menu to accept keyboard input.
     this.activate();
@@ -1561,6 +1562,7 @@ namespace Private {
     y: number,
     forceX: boolean,
     forceY: boolean,
+    align: 'left' | 'right',
     host: HTMLElement | null,
     ref: HTMLElement | null
   ): void {
@@ -1590,6 +1592,13 @@ namespace Private {
 
     // Measure the size of the menu.
     let { width, height } = node.getBoundingClientRect();
+
+    // align the menu to the right corner of the host
+    if (align === 'right') {
+      x -= width;
+    }
+
+
 
     // Adjust the X position of the menu to fit on-screen.
     if (!forceX && x + width > px + cw) {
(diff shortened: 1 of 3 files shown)


krassowski:
Thank you @andrewfulton9!

1. To refresh `review/api/widgets.api.md` you can run `yarn run api` locally
2. The tests are failing on Mac/Safari due to rounding issues:

```
 ❌ @lumino/widgets > Menu > #open() > should accept align menu flags
      AssertionError: expected 'translate(205.53125px, 300px)' to equal 'translate(205.531px, 300px)'
      + expected - actual
      
      -translate(205.53125px, 300px)
      +translate(205.531px, 300px)
      
      at tests/lib/bundle.test.js:39356:59

 ❌ @lumino/widgets > Menu > #open() > align should default to right if language direction is rtl
      AssertionError: expected 'translate(205.53125px, 300px)' to equal 'translate(205.531px, 300px)'
      + expected - actual
      
      -translate(205.53125px, 300px)
      +translate(205.531px, 300px)   
```

Maybe we could use a regular expression pattern here, optionally allowing the number to have more decimal digits?

andrewfulton9:
@krassowski, I updated the tests so only the whole number of the x value is checked. It should be good to go now!

packages/widgets/src/menu.ts
```diff
@@ -1009,6 +1012,13 @@ export namespace Menu {
      * menu to be added as the last child of the host.
      */
     ref?: HTMLElement;
+
+    /**
+     * The alignment of the menu.
+     *
+     * The default is `'left'` unless the document `dir` attribute is `'rtl'`
+     */
+    align?: 'left' | 'right';
```
krassowski, line 1021: To make this more future-proof, should we call it `horizontalAlignment` instead? So that we can add `verticalAlignment` if we need it later?

Otherwise it looks great!

andrewfulton9: Good point! I just updated that.

krassowski (approved):
Thank you @andrewfulton9!

PR #733, opened 2024-12-20
Move MenuBar index check to an overridable method

### LINKED ISSUE #729: Allow empty menus to activate in menu bar
Hi! I'm working on [migrating to Lumino in Theia](https://github.com/eclipse-theia/theia/pull/14320). We’ve previously relied on PhosphorJS, and decided to transition to Lumino.

### Problem

We implemented a [DynamicMenuWidget](https://github.com/eclipse-theia/theia/blob/1a7486e492145c3d6f9aafbeca3b7e329eef8fe1/packages/core/src/browser/menu/browser-menu-plugin.ts#L244) whose core functionality involves dynamically [calculating its menu items just before being shown](https://github.com/eclipse-theia/theia/blob/1a7486e492145c3d6f9aafbeca3b7e329eef8fe1/packages/core/src/browser/menu/browser-menu-plugin.ts#L266-L274). This worked in PhosphorJS

However, we’ve encountered an issue with Lumino due to a change introduced in https://github.com/jupyterlab/lumino/pull/607. Specifically, [this check](https://github.com/jupyterlab/lumino/blame/204a845414f5f55ba4d97f95fed3ed3c3265490d/packages/widgets/src/menubar.ts#L148-L151) in the MenuBar prevents our menus from activating:

```ts
    // An empty menu cannot be active
    if (value > -1 && this._menus[value].items.length === 0) {
      value = -1;
    }
```

As a result, Theia’s dynamic menus are never activated and therefore never displayed.

### Workarounds

We’ve explored a few workarounds:

- Patching Lumino locally via `patch-package` to remove the check, but this requires maintenance with every new Lumino release.
- Subclassing MenuBar and overriding the `activeIndex` setter to bypass the check, but this requires duplicating the setter code and working around type checks on private fields.

Both approaches feel brittle and suboptimal for long-term maintenance.

### Proposed Solution?

Would you be open to changes that make the `MenuBar` more flexible for downstream adopters like us? If so, we’d appreciate your input on the best way forward. Some potential approaches:

- Removing the check entirely if it’s not critical for some other use cases
- Making private fields protected in `MenuBar` to allow subclassing without type workarounds.
- Moving the check into a separate method that subclasses can override.

Alternatively, if you have suggestions for a better way to implement dynamic menus in Theia that align more closely with Lumino's current design, I’d be very interested to hear your thoughts.

Thanks for all your work on Lumino—it’s a fantastic library, and we look forward to using it in Theia!

sdirix:
Added a protected isValidIndex method to the MenuBar class to handle index validation before setting an index.

This enables adopters to customize the validation logic, e.g., for supporting dynamic menus that may initially be empty.

Resolves #729

### DIFF
--- a/packages/widgets/src/menubar.ts
+++ b/packages/widgets/src/menubar.ts
@@ -140,13 +140,8 @@ export class MenuBar extends Widget {
    * If the menu cannot be activated, the index will be set to `-1`.
    */
   set activeIndex(value: number) {
-    // Adjust the value for an out of range index.
-    if (value < 0 || value >= this._menus.length) {
-      value = -1;
-    }
-
-    // An empty menu cannot be active
-    if (value > -1 && this._menus[value].items.length === 0) {
+    // Adjust the value for an invalid index.
+    if (!this.isValidIndex(value)) {
       value = -1;
     }
 
@@ -160,6 +155,16 @@ export class MenuBar extends Widget {
 
     // Schedule an update of the items.
     this.update();
+  }
+
+  /**
+   * Before setting a new index, this method is used to validate it.
+   * 
+   * By default it checks whether the index is within menu range
+   * and whether the corresponding menu has at least one item.
+   */
+  protected isValidIndex(index: number): boolean {
+    return index >= 0 && index < this._menus.length && this._menus[index].items.length > 0;
   }
 
   /**
--- a/packages/widgets/src/menubar.ts
+++ b/packages/widgets/src/menubar.ts
@@ -159,12 +159,16 @@ export class MenuBar extends Widget {
 
   /**
    * Before setting a new index, this method is used to validate it.
-   * 
+   *
    * By default it checks whether the index is within menu range
    * and whether the corresponding menu has at least one item.
    */
   protected isValidIndex(index: number): boolean {
-    return index >= 0 && index < this._menus.length && this._menus[index].items.length > 0;
+    return (
+      index >= 0 &&
+      index < this._menus.length &&
+      this._menus[index].items.length > 0
+    );
   }
 
   /**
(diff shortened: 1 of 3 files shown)


sdirix:
Let me know whether I can do anything to help resolve the errors in the CI

sdirix:
@krassowski I pushed the linter fix as a separate commit. If you prefer squashed commits, I can also happily do that

krassowski:
Separate commits work well, we can squash on merge.

To fix `Does PR have API changes? / api-changes` you will need to refresh `review/api/widgets.api.md` by running `yarn run api` and committing changes.

I am not sure why `Fix License Headers / header-license-fix` is failing, it appears unrelated.

krassowski (approved):
This looks good to me, thank you @sdirix!

sdirix:
@krassowski Please wait with the merge. While this fix enables us to build our menus, I just realized that there are some usability edge cases remaining, e.g. the menus can be opened but do not react on hover.

I had a quick glance at the code base: There are more places in which `items.length` are checked, which could be the culprit, so this PR is overall incomplete. If we settle on a different approach to tackle the issues, then the menubar was "enhanced" without a real benefit to anyone.

I'll come back to you. Sorry for the inconvenience!

sdirix:
I will close the PR for now until we come up with a better solution . As described above, this PR is not sufficient to support empty menus properly and therefore neither improves the code base nor helps adopters like us. Instead some more changes are necessary . See for example [here](https://github.com/jupyterlab/lumino/blob/8a4c5f59ed53bdfa1cedb0eb5641ba705a176a63/packages/widgets/src/menubar.ts#L435) and [here](https://github.com/jupyterlab/lumino/blob/8a4c5f59ed53bdfa1cedb0eb5641ba705a176a63/packages/widgets/src/menubar.ts#L505) for more hard coded checks.

For now we use a workaround: We always add at minimum one dummy element into all menus to bypass the Lumino checks. Once the (sub) menu is about to be shown, we replace the element with the real elements. This works as we don't have empty top level menus.

Once we come up with a more elegant solution I'll come back to you. Thanks for your patience :)

jtpio:
Alright, thanks @sdirix for the summary and spending time on this!

PR #743, opened 2025-03-02
"npm run" -> yarn

vidartf:
If we are using yarn, then it should be everywhere. Without this, binary resolution fails on Windows, e.g. `rimraf` will fail to find its main entry point.


packages/coreutils/package.json
```diff
@@ -64,5 +63,6 @@
   },
   "publishConfig": {
     "access": "public"
-  }
+  },
+  "module-node": "dist/index.node.es6"
```
fcollonval, line 67: Could we keep `module-node` close to the other output configuration to ease configuration maintenance?

fcollonval (changes requested):
Thanks @vidartf 

I have a minor suggestion. But otherwise LGTM.

packages/coreutils/package.json
vidartf: I didn't do any of the sorting manually, but something did it automatically at some point. I can't seem to get it to reproduce this though. Maybe this was an old config of the linter that was pulled in automatically by git when rebasing some old code? 🤔  If the linters are no longer set up to sort things, I will undo the sorting.

fcollonval (approved):
Thanks @vidartf

PR #745, opened 2025-03-07
Forbid .only and .skip on CI

fcollonval:
This is a follow-up of the error pushed as part of #732


krassowski (approved):
Thank you @fcollonval!

PR #746, opened 2025-03-07
Don't refresh if palette is within a hidden branch

### LINKED ISSUE #744: CommandPalette reacts to notifyCommandChanged even when hidden
## Description

The command palette is rerendering on `notifyCommandChanged` even if it is hidden. This triggers a lot of command attribute request for nothing as can be seen by the following warnings:

Image

> This is with JupyterLab 4.3.5 with extensions git, lsp and spellchecker activated.

## Reproduce

1. Install jupyterlab-git, jupyterlab-lsp and jupyterlab-spellchecker extensions
2. Open a notebook in JupyterLab and look at the console log

> Note the screenshot was taken after running `jupyter lab build --no-minize` to introspect the core code.

## Expected behavior

The command palette is not rerender when hidden.

## Context

Sequence flows:

- notifyCommandChanged is emitted
- [_onGenericChange](https://github.com/jupyterlab/lumino/blob/9f8755eab298acd83b98e022f615688d1e51cd07/packages/widgets/src/commandpalette.ts#L44) is invoked
- [refresh](https://github.com/jupyterlab/lumino/blob/9f8755eab298acd83b98e022f615688d1e51cd07/packages/widgets/src/commandpalette.ts#L522) is called
- [Widget.update](https://github.com/jupyterlab/lumino/blob/9f8755eab298acd83b98e022f615688d1e51cd07/packages/widgets/src/commandpalette.ts#L222) is called

fcollonval:
Fixes #744

- Add new method `Widget.isWithinHiddenWidget` to test if a widget is hidden or within an hidden parent

> I'm not happy with the name `isWithinHiddenWidget` nor with `isAnyHidden`. Happy to change it to something else; I thought about `isAnyParentHidden`

I would recommend merging #745 prior to this to ensure tests are passing on CI.

---

This PR has two side effects:
- commenting a left over skipped test from #745 - the skipped test are marked as failing when run in debug mode but not otherwise it seems. I have no ideas why this is happening.
- I turn on the generation of source maps for tests; this makes debugging much more efficient.

### DIFF
--- a/packages/widgets/src/widget.ts
+++ b/packages/widgets/src/widget.ts
@@ -882,7 +882,7 @@ export namespace Widget {
      * The widget is visible.
      *
      * @deprecated since 2.7.0, apply that flag consistently was not reliable
-     * so it was drop in favor of a recursive check of the visibility of all parents.
+     * so it was dropped in favor of a recursive check of the visibility of all parents.
      */
     IsVisible = 0x8,
 
--- a/packages/widgets/src/widget.ts
+++ b/packages/widgets/src/widget.ts
@@ -112,6 +112,10 @@ export class Widget implements IMessageHandler, IObservableDisposable {
 
   /**
    * Test whether the widget is explicitly hidden.
+   *
+   * #### Notes
+   * You should prefer `!{@link isVisible}` over `{@link isHidden}` if you want to know if the
+   * widget is hidden as this does not test if the widget is hidden because one of its ancestors is hidden.
    */
   get isHidden(): boolean {
     return this.testFlag(Widget.Flag.IsHidden);
@@ -123,9 +127,20 @@ export class Widget implements IMessageHandler, IObservableDisposable {
    * #### Notes
    * A widget is visible when it is attached to the DOM, is not
    * explicitly hidden, and has no explicitly hidden ancestors.
+   *
+   * Since 2.7.0, this does not rely on the {@link Widget.Flag.IsVisible} flag.
+   * It recursively checks the visibility of all parent widgets.
    */
   get isVisible(): boolean {
-    return this.testFlag(Widget.Flag.IsVisible);
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    let parent: Widget | null = this;
+    do {
+      if (parent.isHidden || !parent.isAttached) {
+        return false;
+      }
+      parent = parent.parent;
+    } while (parent != null);
+    return true;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


krassowski:
> I would recommend merging https://github.com/jupyterlab/lumino/pull/745 prior to this to ensure tests are passing on CI.

I just merged #745.

> I'm not happy with the name `isWithinHiddenWidget` nor with `isAnyHidden`. Happy to change it to something else

Sounds good. Some food for thought: in Qt, `isVisible()` and `isHidden()` are not exact equivalents. `isHidden == true` means that `isVisible == false`, but `isVisible == false` does not mean that the widget `isHidden` (because it may not be visible due to the parent being hidden, or due to being detached). `isHidden` returns per-widget state ignoring parents, `isVisible` returns the actual visibility (ignoring it being obscured by other windows etc).

> If an ancestor is not visible, the widget won't become visible until all its ancestors are shown

> A hidden widget will only become visible when `show()` is called on it. It will not be automatically shown when the parent is shown.
>
> To check visibility, use `!isVisible()` instead (notice the exclamation mark).
> 
> `isHidden()` implies `!isVisible()`, but a widget can be not visible and not hidden at the same time. This is the case for widgets that are children of widgets that are not visible.

https://doc.qt.io/qt-5/qwidget.html#isHidden

>  `isAnyParentHidden`

I think this name is fine - unless we want to mirror Qt way of doing things.

One advantage of doing it Qt way is that we could also encapsulate "is any parent not attached to the DOM" in the same logic. What do you think?

fcollonval:
Thanks @krassowski for all the thoughts and ideas. And actually as PhosphorJS mimick Qt API, this is indeed what `isVisible` doc mentions...

https://github.com/jupyterlab/lumino/blob/72519571202da304be5b9bf3824ad535020af126/packages/widgets/src/widget.ts#L120-L127

So in [e40bd13](https://github.com/jupyterlab/lumino/pull/746/commits/e40bd13186c1b18a01b1ec31f5edac01e28de918), I removed the new API to instead change the behavior of `isVisible` and I deprecated the flag `Widget.Flag.isVisible` as this is not reliable.

> I gave a shot at adding a default layout that I thought would solve the issue. But it did not and more importantly it was having much more side effects.

packages/widgets/src/widget.ts
```diff
@@ -856,6 +880,9 @@ export namespace Widget {
 
     /**
      * The widget is visible.
+     *
+     * @deprecated since 2.7.0, apply that flag consistently was not reliable
+     * so it was drop in favor of a recursive check of the visibility of all parents.
```
fcollonval, line 885:
```suggestion
     * so it was dropped in favor of a recursive check of the visibility of all parents.
```

krassowski (approved):
Thank you @fcollonval!

PR #753, opened 2025-04-28
fix: improve accessibility for menu items with toggle state

nkn2022:
Here is a proposal on how to fix the issue: https://github.com/jupyterlab/jupyterlab/issues/17501
This was found when working on another accessibility issue. PR: https://github.com/jupyterlab/jupyterlab/pull/17472

References used:
https://www.w3.org/TR/wai-aria/states_and_properties#introstates
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menuitemcheckbox_role

@krassowski ^

### DIFF
--- a/packages/widgets/src/menu.ts
+++ b/packages/widgets/src/menu.ts
@@ -1371,7 +1371,7 @@ export namespace Menu {
           }
           if (data.item.isToggled) {
             aria.role = 'menuitemcheckbox';
-            aria['aria-checked'] = data.item.isToggled.toString();
+            aria['aria-checked'] = 'true';
           } else {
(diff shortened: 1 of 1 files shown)


krassowski (approved):
I think ideally the role for menu entries which can be toggled would always have `menuitemcheckbox` role, not only when they are toggled. I appreciate it is not trivial to achieve with the information available on menu items right now so it would require a larger change. I think this PR is a significant improvement as-is though.

The workflow failures are unrelated and will be fixed with https://github.com/jupyterlab/lumino/pull/754

nkn2022:
Thank you, @krassowski.

PR #754, opened 2025-04-29
Run workflows that were pinned to no-longer supported `ubuntu-20.04` on `ubuntu-latest`

krassowski:
Ubuntu 20.04 is no longer supported on GH Actions, see https://github.com/actions/runner-images/issues/11101

These two workflows are currently failing to run on PRs:

image

The codeql template recommends using `ubuntu-latest` anyways.


krassowski:
I suspect the `header-license-fix` will stop failing after merge to `main`:

image

PR #777, opened 2026-01-14
Fix: removing a non-existent menu item should be a no-op

jasongrout:
Currently, removing a menu item that does not exist in the menu accidentally removes the last item in the menu. This is because the current code:

```ts
  /**
   * Remove an item from the menu.
   *
   * @param item - The item to remove from the menu.
   *
   * #### Notes
   * This is a no-op if the item is not in the menu.
   */
  removeItem(item: Menu.IItem): void {
    this.removeItemAt(this._items.indexOf(item));
  }
```
uses indexOf to find a menu item. If the menu item is not found, indexOf returns -1, but -1 is interpreted by removeItemAt as the last item. Instead we should check the return value of indexOf.

We're seeing this cause a bug in JupyterLab: https://github.com/jupyterlab/jupyterlab/issues/18199

### DIFF
--- a/packages/widgets/src/menu.ts
+++ b/packages/widgets/src/menu.ts
@@ -376,7 +376,11 @@ export class Menu extends Widget {
    * This is a no-op if the item is not in the menu.
    */
   removeItem(item: Menu.IItem): void {
-    this.removeItemAt(this._items.indexOf(item));
+    let index = this._items.indexOf(item);
+    if (index === -1) {
+      return;
+    }
+    this.removeItemAt(index);
   }
 
   /**


jasongrout:
As expected, the test that we add in the first commit failed:

```
 ❌ @lumino/widgets > Menu > #removeItem() > should be a no-op if the item is not in the menu
      AssertionError: expected +0 to equal 1
      + expected - actual
      
      -0
      +1
      
      at n.<anonymous> (tests\lib\bundle.test.js:39411:46)
```
I'll add the fix in the second commit.

Issue #760, opened 2025-05-28
Split handle dragging does not work with embedded frames


tsmaeder:
## Description
When I embed an `iframe` inside a widget and put that widget inside either a `SplitPanel` or `DockPanel`, dragging the split handles to change the widget sizes does not work anymore as soon as you drag the handle quickly over the embedded frame. I've created a modified dock panel example to show the issue (attached below)

## Reproduce

1. `yarn & yarn build` the attached example
2. Open `index.html`
3. Observe: you see a modified dock panel example with the biggest "squares" being blue and red
4. Very slowly drag the split handle between the red and blue squares 
5. Observe: the dragging works and resizes the left and right sides
6. Now drag the handle quickly over the red or right square
7. Observe: the left and right sides of the split are no longer resized. 

## Expected behavior
The resizing works like it does in the original example

## Context
Windows 11, Firefox

tsmaeder:
And here's a video: 

https://github.com/user-attachments/assets/76e3d957-9b9f-4e1e-9342-9e0656c199b8

PR #782, opened 2026-01-24
Capture the pointer for split panel drags

jasongrout:
In https://github.com/jupyterlab/lumino/issues/760, it was reported that dragging a handle near an iframe can lose the drag events when the mouse goes over the iframe. This attempts to use the pointer capture semantics to ensure that the drag events continue to go to the split panel.

The thinking behind capturing the pointer events is that if the split panel or dock panel captures the pointer events, then those events will cascade through the document and trigger the event handling for pointermove/pointerup/etc. that we are setting up to handle the move. If we don't capture the pointer events, then those events potentially can leave the document (like if we are over an iframe), and we don't get our document-level handlers triggered.

## Testing

Based on the example provided in #760, apply this patch to modify the dockpanel example to use iframes:

```diff
diff --git a/examples/example-dockpanel/src/index.ts b/examples/example-dockpanel/src/index.ts
index 7613dd54..ec2c2d54 100644
--- a/examples/example-dockpanel/src/index.ts
+++ b/examples/example-dockpanel/src/index.ts
@@ -69,9 +69,11 @@ class ContentWidget extends Widget {
   static createNode(): HTMLElement {
     let node = document.createElement('div');
     let content = document.createElement('div');
-    let input = document.createElement('input');
-    input.placeholder = 'Placeholder...';
-    content.appendChild(input);
+    let frame = document.createElement('iframe');
+    frame.style.width = '100%';
+    frame.style.height = '100%';
+    frame.srcdoc = '<!DOCTYPE html><html><head><meta charset="UTF-8"></head><body>Iframe</body></html>';
+    content.appendChild(frame);
     node.appendChild(content);
     return node;
   }
```
and compile and open the dockpanel example.

Before: choppy, often loses events, which means it loses the mouseup as well

https://github.com/user-attachments/assets/dc4d725c-6d5c-4091-8937-0cc92a302039

After: buttery smooth, split stays with the cursor without fail:

https://github.com/user-attachments/assets/97eee5eb-f9f8-457a-a503-8f45dca5e256

### DIFF
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -721,6 +721,17 @@ export class DockPanel extends Widget {
     this._document.addEventListener('pointermove', this, true);
     this._document.addEventListener('contextmenu', this, true);
 
+    // Set the pointer capture to this split panel, ensuring the document
+    // listeners above will be triggered even if the pointer drifts over another
+    // document, like an iframe.
+    let pointerId = event.pointerId;
+    try {
+      this.node.setPointerCapture(pointerId);
+    } catch (e) {
+      // Pointer capture is nice to try, but may fail. In that case, ignore the
+      // failure.
+    }
+
     // Compute the offset deltas for the handle press.
     let rect = handle.getBoundingClientRect();
     let deltaX = event.clientX - rect.left;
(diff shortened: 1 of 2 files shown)


jasongrout:
When I run `yarn run test:firefox` on macos, the tests pass for me. I'm not sure why they are failing in CI.

jasongrout:
When I reverted back to main, macos:firefox tests passed. Let's go back to the PR and see what happens.

jasongrout:
@krassowski - the macos:firefox tests passed again. I don't see an obvious connection between this PR and those tests that seemed to fail, and it seemed similar tests failed on other PRs. So I don't think the tests failures are related to this PR.

PR #784, opened 2026-02-02
Capture pointer events to the cursor override div instead of the dockpanel or splitpanel

jasongrout:
This reverts changes in https://github.com/jupyterlab/lumino/pull/782 which captured pointer events to the dockpanel or splitpanel, which broke the cursor override. Apparently browser implementations shortcut the hit test for pointer events and cursor styles when the pointer is captured, so browsers were ignoring the cursor style in the cursor override div and instead using the cursor style for the element that did pointer capture.
    
The cursor override centers a small div under the cursor to ensure that the cursor is overridden irrespective of what else is under the cursor (and under the small override div). As such, pointer events usually go to the cursor override div anyway instead of the underlying elements. The pointer capture on the override div reinforces this pointer event capture that was already naturally happening.
    
This follows up on work on the cursor override div from https://github.com/jupyterlab/lumino/pull/502 and https://github.com/jupyterlab/lumino/pull/564.

Claude Code (Opus 4.5) assisted in this PR.

### DIFF
--- a/packages/dragdrop/src/index.ts
+++ b/packages/dragdrop/src/index.ts
@@ -1280,6 +1280,8 @@ namespace Private {
           propagateBackdropScroll,
           true
         );
+
+        // Removing the element from the DOM also releases the pointer capture.
         body.removeChild(cursorBackdrop);
       }
     });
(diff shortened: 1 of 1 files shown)


jasongrout:
I rebased on #787 - let's see if that fixes the flaky tests here.

jasongrout:
Woohoo, macos-firefox passed on the first run! Mike, I think this is ready for final review / merging.

PR #793, opened 2026-02-21
Keep tab widths frozen when closing tabs

### LINKED ISSUE #10514: Keep size of Titlebar when pressing "x".
https://user-images.githubusercontent.com/44469195/124001368-d84ab180-d9d4-11eb-95cf-7812e5c7fcc2.mov

### Problem

I would like it if the "x" Button would not change position when having lots of open notebooks. Currently, it is moving (see movie).

### Proposed Solution

The Titlebar could only expend, when the mouse is moving, as it is the case in Google Chrome (see second part of the video)

MUFFANUJ:
fixes https://github.com/jupyterlab/jupyterlab/issues/10514 

This pull request brings Chrome-like tab closing behavior to the TabBar. When a user clicks the close icon (x) on a tab, the remaining tabs temporarily maintain their current widths rather than immediately reflowing to fill the empty space. This ensures that the next tab's close button lands directly under the user's cursor. Once the user moves their cursor away from the tab bar, the tabs smoothly transition back to their natural sizes.

### Before
https://github.com/user-attachments/assets/347205a7-6494-49ca-9229-306b7d4c1e0b

### After
https://github.com/user-attachments/assets/d319f918-434e-48fb-bccf-5bcb8aab98c6

### DIFF
--- a/packages/widgets/src/tabbar.ts
+++ b/packages/widgets/src/tabbar.ts
@@ -1406,14 +1406,21 @@ export class TabBar<T> extends Widget {
 
     // Clear the inline width on all tabs, triggering the CSS transition.
     let tabs = this.contentNode.children;
+    if (tabs.length === 0) {
+      this.removeClass('lm-mod-unfreezing');
+    } else {
+      const onTransitionEnd = (event: Event) => {
+        if ((event as TransitionEvent).propertyName === 'width') {
+          this.removeClass('lm-mod-unfreezing');
+          this.node.removeEventListener('transitionend', onTransitionEnd);
+        }
+      };
+      this.node.addEventListener('transitionend', onTransitionEnd);
+    }
+
     for (let i = 0, n = tabs.length; i < n; ++i) {
       (tabs[i] as HTMLElement).style.width = '';
     }
-
-    // Remove the unfreezing class after the transition completes.
-    setTimeout(() => {
-      this.removeClass('lm-mod-unfreezing');
-    }, 150);
 
     // Schedule an update to re-render the tabs at natural size.
     this.update();
--- a/packages/widgets/src/tabbar.ts
+++ b/packages/widgets/src/tabbar.ts
@@ -502,6 +502,17 @@ export class TabBar<T> extends Widget {
    * This is a no-op if the index is out of range.
    */
   removeTabAt(index: number): void {
+    if (this._tabSizeFrozen) {
+      // Capture the widths of all tabs that will remain.
+      let tabs = this.contentNode.children;
+      this._frozenTabWidths = [];
+      for (let i = 0, n = tabs.length; i < n; ++i) {
+        if (i !== index) {
+          this._frozenTabWidths.push((tabs[i] as HTMLElement).offsetWidth);
+        }
+      }
+    }
+
     // Release the mouse before making any changes.
     this._releaseMouse();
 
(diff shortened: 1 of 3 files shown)


packages/widgets/src/tabbar.ts
```diff
@@ -1349,6 +1385,40 @@ export class TabBar<T> extends Widget {
...
+    // Clear the inline width on all tabs, triggering the CSS transition.
+    let tabs = this.contentNode.children;
+    for (let i = 0, n = tabs.length; i < n; ++i) {
+      (tabs[i] as HTMLElement).style.width = '';
+    }
+
+    // Remove the unfreezing class after the transition completes.
+    setTimeout(() => {
+      this.removeClass('lm-mod-unfreezing');
+    }, 150);
```
krassowski: That's a minor issue but this introduces a tight coupling between CSS and JS - downstreams cannot increase transition speed with CSS alone. I could see this either way:

- a) we drop the class and just set the style via JS; this is slightly more performant than adding/removing classes; maybe we could make the timeout configurable; the downside is we need to add new config options to public API on JS side.
- b) we use [`transitionend`](https://developer.mozilla.org/en-US/docs/Web/API/Element/transitionend_event) event to remove the class once transition completed, whatever time it was set to.
- c) we don't special case this and just have the transition always enabled

krassowski: I would probably prefer (b) it feels more aligned with the `lumino` design but I wonder if other folks have thoughts on it.

packages/widgets/src/tabbar.ts
```diff
@@ -502,6 +502,17 @@ export class TabBar<T> extends Widget {
    * This is a no-op if the index is out of range.
    */
   removeTabAt(index: number): void {
+    if (this._tabSizeFrozen) {
```
krassowski: What is user closed tabs using keyboard shortcuts? There will be no `pointerleave` and the widths won't get unfrozen, right? I think we should check if mouse click was used to close the tab first. This would be a good test case to add btw.

krassowski (commented):
I like this, I think this is quite desirable. I see that both Firefox and Chromium based browsers implement this, as do tab bars in GNOME/GTK based applications (I think that's [`Adw.TabBar`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.2/class.TabBar.html)).

While right now I don't see a reason why someone would want to opt-out of this behaviour\*, we could consider making it customizable the way [`removeBehavior` is](https://github.com/jupyterlab/lumino/blob/4997384b24d5a5f6baece4f712f87221e2b015d1/packages/widgets/tests/src/tabbar.spec.ts#L730-L778)? Or maybe it's better kept to a future PR if someone actually asks for a way to opt-out so that we don't expand API unnecessarily.

krassowski (commented):
Note: 150ms looks good to me, though I will note that both Chrome and GTK (adw) use 200ms.

krassowski (commented):
It looks good on preview, I tagged copilot for re-review

packages/widgets/tests/src/tabbar.spec.ts
```diff
@@ -1039,6 +1054,106 @@ describe('@lumino/widgets', () => {
           });
         });
       });
+
+      it('should not freeze tab widths if closed programmatically (e.g., keyboard shortcut)', () => {
+        populateBar(bar);
+        MessageLoop.sendMessage(bar, Widget.Msg.UpdateRequest);
+        let tabs = bar.contentNode.children;
```
krassowski: Highlighting a warning check:

```
The initial value of tabs is unused, since it is always overwritten.
```

krassowski (approved):
Thank you @MUFFANUJ!

MUFFANUJ:
Can we merge this? @krassowski

Issue #798, opened 2026-03-13
Support resizable Groups


emilkrebs:
### Problem

Resizing multiple widgets at a single intersection currently requires individual boundary adjustments. This is inefficient for complex layouts (e.g., 4-way splits) where a user wants to shift the focal point of the entire workspace in one motion.

### Proposed Solution

Implement **multi-axis resizing** at widget intersection points. When a user hovers over or drags the "cross-junction" where multiple panels meet, all adjacent widgets should resize simultaneously to maintain layout integrity, similar to the behavior in VS Code.

### Additional context

This functionality is a standard in modern IDEs and layout libraries:

* **Reference:** https://react-resizable-panels.vercel.app/examples/nested-groups
* **Screenshot from VSCode:**

<img width="361" height="285" alt="Image" src="https://github.com/user-attachments/assets/527caa45-981d-4061-b591-34166d8c9435" />

---

This issue has been originally noticed in:
https://github.com/eclipse-theia/theia/issues/17155

PR #799, opened 2026-03-13
Support resizing three panels at corners (resizable groups)

emilkrebs:
**Disclaimer**:
This PR has been mostly written by Claude Sonnet 4.6 and Opus ~4.6~ 4.8 since I do not have a clear overview over the codebase. Human critical thinking skills still apply :) 

It works pretty well I think and I desperately want this feature in Eclipse Theia for a more modern IDE expierence.
I will take some time to review the PR myself but since I am not an expert here feel free to drop your opinion on this :)

Thanks a lot! 

**Fixes:** 

https://github.com/jupyterlab/lumino/issues/798


krassowski (commented):
On https://lumino--799.org.readthedocs.build/en/799/examples/dockpanel/index.html preview it looks like it can work, but it's really hard do know when it will because the cursor does not change to 4-way arrow when on hover.

IMO there ought to be a handle in DOM or some specific pixel size of a corner that you can grab otherwise it will be unintuitive and hard to use.

packages/widgets/src/dockpanel.ts
```diff
@@ -477,6 +482,8 @@ export class DockPanel extends Widget {
     this.node.addEventListener('lm-dragover', this);
     this.node.addEventListener('lm-drop', this);
     this.node.addEventListener('pointerdown', this);
+    this.node.addEventListener('pointermove', this);
+    this.node.addEventListener('pointerleave', this);
```
krassowski, line 486: I am somewhat worried about performance impact of constantly listening to `pointermove` on the entire widget. Can we instead only add listeners on the layout handles objects?

krassowski (commented):
It looks great on the example preview. I am somewhat worried about potential performance impact here, see my comment. I am not saying that we have to move the listener to handles, but I would suggest trying that and if that is not feasible considering if some of the hit tests can be cached.

emilkrebs:
> It looks great on the example preview. I am somewhat worried about potential performance impact here, see my comment. I am not saying that we have to move the listener to handles, but I would suggest trying that and if that is not feasible considering if some of the hit tests can be cached.

Good point! I have actually tested the performance using the chrome devtools snd noticed no significant increase in the before and after. However its definitely a good suggestion since I noticed a significant performance drop in both instances when resizing panels, might make sense to think about that in the future 🤔

emilkrebs:
Any review updates on this? Would love to finally have this feature in Theia. What can I do to help get this done?

krassowski:
I think my earlier concern around performance still stands. I benchmarked the performance before and after with help from Claude and it does look like a noticeable regression:

> On a 24-widget `DockPanel` (46 handles), 2000 `pointermove` dispatches, with and without the new node-level listener:
> 
> ```
> over content: ON=34.5ms  OFF=3.8ms      (~15µs added per event)
> over handle : ON= 9.2ms  OFF=4.5ms
> ```
> 
> ~9× on the common case (pointer over content, not a handle), because `find(layout.handles(), …)` must walk all handles before concluding "not a handle". It's linear in handle count, and it compounds: `pointermove` bubbles, so every ancestor `SplitPanel` runs its own scan on top of the `DockPanel`'s.
> 
> The benchmark also _understates_ the handle-hover case: nothing mutates the DOM between events there, so the `getBoundingClientRect()` loop in `findIntersectingHandle` hits cached `rects`. In a real app with concurrent DOM writes, that's up to 46 forced synchronous layouts per `pointermove`.
> 
> Three options, cheapest first:
> 
> 1. Fast bail before the scan when `event.target` isn't a handle - this removes the dominant cost entirely. A `Set<HTMLDivElement>` of live handles kept by the panel would be renderer-agnostic (a `classList` check would break custom renderers).
> 2. Coalesce the hit-test with `requestAnimationFrame` so at most one runs per frame regardless of pointer event rate.
> @krassowski's original suggestion - listeners on the handle elements themselves - is the structurally correct fix and makes the content-hover cost zero, at the price of syncing listeners as the layout changes.
> 
> (1) + (2) is probably the best effort/benefit trade here.

34.5ms makes work borderline unusable. 24 widgets is on the high end but not rare.

packages/widgets/src/dockpanel.ts
```diff
@@ -1559,7 +1640,7 @@ namespace Private {
   });
 
   /**
-   * Create a single document config for the widgets in a dock panel.
+  as * Create a single document config for the widgets in a dock panel.
```
krassowski:
```suggestion
  * Create a single document config for the widgets in a dock panel.
```

emilkrebs:
Okay, thanks a lot I will dig into it!

emilkrebs:
@krassowski okay, I investigated this issue further with claude and the improvements you have mentioned had a major impact on performance, especially by reducing DOM-Load due to preventing unessecary updates and calculating per frame change.
I included the benchmarking tool claude has written so you can test it out yourself, I would remove it before merge tho.

## Results

Group resize, nested 9-pane x 3000 nodes, 60 moves, 2 runs per side:
  | before | after
-- | -- | --
LayoutCount | 121 | 61
LayoutDuration | 3532 ms | 2426 ms (−30%)
fps | 11.0 | 14.0 (+26%)
frame p95 | 108 ms | 84 ms (−19%)
RecalcStyleCount | 124 | 64
forced reflow inside JS | 8.8 ms mean / update | gone (0.02 ms)

## Benchmarking

`perf/` contains the harness these numbers come from. **It is not part of the
library and should be removed before merge** — it's here so the results can be
reproduced independently.

### Reproducing

```bash
cd packages/widgets && npx tsc --build && ../../node_modules/.bin/rollup -c
node perf/run.mjs --mode nested --n 9 --dom 3000 --at intersect --steps 60
```

### Modes 
 | 
-- | --
?mode=stress | 24 panes, 800 DOM nodes each — feels smooth (~64 fps measured)
?mode=stress&dom=2000 | 54k DOM nodes — visibly hitchy (~36 fps), each move a ~95ms long task
?mode=stress&n=40 | more panes
?mode=nested, ?mode=dock, ?mode=split | the focused measurement scenes


-> even with the 24 panes the application is still usable

packages/widgets/style/splitpanel.css
```diff
@@ -42,6 +42,10 @@
   cursor: ns-resize;
 }
 
+.lm-SplitPanel-handle.lm-mod-intersection {
```
krassowski, line 45: This rule loses to the other cursor rules; the specificity of the rules above is higher than in dock panel (see that these include additional `> .lm-SplitPanel-handle`)

```suggestion
.lm-SplitPanel > .lm-SplitPanel-handle.lm-mod-intersection {
```

should give us enough specificity to win over the rules above.

packages/widgets/src/splitpanel.ts
```diff
@@ -340,8 +451,92 @@ export class SplitPanel extends Panel {
...
+          continue;
+        }
+        const r = h.getBoundingClientRect();
+        const lo = layout.orientation === 'horizontal' ? r.top : r.left;
+        const hi = layout.orientation === 'horizontal' ? r.bottom : r.right;
+        if (crossPos >= lo - tolerance && crossPos <= hi + tolerance) {
+          return {
+            panel: candidate,
+            index: i,
+            delta: Math.max(0, crossPos - lo)
```
krassowski, line 492: The `Math.max(0, crossPos - lo)` discards a legitimately negative offset. If you press inside the tolerance band but above the inner handle, the offset is thrown away and the handle snaps to the cursor on the first move. `DockPanel`'s equivalent (`intersect.deltaX`/`deltaY`) doesn't clamp and is correct.

packages/widgets/src/docklayout.ts
```diff
@@ -253,6 +255,150 @@ export class DockLayout extends Layout {
...
+        right = rect.right;
+      let top = rect.top,
+        bottom = rect.bottom;
+      if (primaryOrientation === 'vertical') {
+        top -= tol;
+        bottom += tol;
+      } else {
+        left -= tol;
+        right += tol;
+      }
```
krassowski, line 315: The expansion goes only on the primary's movement axis, which is the candidate's long axis - that bridges the parent/child handle gap (necessary), but it leaves the candidate's thin axis with no tolerance at all. So the grab target ends up being just the handle thickness. Is it intended?

The issue is mainly that the `SplitPanel` behaves differently. Would it make sense to align the two?

packages/widgets/src/dockpanel.ts
```diff
@@ -722,37 +734,153 @@ export class DockPanel extends Widget {
...
+  private _updateIntersectionHover(event: PointerEvent): void {
+    const handle = this._hitTestHandle(event.target);
+    const peer = handle
+      ? (this.layout as DockLayout).findIntersectingHandle(
+          handle,
+          event.clientX,
+          event.clientY
+        )
+      : null;
+    this._setIntersectionHoverHandle(peer ? handle : null, peer);
```
krassowski, line 861: The drag path got rAF coalescing but the hover path didn't, and this is the one that reads geometry. `findIntersectingHandle`'s first `getBoundingClientRect` forces a synchronous document layout on every `pointermove` whenever anything else has dirtied the layout.

This is one forced layout per event, not one per handle. But sliding slowly along a divider hunting for the corner forces a full-document layout at pointer-event rate. Same rAF treatment as `_applyDrag`, or caching the handle rects for the duration of a hover, would close it.

krassowski (commented):
Thank you for addressing the main performance concern. Re-running the local performance test, it looks much better, adding 50% overhead when moving pointer over content and 100% when over a handle. The latter can still be improved.

| scenario | this PR earlier | this PR now | `main` |
|--|--|--|--|
| pointer over content|	34.5 ms|	6.0 ms|	4.2 ms|
| pointer over a handle|	9.2 ms|	9.1 ms|	4.1 ms |

I agree the `perf/` should be dropped.

packages/widgets/src/dockpanel.ts
emilkrebs: Thanks a lot for your review, will take care of that and drop `perf/`!

Issue #505, opened 2022-12-27
Resizing of panels is slow when there are many nodes in resized notebooks/widgets


krassowski:
## Description

Resizing split/dock panels is slow. https://github.com/jupyterlab/lumino/issues/504 proposes a solution for bottom-down resizing which is specific to Chrome.

## Reproduce

As in https://github.com/jupyterlab/lumino/issues/504 but left-right.

## Expected behavior

Resizing does not jitter the UI allowing users to precisely set the layout they need at a first attempt.

## Proposed solution

When we start the drag action of a `SplitPanel` we could throttle the relayout of the contents.
This requires isolating the contents of the panel to preserve its layout by pinning width and height, adding `contain:strict` and then updating the width/height every once in a while to match the size of the outer container.

There is a trade-off: throttling the contents updates would mean that contents are sometimes out of sync with the container size. Some applications deal with this by stretching the contents in compositor (using `transform: scale`) but it only looks good if the content has little text and resizes proportionally rather than reflowing.

Ideally we would be able to only isolate the problematic elements. With windowed notebook in 4.x we are not concerned about the notebook having many cells but about cells having large outputs\*, e.g. an SVG with 1000s elements ; if we could contain the SVG size (throttle its updates) for the time of resizing, the UX would be much nicer as the overall layout of notebook would still reflow smoothly.

Technically we would need to implement a special method in split panel children to handle requests:
- "isolate large elements" (invoked by `SplitPanel` when user starts resizing)
- "update size of isolated elements" (invoked by `SplitPanel` with throttling on mousemove when resizing)

\*) or by individual cells having incredibly long code, say thousands of lines

vidartf:
An alternative to throttling the relayout is to replace the content of a node with a placeholder image (not a screenshot but a generic "ui element here" image). We might also be able to only do this conditionally if we detect that layout events are slow (e.g. we are resizing a panel, and start updating layout as we go. This assumes that we can know when the layout has completed).

krassowski:
> We might also be able to only do this conditionally if we detect that layout events are slow

We could that with some elements of the https://developer.mozilla.org/en-US/docs/Web/API/Performance_API

krassowski:
I opened a draft PR to show what we could achieve without throttling, feedback on UX questions is welcome: https://github.com/jupyterlab/lumino/pull/812 (code itself is not ready for review)

PR #812, opened 2026-04-21
Freeze size of heavy nodes when resizing panels

krassowski:
### Description

- tackles #505 
- this is an early draft; I plan to resolve duplication and drop some changes to reduce diff later on

#### Trigger condition

- using the Performance API works but we can detect the `longtask` or `long-animation-frame` but it comes after user already saw a freeze for a few seconds; I think this is not the right approach - the Performance API was designed primarily as a measurement tool, not to make the rendering system reactive.

#### Freeze implementation
- using `contain: strict` alone is not enough because it tells the layout engine that when considering layout of _parents_ it can use the size as given in `width`/`height` but the browser still needs to relay all the text nodes inside the widget, which is more expensive
- we can address this by using `minWidth`/`maxWidth` on the widgets with many children; it works but it is still not as fast as it could be; we know that it could be faster because resizing the same panels top-down is super smooth (without tricks) since we enabled strict containment. I believe that the difference is that the leaf layout, including text reflow, is never touched in top-down resize because the content is just scrolled rather than re-laid
- freezing the sizes of all leaves of heavy widgets does give a smooth resizing experience; this allows the rendering pipeline to skip layout of the contents altogether when resizing
- I was also experimenting with applying `content-visibility: auto` to DOM children of heavy widgets, I am yet to confirm if this is warranted in the final form

Freezing the width means that the `Widget` in panels which have `overflow: auto` enabled get scrollbars, which you can see on the recordings below.

#### Throttled/delayed updates
- we can update periodically to ensure user gets some idea of how the content will look like; the more often we perform the update, the more jittery the resize is. In recordings below I am using updates once every 3 seconds
- we can update when user pauses resizing; currently I wait 300ms to decide if user paused resizing
- the updates are staggered, doing things one panel (group) at a time, spread across frames to avoid any long tasks/dropped frames

### Before

[Screencast From 2026-04-21 15-21-43.webm](https://github.com/user-attachments/assets/2776d939-209a-4313-a214-73d057d24fbb)

### After (active on pointer down, no long animation frame detection)

[Screencast From 2026-04-21 15-11-26.webm](https://github.com/user-attachments/assets/b7550824-f4fe-4ff2-8570-c34bdb0caafc)

### After (activates only after long frame/task detection)

[Screencast From 2026-04-21 14-52-24.webm](https://github.com/user-attachments/assets/f34cb5cf-7512-401d-a4b7-f4430ca1cdc2)

### DIFF
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -1184,6 +1184,7 @@ export class DockPanel extends Widget {
       entry.el.style.height = `${entry.rect.height}px`;
       entry.el.style.minHeight = `${entry.rect.height}px`;
       entry.el.style.contain = 'strict';
+      entry.el.classList.add('lm-layout-frozen');
     }
 
     console.log(`[DockPanel] froze ${this._frozenElements.length} elements`);
@@ -1274,6 +1275,7 @@ export class DockPanel extends Widget {
       entry.element.style.minWidth = entry.prevMinWidth;
       entry.element.style.height = entry.prevHeight;
       entry.element.style.minHeight = entry.prevMinHeight;
+      entry.element.classList.remove('lm-layout-frozen');
     }
     this._frozenElements = [];
   }
@@ -1686,7 +1688,7 @@ namespace Private {
    * The periodic interval (in ms) for refreshing frozen element
    * sizes during long continuous drags.
    */
-  export const REFRESH_INTERVAL_MS = 2000;
+  export const REFRESH_INTERVAL_MS = 5000;
 
   /**
    * An object which holds mouse press data.
--- a/packages/widgets/src/dockpanel.ts
+++ b/packages/widgets/src/dockpanel.ts
@@ -761,6 +761,9 @@ export class DockPanel extends Widget {
     // pointerdown will detect if the resulting frame is slow.
     let layout = this.layout as DockLayout;
     layout.moveHandle(this._pressData.handle, xPos, yPos);
+
+    // Debounce: refresh frozen elements after the user stops moving.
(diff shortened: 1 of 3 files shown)


krassowski:
Testing locally using scripted JupyterLab galata benchmarks, we can drop most optimizations apart for dimension freezing:

### Setup

Two pre-rendered notebooks opened side-by-side (horizontal split):

| Notebook | Content |
|---|---|
| `print_output_notebook` | 1 cell · 1 `stream` output · 10 000 lines of `qwertyuiopasdfghjklzxcvbnm` |
| `display_output_notebook` | 1 cell · 10 000 separate `display_data` outputs |

Measurement: browser `performance.measure()` over a full drag gesture - pointer-down on the divider handle, move 200 px right, move back, pointer-up (40 synthetic pointer events via Playwright).
Browser: Chromium · 3 samples per configuration.

### Results

| Configuration | What is applied | Mean (ms) | Δ vs no optimisation |
|---|---|---:|---:|
| No optimisation | - | **1453** | - |
| `contain: strict` only | `contain: strict` on the panel head  | 1421 | −2 % (noise) |
| `content-visibility: auto` only | `content-visibility: auto` + `contain-intrinsic-*` on children | 1507 | +4 % (overhead) |
| Freeze dimensions only | `width/min-width/max-width`, `height/min-height` fixed to measured rect | **727** | **−50 %** |
| All (current code) | All of the above combined  | **757** | **−48 %** |

krassowski:
I think this is better maintained downstream rather than contributed here - there are some assumptions on the content which go in. Closing. If someone wants to see the same in action we are maintaining `OptimizedDockPanelSvg` in JupyterLab now.

PR #818, opened 2026-05-18
Fix menu not closing when clicking tabs

### LINKED ISSUE #18851: Menus Remain Open When Clicking Interface Buttons
## Description

Similar to https://github.com/jupyterlab/jupyterlab/issues/16853, menus do not close when clicking certain buttons in the interface.

Using the context menu as an example, the issue also affects other menus such as the Settings menu. These menus close when clicking elsewhere in the document, but remain open when clicking interface buttons.

https://github.com/user-attachments/assets/dd46e25d-ca69-4b87-a1ea-43791c2c17d6

## Steps to Reproduce

1. Open a notebook
2. Right-click to open the context menu
3. Click the notebook close button
4. Observe that the context menu remains open

## Expected Behavior

Menus should close when clicking interface buttons as well.

## Context

- Operating System: macOS 26.3.1
- Browser: Brave
- JupyterLab version: 4.6.0a5

Darshan808:
### Description

Switch document listener from `mousedown` to `pointerdown` so outside clicks are detected even when tab clicks call 
`preventDefault()`, which was blocking the old event flow.

#### Before:
https://github.com/user-attachments/assets/4ef2e319-ffb4-4894-98d6-25120a4c509d

#### After:
https://github.com/user-attachments/assets/702d03dd-846e-499f-9e75-3870da55e735

### DIFF
--- a/packages/widgets/src/menu.ts
+++ b/packages/widgets/src/menu.ts
@@ -504,20 +504,20 @@ export class Menu extends Widget {
       case 'keydown':
         this._evtKeyDown(event as KeyboardEvent);
         break;
-      case 'mouseup':
-        this._evtMouseUp(event as MouseEvent);
+      case 'pointerup':
+        this._evtPointerUp(event as PointerEvent);
         break;
-      case 'mousemove':
-        this._evtMouseMove(event as MouseEvent);
+      case 'pointermove':
+        this._evtPointerMove(event as PointerEvent);
         break;
-      case 'mouseenter':
-        this._evtMouseEnter(event as MouseEvent);
+      case 'pointerenter':
+        this._evtPointerEnter(event as PointerEvent);
         break;
-      case 'mouseleave':
-        this._evtMouseLeave(event as MouseEvent);
+      case 'pointerleave':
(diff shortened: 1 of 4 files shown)


packages/widgets/src/menu.ts
```diff
@@ -516,7 +516,7 @@ export class Menu extends Widget {
       case 'mouseleave':
         this._evtMouseLeave(event as MouseEvent);
         break;
```
krassowski, line 518: Should we also update the other event listeners? I think in quite a few places we migrated away from `mouse*` to `pointer*` events, see https://github.com/jupyterlab/lumino/pull/319, not sure why this one was not covered.

krassowski (approved):
Looks good to me.

Another very long shot but @afshin any chance you recall why this was not changed in lumino v2 PR? More likely forgotten about or was there a reason not to do it?

afshin:
@krassowski @Darshan808 I can't remember a reason. I think this change is good. It's possible I was being cautious for some now-forgotten reason. It's possible it was just an omission, but it's pretty big, so I think it was more likely caution than anything else. Thanks for the updates!

PR #819, opened 2026-05-18
Fix incorrect accordion title state on widget move

Darshan808:
### Description
`attachWidget` always created a new title with `expanded: true`, so moved hidden widgets appeared expanded in the new panel.
Now we pass `!widget.isHidden` so the title correctly reflects the widget’s actual state.

#### Before
https://github.com/user-attachments/assets/2d453371-8c18-40cd-8040-db6933525675

#### After
https://github.com/user-attachments/assets/2a24bb90-e591-45a2-beec-65a145f5e42e


krassowski (approved):
Thank you!

PR #821, opened 2026-05-25
Disable scripts and telemetry in `.yarnrc.yml`

krassowski:
Some hardening in for todays reality. Aligns it with settings used by https://github.com/jupyterlab/jupyterlab/blob/main/.yarnrc.yml


jtpio (approved):
Thanks!

PR #823, opened 2026-05-27
Assign `menuitemcheckbox` role to toggleable menu items

### LINKED ISSUE #755: Assign `menuitemcheckbox` role to all entries which can be toggled
> I think ideally the role for menu entries which can be toggled would always have `menuitemcheckbox` role, not only when they are toggled. I appreciate it is not trivial to achieve with the information available on menu items right now so it would require a larger change. I think this PR (#753) is a significant improvement as-is though.

_Originally posted by @krassowski in https://github.com/jupyterlab/lumino/pull/753#pullrequestreview-2802482750_

MUFFANUJ:
Closes #755 

This updates the menu `ARIA` behavior so any toggleable command item is always rendered with `role="menuitemcheckbox"` (not only when toggled), while non-toggleable items remain `menuitem`. It also sets `aria-checked` to the current `true`/`false` state.

### DIFF
--- a/packages/widgets/src/menu.ts
+++ b/packages/widgets/src/menu.ts
@@ -1152,7 +1152,7 @@ export namespace Menu {
     /**
      * Whether the menu item is toggleable.
      */
-    readonly isToggleable: boolean;
+    readonly isToggleable?: boolean;
 
     /**
      * Whether the menu item is visible.
@@ -1378,7 +1378,7 @@ export namespace Menu {
           if (!data.item.isEnabled) {
             aria['aria-disabled'] = 'true';
           }
-          if (data.item.isToggleable) {
+          if (data.item.isToggleable || data.item.isToggled) {
             aria.role = 'menuitemcheckbox';
             aria['aria-checked'] = `${data.item.isToggled}`;
           } else {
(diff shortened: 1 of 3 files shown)


packages/widgets/src/menu.ts
```diff
@@ -1149,6 +1149,11 @@ export namespace Menu {
      */
     readonly isToggled: boolean;
 
+    /**
+     * Whether the menu item is toggleable.
+     */
+    readonly isToggleable: boolean;
```
krassowski: Since this is an addition to public API, maybe:

```suggestion
    readonly isToggleable?: boolean;
```

packages/widgets/src/menu.ts
```diff
@@ -1373,9 +1378,9 @@ export namespace Menu {
           if (!data.item.isEnabled) {
             aria['aria-disabled'] = 'true';
           }
-          if (data.item.isToggled) {
+          if (data.item.isToggleable) {
```
krassowski: For backward compat if we are making item interface backwart-compat (with optionally undefined `isToggleable`):

```suggestion
          if (data.item.isToggleable || data.item.isToggled) {
```

krassowski (commented):
Looks good, just compatibility note

packages/widgets/src/menu.ts
MUFFANUJ: Sure

krassowski (approved):
Thanks!

PR #827, opened 2026-06-17
Document the use of `PluginRegistry` on the server

jtpio:
Add a new example to demonstrate how the `PluginRegistry` can be used in a node application.

Documentation: https://lumino--827.org.readthedocs.build/en/827/plugin-registry-server.html

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -15,3 +15,12 @@ Rendered static examples
 `DockPanel <examples/dockpanel/index.html>`_
 
 `MenuBar <examples/menubar/index.html>`_
+
+.. seealso::
+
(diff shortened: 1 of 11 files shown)


jtpio:
cc @agoose77 maybe this is good enough already for a first iteration.

Overall the Lumino documentation could be benefit from general improvements, and next maybe there could be a section on module federation for this server-side use case.

krassowski (approved):
Looks good for a first pass!

Looking at https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#plugins-interacting-with-each-other not sure if deduplication need is worth mentioning.

https://eslint-plugin.readthedocs.io/en/latest/rules/plugin-activation-args/ could be worth a mention too if someone wants to pick only that rule from the linter

jtpio:
Thanks @krassowski for the review!

I opened https://github.com/jupyterlab/lumino/issues/829 to track that.

PR #832, opened 2026-07-02
Apply opacity together with content visibility if enabled via `hiddenMode` setting

krassowski:
- Required for https://github.com/jupyterlab/jupyterlab/pull/18879
- Follow-up to https://github.com/jupyterlab/lumino/pull/497

Before, the content-visibility mode was mostly working, except that the widgets were shining through the "borders" of JupyterLab shell - this is because it only disables their contents painting, and does not hide the container itself; adding the opacity is a simple solution. I benchmarked it and the performance advantage now is more pronounced than before.


Darshan808 (approved):
Thanks.

Issue #254, opened 2021-10-29
Widget resizing doesn't work on some devices and using Browser Developer Tools


martaszmit:
Hi,
While working on resizing widgets on touch screens, I spotted interesting issue. It works fine on my iPad, but it works not very well on my touch-screen laptop, and on every 'normal' laptop using browser developer tools.

It works quite randomly, as you can see on video below.

https://user-images.githubusercontent.com/19773837/139410039-59cf8f1d-96f5-41c7-8d7b-cd0b65844e55.mov

Issue #247, opened 2021-10-11
Touch Events -> Resize widgets 


martaszmit:
Hello!
First of all I would like to thank you guys for last changes in Lumino and for possibility to rearrange widgets on touch screens (@bign8 👏 ).

Regarding this changes, would be great to _add possibility to resize widgets on touch screens_ as well.

What do you think about this proposal?

Cheers!

blink1073:
Hmm, I just tried in a fresh environment with `node v14.17.4` and `yarn 1.22.17` on my macOS 11.6 and it built fine.  What version of yarn and which OS are you using?

martaszmit:
I confirm, works fine now, I started again in a fresh environment 👍

PR #834, opened 2026-07-15
Fix touch click handling for menus and tab close actions

MUFFANUJ:
Refs: https://github.com/jupyterlab/jupyterlab/issues/19124

This fixes touch interactions where menu item clicks or tab close clicks could be lost or immediately dismissed by the browser’s follow-up `click` event. The change lets touch `pointerup` record the intended action, handles the matching click, stops that click from bubbling into surrounding UI, ignores touch hover-style pointer movement for menus, and clears stale pending touch state on close, detach, or a new pointer press. Tests cover touch menu activation, touch tab close, click bubbling suppression, and touch pointer-move behavior

### Before 

https://github.com/user-attachments/assets/0a6d1a1c-845b-4577-bdd9-29354d1f1f6c

### After

https://github.com/user-attachments/assets/ce3655c1-a993-4bce-afec-031539386f75

### DIFF
--- a/packages/widgets/src/menu.ts
+++ b/packages/widgets/src/menu.ts
@@ -30,6 +30,8 @@ import {
   VirtualElement
 } from '@lumino/virtualdom';
 
+import Utils from './utils';
+
 import { Widget } from './widget';
 
 interface IWindowData {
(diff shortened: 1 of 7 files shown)


krassowski (approved):
Looks good, tested on preview (https://lumino--834.org.readthedocs.build/en/834/examples/dockpanel/index.html).

Of note:
- Related to https://github.com/jupyterlab/lumino/issues/77
- Resizing panels does not work reliably; this is tracked in #247 and #254

MUFFANUJ:
> * Resizing panels does not work reliably; this is tracked in #247 and #254

thanks for pointing to this i will look into these and open a PR for these soon.

PR #836, opened 2026-07-17
Add extension points for customizing command palette search results

### LINKED ISSUE #735: Support recent commands
### Problem

Users of the command palette may often run the same command.

However when opening the command palette again the same set of commands is always displayed in the same order. And users have to search for the commands.

### Proposed Solution

Have the `CommandPalette` offer an opt-in way to support recently executed commands.

For reference VS Code offers something like this:

image

With a setting to control the "number of recently used commands to keep in history":

image

So the Lumino `CommandPalette` could also expose a way to configure the history size.

This change should of course be backwards compatible.

### Additional context

Some comments in https://github.com/jupyterlab/jupyterlab/issues/4365 indicate we could add a `frequency` or "execution count" for each command item in the `CommandPalette`.

jtpio:
Fixes https://github.com/jupyterlab/lumino/issues/735

This will allow downstreams like JupyterLab to better show recently used commands in the palette:

image

### DIFF
--- a/examples/example-dockpanel/src/index.ts
+++ b/examples/example-dockpanel/src/index.ts
@@ -301,7 +301,7 @@ function main(): void {
   bar.addMenu(emptyMenu);
   bar.id = 'menuBar';
 
-  let palette = new CommandPalette({ commands });
+  let palette = new CommandPalette({ commands, maxRecentCommands: 3 });
   palette.addItem({ command: 'example:cut', category: 'Edit' });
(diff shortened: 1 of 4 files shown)


jtpio:
Opened early for visibility and it's still a draft for now.

Gonna check if there could be some more generic API / hooks the palette could expose, so downstream like JupyterLab would own the implementation of the recent feature instead.

